PR #15, opened 2015-05-06
Updated typescript dependency

vasek17:
Did not work with current typescript 1.5.0-beta.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "colors": "^1.0.3",
     "loader-utils": "^0.2.6",
     "object-assign": "^2.0.0",
-    "typescript": "~1.5.0"
+    "typescript": "~1.5.0-beta"
   },
   "devDependencies": {
     "jsx-typescript": "^1.5.0-alpha.4",


jbrantly:
Really? This has worked fine for me so far in all of my various tests. "~1.5.0" was enough to pull "1.5.0-beta".

vasek17:
Getting npm ERR ETARGET
No compatible version found: typescript@'>=1.5.0 <1.6.0'

jbrantly:
What version of NPM are you using so I can compare to what I'm using. Going to go ahead and merge though because I'd rather it work for everyone.

vasek17:
I am using npm 2.7.5. 
If I mark my app version (where listing ts-loader as a dependency) as beta, it works fine for me with ~1.5.0. But if the version is stable, tilde operator does not consider prereleases.

jbrantly:
Ah, good to know. Thanks!

jbrantly:
Published in v0.4.2. Thanks again!

PR #22, opened 2015-06-25
Adding an option to suppress log messages.

NickHeiner:
In my project, I run `webpack` from the shell, and expect its output to be valid json. Having this module log to the console breaks the json parsing. In another context, I am spawning a process that where running this causes a console log that breaks the TAP output. In both cases, I would like the ability to turn off the logging.

I thought about adding a test but it seems like it would be quite a bit of work, and I'm not sure it's a good ROI.

Thanks for making this module!

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,9 @@ Most TypeScript-related options should be set using a
 file. There are a few loader-specific options you can set although in general
 you should not need to.
 
+##### silent *(boolean) (default='false')*
+If true, no console.log messages will be emitted.
+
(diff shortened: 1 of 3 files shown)


jbrantly:
Good idea! I'm going to merge but I'll make a couple tweaks before publishing (a `console.log` was missed). Also, agree that a test here isn't worth it. Thanks!

NickHeiner:
Awesome, thanks!

jbrantly:
Published as v0.4.5

PR #27, opened 2015-07-31
Handle declaration

Nemo157:
Handle cases where the tsconfig makes tsc output declaration files by searching
for the correct file by extension instead of assuming order of output files.


jbrantly:
Thanks for this! I'll also add a test specifically for this case.

PR #32, opened 2015-08-25
Added compile error category and code to output

use-strict:
Match the output pattern of 'tsc', for integration with VSCode.


jbrantly:
Good call. Thanks! I'll get a release out before tomorrow.

jbrantly:
Published in v0.5.1

PR #36, opened 2015-08-28
Fix for ES6 with transpileOnly option

thomasguillory:
When using the transpileOnly option, the "special handling" code block for ES6 is skipped, preventing to target ES6 when using this option.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -203,6 +203,14 @@ function ensureTypeScriptInstance(options: Options, loader: any): { instance?: T
     objectAssign(compilerOptions, configParseResult.options);
     filesToLoad = configParseResult.fileNames;
     
+    var libFileName = 'lib.d.ts';
+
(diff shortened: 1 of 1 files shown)


jbrantly:
Thanks! Apparently there is some weirdness in the test runner where this wasn't getting picked up by the es6 test. I'm investigating that now.

Also, the reason the integration tests failed for this PR has nothing to do with the PR. The latest TypeScript nightly broke the build process is all.

jbrantly:
Alrighty. I was able to fix the test runner so that this issue is actually exposed (see https://travis-ci.org/jbrantly/ts-loader/builds/77835580). Re-running the travis build on this PR fixes the issue and passes. Thanks!

jbrantly:
Published in v0.5.3

thomasguillory:
Awesome, thanks!

PR #51, opened 2015-09-23
Update repository links to TypeStrong

blakeembrey:
A small start :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/jbrantly/ts-loader.svg?branch=master)](https://travis-ci.org/jbrantly/ts-loader)
+[![Build Status](https://travis-ci.org/TypeStrong/ts-loader.svg?branch=master)](https://travis-ci.org/TypeStrong/ts-loader)
 [![Downloads](http://img.shields.io/npm/dm/ts-loader.svg)](https://npmjs.org/package/ts-loader)
 
 # TypeScript loader for webpack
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/jbrantly/ts-loader.svg?branch=master)](https://travis-ci.org/jbrantly/ts-loader)
+[![Build Status](https://travis-ci.org/jbrantly/ts-loader.svg?branch=master)](https://travis-ci.org/TypeStrong/ts-loader)
```
jbrantly: Need to change the path to the SVG as well

jbrantly:
How do we want to do this? Merge your own after you get a +1? Let the other person merge?

README.md
blakeembrey: Fix, thanks :+1:

blakeembrey:
How about between collaborators, they merge their own PR after a LGTM and for other contributors we can merge after a review?

jbrantly:
Saw your message literally 10 seconds after I merged :grin: Sounds good to me, will go with that in the future.

blakeembrey:
Haha, it's all good - I don't think it's too important but if it's a new feature (not a patch) we could make sure there's at least two eyes on it to comment. E.g. that'll still be the same if you're the one that opens it since you're the maintainer :smile: I think if we both sign off it's not too important who merges, as long as it's passing.

PR #52, opened 2015-09-25
Watch imports that are not emitted

jbrantly:
There is a bug where if you use an import that is not emitted that import is not watched by webpack and any updates to it will not be reflected when watching.

This is just a test for the moment for documentation purposes. I'll add a fix at a later date. I've looked at this a little bit already and my current thinking is to use preProcessFile to get a list of imports and adding them as a dependency so that they'll be watched.

### DIFF
--- a/test/importWatch/app.ts
+++ b/test/importWatch/app.ts
@@ -1,0 +1,5 @@
+import { Something } from "./something";
+
+var a: Something = {
+	foo: 'bar'
+}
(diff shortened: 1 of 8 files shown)


blakeembrey:
Ah, yes, my solution was a bit jank but I solved it by doing a hack where I kept the current loader around (since it's all synchronous) and added anything to that (https://github.com/blakeembrey/typescript-simple-loader/blob/master/src/typescript-simple-loader.ts#L77-L79).

jbrantly:
That's really interesting! I would have assumed that `getScriptSnapshot` would only be called once for any particular version of the file. I'll have to explore that method for sure.

johnnyreilly:
Closing as covered by importsWatch test I believe

PR #69, opened 2015-09-29
Add contributor's guide

jbrantly:
First pass at a contributor's guide. Does this make sense? Anything confusing or left out?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,87 @@
+# Contributer's Guide
+
+We welcome contributions from the community and have gathered guidelines 
+here to help you get started.
+
+## Discussion
+
+While not absolutely required, it is encouraged that you first open an issue 
+for any bug or feature request. This allows discussion on the proper course of
+action to take before coding begins.
+
+## Building
+
+To build from source you first need to `npm install`. You will also need a
+version of TypeScript available, either globally or install locally with
+`npm install typescript`. Then run `node build`.
+
+## Changing
+
+Most of the information you need to contribute code changes can [be found here](https://guides.github.com/activities/contributing-to-open-source/).
+In short: fork, branch, make your changes, and submit a pull request.
+
+## Testing
+
+This project makes use of a test suite both to make sure we don't break
+anything and also to make sure new versions of webpack and TypeScript don't
+break anything. The tests are full integration tests, meaning each test is a
+mini project that is run through webpack and then the output is compared to the
+expected output. Not all bugs/features necessarily fit into this framework and
+that's OK. However, most do and therefore you should make every effort to
+create at least one test which demonstrates the issue or exercises the feature.
+
+The test harness uses certain conventions. All tests have their own directory
+under `/test`, eg `/test/someFeature`. Each test should have a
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,87 @@
...
+
+Each test should have an `expectedOutput` directory which contains any webpack
+filesystem output (typically `bundle.js` and possibly `bundle.js.map`) and any 
+console output. stdout should go in `output.txt` and stderr should go in
+`err.txt`.
+
+As a convenience it is possible to regenerate the expected output from the 
+actual output. This is useful when creating new tests and also when making a
+change that affects multiple existing tests. To run use 
+`npm test -- --save-output`. Note that all tests will automatically pass when
```
basarat: `-- --save-output`. Typo?

basarat:
nice :sunglasses:

johnnyreilly:
Looks good!

CONTRIBUTING.md
jbrantly: Not a typo. Allows arguments to be passed to the executing script. See https://docs.npmjs.com/cli/run-script

CONTRIBUTING.md
```diff
@@ -0,0 +1,87 @@
...
+
+While not absolutely required, it is encouraged that you first open an issue 
+for any bug or feature request. This allows discussion on the proper course of
+action to take before coding begins.
+
+## Building
+
+To build from source you first need to `npm install`. You will also need a
+version of TypeScript available, either globally or install locally with
+`npm install typescript`. Then run `node build`.
```
blakeembrey: I'd simply this and make `typescript` a dev dependency. Then add an `npm` script for building the project. Here's my normal set up: https://github.com/blakeembrey/typescript-simple-loader/blob/master/package.json#L15-L21

CONTRIBUTING.md
```diff
@@ -0,0 +1,87 @@
...
+`webpack.config.js` file which follows this general convention:
+
+```javascript
+module.exports = {
+    entry: './app.ts',
+    output: {
+        filename: 'bundle.js'
+    },
+    resolve: {
+        extensions: ['', '.tsx?', '.js']
```
blakeembrey: I'm not sure that extensions is a psuedo-regexp - shouldn't it be `'.ts', '.tsx'` as two entries?

CONTRIBUTING.md
jbrantly: Yea I'm good with changing to `npm run build`. I considered doing that originally. Not sure why I didn't. Regarding the dev dependency, that would probably work as long as it doesn't mess up the [travis build](https://github.com/TypeStrong/ts-loader/blob/master/.travis.yml).

blakeembrey:
Awesome :+1:

CONTRIBUTING.md
jbrantly: Wow, you're right. I added that on the fly from a config that didn't have the tsx extension and wasn't thinking. Good catch, I'll fix.

jbrantly:
Fixed the extensions issue. The other note regarding the build script I'll do as a separate PR. Thanks everyone.

PR #70, opened 2015-09-29
Simplify license and update to TypeStrong

jbrantly:
This simplifies the license in the README and transfers the copyright over to TypeStrong. I want it to be more of a community project (that was the point of the transfer, right?) so I feel this is a necessary step. IANAL but I think I can do this.

cc @basarat @blakeembrey

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2015 James Brantly
+Copyright (c) 2015 TypeStrong
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


basarat:
Sure. You can even keep it open (no year, no author) e.g https://github.com/TypeStrong/grunt-ts/blob/master/LICENSE :rose:

jbrantly:
@basarat That is interesting. I did a little research on that this morning and couldn't find much (any) info on an MIT license without a copyright notice. The license text even mentions a copyright notice ("The above copyright notice"). I can assume that even without an explicit copyright notice the work is still copywritten but it's not clear by who, but since IANAL I wish I could find an authoritative source on the matter.

It seems that by explicitly changing to TypeStrong that would allow the organization as a whole to potentially change the copyright and/or license in the future without requiring signoff from me (in case I'm hit by a bus).

basarat:
Cool. Merge away :+1:

PR #75, opened 2015-10-06
Fix issue with `getScriptFileNames`

### LINKED ISSUE #71: Incorrect file list maintenance causes compiler error in particular cases
Prerequisites:
 * There are several TypeScript files in the project (as separated external modules because I want to use multiple bundles)
 * There are some npm based dependencies

The flow which causes to error:
 * When fist file are processed by ts-loader - typescript tries to load dependent npm modules. It checks package.json of the appropriate module (it can contain "typings" with direct reference to .d.ts).
 * TypeScript calls moduleResolutionHost.fileExists and readFile. And it's processed (in the ts-loader) through servicesHost.getScriptSnapshot (which is incorrect I think). As result - package.json is added to files array.
 * When next file is loaded using ts-loader (but in the same webpack instance of course) - TypeScript compiler calls serviceHost.getScriptFileNames and it returns package.json files which causes build fail without any understandable error, only "TypeScript emitted no output", because ts-loader doesn't show error messages from getCompilerOptionsDiagnostics and there is no error for specific source file, just for input file list (it contains json files).

As a temporary fix (obiously, it's better to decouple module resolution host from the service host, it's just a workaround) I fixed ts-loader/index.js (directly in the my node_modules) using following code in the getScriptFileNames:

```
return Object.keys(files).filter(function(f) { return !!f.match(/\.tsx?$/); });
```

It causes that json files are not returned for TypeScript input but still are saved in the files array (maybe it's ok to cache package.json in this array, I'm not sure actually).

Also I suggest to add getCompilerOptionsDiagnostics to error list in some way. Because it's very hard to understand what is causing error in some cases.

jbrantly:
This addresses #71 using the suggested fix to filter the results of `getScriptFileNames`.

I considered not using `getScriptSnapshot` for `moduleResolutionHost` but the caching is useful so leaving as-is.


jbrantly:
I originally had this making changes to the `nodeResolution` test but that broke the transpile version (you can't specify both `isolatedModules` and `noEmitOnError`). So instead I just made a specific test just for this issue, probably better anyways.

@blakeembrey Could you review?

index.ts
```diff
@@ -271,7 +271,7 @@ function ensureTypeScriptInstance(options: Options, loader: any): { instance?: T
 
     // Create the TypeScript language service
     var servicesHost = {
-        getScriptFileNames: () => Object.keys(files),
+        getScriptFileNames: () => Object.keys(files).filter(filePath => !!filePath.match(/(\.d)?\.tsx?$/)),
```
blakeembrey: You could actually remove the `(\.d)?`, since it still ends with `.ts`.

jbrantly: You and your logic.... :grin:

blakeembrey:
LGTM once you make the minor regexp change :+1:

index.ts
blakeembrey: Oh, also, I'd use `Regexp.prototype.test` instead of `match`, it's always boolean - `/\.tsx?$/i.test(filePath)`. I added the `i` flag.

jbrantly:
Updated based on your comments. Also pulled the regex out for a little speed.

Issue #60, opened 2015-09-25
Loader option: ignoreWarnings


jbrantly:
We need to add the loader option `ignoreWarnings` like [typescript-simple-loader](https://github.com/blakeembrey/typescript-simple-loader/blob/master/src/typescript-simple-loader.ts#L223-L225)

One topic of discussion here is that ts-loader only emits TS errors as webpack errors and not warnings. Before we add this we need to decide between errors only vs errors/warnings, and if we go errors only then the option should probably be called `ignoreErrors`. Or we could sidestep the issue and come up with a more generic name.

jbrantly:
Idea for a generic name: `ignoreDiagnostics`

Reasons why ts-loader does what it does:

1. webpack errors aren't necessarily errors in the sense that there is no output. Files with TS errors continue to be emitted.
2. I looked into making TS warnings into webpack warnings, but when I looked I found out [there are no TS warnings](https://github.com/Microsoft/TypeScript/blob/master/src/compiler/diagnosticMessages.json)
3. Using errors interacts with the webpack NoErrorsPlugin which causes webpack to not emit if there are errors. I feel like this puts the user in ultimate control.

gdi2290:
this feature is need for alpha software where the typings may produce unneeded warning.  For this reason the feature was requested in typescript-simple-loader. Since we're merging all the ts loaders I would like to replace typescript-simple-loader in my starter repo which uses alpha software (Angular 2)

https://github.com/angular-class/angular2-webpack-starter

jbrantly:
Yup, I was actually planning on implementing this tonight for that very reason.

gdi2290:
I need to ignore this one error
2309 -> An export assignment cannot be used in a module with other exported elements.

jbrantly:
This was published in v0.5.6

PR #77, opened 2015-10-07
ignoreDiagnostics feature

jbrantly:
This implements #60 

Note that there is an API change from typescript-simple-loader which named this option `ignoreWarnings`. I've opted for `ignoreDiagnostics` since it's more agnostic to warnings vs errors, but certainly open to feedback.

I shamelessly ripped some code from typescript-simple-loader for this. The basic idea is that I've added `loaderOptions` to the `instance` object (and made a partially hydrated `instance` object available earlier in the process). Then I changed the `formatErrors` function to take an instance object instead of the TypeScript compiler. This allows access to both the compiler and the ignoreWarnings option.

There was also some cleanup (as a separate commit) to help distinguish between loader options and compiler options in the code.

cc @blakeembrey @gdi2290

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,11 @@ of this.
 If true, no console.log messages will be emitted. Note that most error
 messages are emitted via webpack which is not affected by this flag.
 
+##### ignoreDiagnostics *(number[]) (default=[])*
+
+You can squelch certain TypeScript errors by specifying an array of diagnostic
+codes to ignore. 
+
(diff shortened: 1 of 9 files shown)


blakeembrey:
LGTM :+1:

gdi2290:
awesome, can we get a new version :+1:

jbrantly:
Done, v0.5.6

PR #96, opened 2015-11-04
Fix "`lib.es6.d.ts` could not be found"

jbrantly:
Tests were failing for typescript@next saying that `lib.es6.d.ts` could not be found. The problem was that `getDefaultLibFileName` should be returning the full path to the default lib and not just the file name.

@blakeembrey


jbrantly:
Published in v0.6.1

Issue #78, opened 2015-10-08
Incremental build performance dropped


use-strict:
Hi,

I noticed that our incremental builds started to be considerably slower than they used to, so I ran webpack with `--profile --progress` and older versions of ts-loader to see if there are differences. It seems that v0.5.2 is twice as fast as the latest version. I'm unsure how to interpret the profiling data and the webpack source code is a complete mess, so it doesn't help much in that regard.

Here are the stats:

```
// ts-loader v0.5.2
379ms build modules   
19ms seal
39ms optimize
60ms hashing
35ms create chunk assets
2ms additional chunk assets 
0ms optimize chunk assets 
0ms optimize assets 
99ms emit
Hash: fdd096aa48ded051dda3
Version: webpack 1.12.2
Time: 639ms
```

```
// ts-loader v0.5.6
357ms build modules   
19ms seal
50ms optimize
51ms hashing
40ms create chunk assets
4ms additional chunk assets 
0ms optimize chunk assets 
826ms optimize assets
78ms emit
Hash: fdd096aa48ded051dda3
Version: webpack 1.12.2
Time: 1429ms
```

It looks like there is a lot of time spent in "optimize assets". I don't know what's causing this, but I have a feeling that ts-loader is using some extra dependencies now which may have the side-effect of enabling some webpack optimizations that are meant for production, not development.

jbrantly:
Hmm... interesting. Thanks for reporting, I'll take a look.

use-strict:
I did a little profiling with node-inspector and it seems that time is spent calling ```languageService.getSemanticDiagnostics()```. Can't figure out from the history, if it's called more often than it's supposed to be or if it's a side-effect of some other change. Any ideas?

use-strict:
Could be this commit: https://github.com/TypeStrong/ts-loader/commit/dc53ef6301ebce21b2d0365c7cf40548a478c067

Before it was checking for "visited" modules.

jbrantly:
Yea, that was my thought as well. It was put in place to fix [this](https://github.com/TypeStrong/ts-loader/tree/master/test/dependencyErrors) but I think #68 is the way to go and there is a decent chance that it will be more performant.

use-strict:
Yes, seems a reasonable assumption. Also, in my case there were only ~20 files or so, because the rest are still ye olde' JS files, but in a larger project, this could sum up considerably to the overall incremental build time.

SonofNun15:
[...]

Issue #56, opened 2015-09-25
Evaluate implementing `getProjectVersion`


jbrantly:
For reference: https://github.com/Microsoft/TypeScript/pull/3131

It seems like this could potentially have speed benefits but it's not clear to me exactly where (I think anywhere that `synchronizeHostData` is called which is pretty much everywhere).

jbrantly:
Closing since ts-loader supports this as of v0.6.1

PR #100, opened 2015-11-06
Improve initial build performance by reducing calls to `synchronizeHostData`

jbrantly:
This PR improves initial build performance. Times vary but I've seen a 4s second build go to 2.5s and a 22s build go to 12s.

The issue is that ts-loader too aggressively increases each file's version number which was invalidating TypeScript's cache. There is a TS method called `synchronizeHostData` that is very expensive to run and is called all over the place. However, it has two ways to short-circuit. One is using `getProjectVersion` and the other is by reading all of the files' `version` property. This PR handles both. When the initial call to `getEmitOutput` is run for the entry file, TypeScript essentially loads up the entire program at that point. All of the files are read from disk, put in memory, parsed, etc. After the entry file is done, webpack calls the loader for each of the dependent files. Previously we would increase the file's `version` number at this point. However, in most cases, the content of the file is exactly the same. By checking if the content is the same and not increasing the file's `version` we do not invalidate TypeScript's cache and it can emit the file much more efficiently. Additionally `getProjectVersion` is used to gain a small performance increase (TS doesn't need to iterate each file's `version`).

The early call to `getCompilerOptionsDiagnostics` was also causing `synchronizeHostData` to run. I moved this to later in the process so it can use the cached data.

One test broke with this change, but I believe I'm just going to update the test. The scenario is that a `.d.ts` is changed which causes errors in the program. Previously the bundle would be re-emitted in this case, but with these changes it won't be. I don't think this is a problem because nothing in the emit actually changes. It's actually a little more efficient.

This partially addresses some of the points regarding initial build time in #78. It also addresses #56.

@blakeembrey


use-strict:
Very nice. Good to have this stuff documented.

blakeembrey:
LGTM :+1: 

I've also seen the perf time issues be around 20% slower with `ts-loader` than `typescript-simple-loader`, but I haven't looked into it 100% yet. Might be something worth profiling against too, since other loaders haven't implemented these things but seem to still be a bit quicker.

jbrantly:
Thanks!

> Might be something worth profiling against too, since other loaders haven't implemented these things but seem to still be a bit quicker.

True. Off the top of my head I bet that's (at least partially) due to `program.getDiagnostics` vs getting diagnostics for each individual file. Something to look into.

jbrantly:
Published in v0.6.1

Issue #95, opened 2015-11-03
Info string on stdout causes problem to --json webpack output


alexdrel:
Hi, 
The ts-loader writes info string to stdout:
>ts-loader: Using typescript@1.6.2 and /[skipped]/src/tsconfig.json

When building JSON module statistics it goes to the json file and has to be removed manually.
>webpack  --progress -j > m.json   

Of course it is a minor nuisance but it would be nice if it will be written to stderr instead of stdout.

Thanks

jbrantly:
If you use the [silent flag](https://github.com/TypeStrong/ts-loader#silent-boolean-defaultfalse) that info won't be written. I'm not sure using stderr would be correct since it's not an error.

blakeembrey:
@jbrantly It's not uncommon to use `stderr`. I just realised today that NPM logs to output to stderr and not stdout.

jbrantly:
@blakeembrey Interesting. Maybe something to look at then.

@alexdrel For the time being does the `silent` flag solve your need?

alexdrel:
`silent` flag works fine but it is actually simpler to remove the first line from json than to change `webpack.config.js` back and forth. 
It is not a big deal, but I consider `stderr` as suitable for anything which is not primary output - the `progress` flag for example writes to `stderr`.

jbrantly:
Why do you have to change the webpack config back and forth? Just leave the silent flag on. All it affects is that one message which you probably don't care much about once things are "set up".

Re: stderr, I'm open to that.

alexdrel:
@jbrantly - I kinda like it :) - it shows ts version and I use to jump between nightly,1.6.2 and 1.7.2 (which I don't quite understand - is it official and how it made its way into sublime plugin).

jbrantly:
> I kinda like it :)

That's... actually a really good reason :grin:

jbrantly:
Fixed in v0.7.0. Thanks for reporting!

PR #103, opened 2015-11-08
Change all log messages to output to stderr

jbrantly:
This addresses #95


jbrantly:
@blakeembrey When you get a moment could you review this? (forgot to cc originally). Thanks!

blakeembrey:
Interesting, never seen `Console` used like this (or at all). LGTM :+1:

jbrantly:
Yea, at first I was going to use `process.stderr.write` but I wanted to keep the `console.log` semantics. So came up with this after reading the Node docs a bit :grin: Thanks!

PR #107, opened 2015-11-13
Add AppVeyor for Windows builds

jbrantly:


jbrantly:
This PR adds AppVeyor support. Full support is still spotty (the `typeSystemWatch` test sometimes times out and sometimes works fine). A number of changes were needed in the test runner to make things work.

- Colors needed to be forced on
- Additional path normalization in the output
- Timings tweaked

I also changed the test runner handling when the number of files between actual and expected are different so that you get more useful failures.

Somewhat unrelated I created a new test for the nightly which made a change to some output. This was done so that I could see passing tests.

Thanks @johnnyreilly for the info about AppVeyor.

johnnyreilly:
No problem - glad you got it working!

johnnyreilly:
Just curious, what's `Install-Product node $env:nodejs_version` for? I believe the VMs that the builds run on already have node installed...

jbrantly:
> I believe the VMs that the builds run on already have node installed

I was just copying the script [found here](http://www.appveyor.com/docs/lang/nodejs-iojs). Looks like they do have a version installed already but this allows you to set the version specifically.

johnnyreilly:
Ah - gotcha.

PR #110, opened 2015-11-19
Fix watching of definition files

### LINKED ISSUE #109: Is ts-loader adding d.ts files as a dependency to the bundle?
I'm running into a situation where changes to the d.ts files are not getting picked up by Webpack during incremental building. Using `webpack 1.12.3` with `ts-loader 0.6.1`. The loader config:

```javascript
{
  test: /\.tsx?$/,
  exclude: /node_modules/,
  loader: 'babel-loader?plugins[]=babel-relay-plugin-wrapper&cacheDirectory!ts-loader'
}
```

Webpack is invoked using the Node.js API (not watching).

jbrantly:
There was a discrepancy between the test runner and behavior when running webpack from the command line - if the previous bundle had the same hash as the current bundle in watch mode, webpack on the command line does not output any information (almost like nothing happened) whereas in the test runner we continued to operate as if something had been output. The first commit fixes that discrepancy. 

Secondly, a change was made so that if a definition file is changed the hash for the compilation is also changed so that webpack does indeed output something to the screen. This reverts the changes to the `declarationWatch` test made in #100 and fixes #109.

Lastly, since the hashes have changed (and since hashes will now be different based on what your environment is since file paths are now included in the hashed data) the test runner has been updated to replace any hash in webpack output with `[hash]`. This allows the tests to run in any environment without issue.

cc @blakeembrey


index.ts
```diff
@@ -556,7 +556,10 @@ function loader(contents) {
         // Make this file dependent on *all* definition files in the program
         this.clearDependencies();
         this.addDependency(filePath);
-        Object.keys(instance.files).filter(filePath => !!filePath.match(/\.d\.ts$/)).forEach(this.addDependency.bind(this));
+        
+        let allDefinitionFiles = Object.keys(instance.files).filter(filePath => !!filePath.match(/\.d\.ts$/));
```
blakeembrey: `/\.d\.ts$/.test(filePath)`?

index.ts
```diff
@@ -556,7 +556,10 @@ function loader(contents) {
...
+        allDefinitionFiles.forEach(this.addDependency.bind(this));
+        this._module.meta['definitionFileVersions'] = allDefinitionFiles.map(filePath => filePath+'@'+instance.files[filePath].version);
```
blakeembrey: Dot access? `this._module.meta.definitionFileVersions`

blakeembrey:
I think it's ok, mind explaining the `this._module` for me quickly (maybe again)?

index.ts
jbrantly: Will change

jbrantly:
Sure. `this._module` is the object that webpack uses to store info about the module (errors, etc). Most of the loader API operates on it under the hood (eg, `addDependency`, etc). In any case, there is a "meta" property that participates in the hash generation for the module. I use (abuse?) it here to make sure that webpack thinks the module is "new" (even though the actual output didn't change).

jbrantly:
Updated based on code review, merging. Thanks!

PR #121, opened 2015-12-14
Update README.md

johnnyreilly:
Added a link to the readme of the Appveyor build.  Because the [repo isn't public (?)](http://www.appveyor.com/docs/status-badges#badges-for-projects-with-public-repositories-on-github-and-bitbucket) the badge doesn't show up; but the link works just fine.  If you log into AppVeyor and go to this URL:

https://ci.appveyor.com/project/jbrantly/ts-loader/settings/badges

You should be able to get the badge from there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 [![Build Status](https://travis-ci.org/TypeStrong/ts-loader.svg?branch=master)](https://travis-ci.org/TypeStrong/ts-loader)
+[![Build Status](https://ci.appveyor.com/api/projects/status/github/jbrantly/ts-loader?svg=true)](https://ci.appveyor.com/project/jbrantly/ts-loader)
 [![Downloads](http://img.shields.io/npm/dm/ts-loader.svg)](https://npmjs.org/package/ts-loader)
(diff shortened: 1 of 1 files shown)


johnnyreilly:
Since mine's a readme change I suspect this is down to something outside the scope of my PR :smile:

jbrantly:
Yea, the tests have been failing for a couple weeks because I haven't yet had the time to update them. I'll see about carving out some time this evening to update the tests.

jbrantly:
I was just looking at that "repo isn't public" thing. That's really weird. I think it's because the repo is under "TypeStrong/ts-loader" but it's created under my GitHub account which is "jbrantly" so it doesn't think I own it or something.

johnnyreilly:
You don't have to use that approach - (the "inferred name" approach)

If you go to the page I mentioned you should be able to select a URL that looks more like this: https://ci.appveyor.com/api/projects/status/pbkddxcsbmxr9l1b/branch/master?svg=true

which should work just as well. Or they're support is pretty good.  Feodor is very helpful.

jbrantly:
Yea I'm going to do that. I was just interested in why the "public repo" way wasn't working since this is clearly a public repo.

johnnyreilly:
It is peculiar..

PR #122, opened 2015-12-15
Fix tests for 1.7 and latest nightly

jbrantly:
This mostly fixes tests for the nightly which introduced `"use strict"` to the output. Since there's no way to turn this off in TypeScript, and since we want to support testing multiple versions of TypeScript, I'm manually removing `"use strict"` via a special loader.

Also adds specific support for running tests on 1.6, 1.7 and the nightly.


jbrantly:
Closed in favor of #124

PR #125, opened 2015-12-21
Fix AppVeyer badge

jbrantly:


jbrantly:
Fixes AppVeyer badge image as discussed in #121

PR #138, opened 2016-01-17
Made tester mostly resilient to escaped windows end of lines

johnnyreilly:
Whether you want to take this PR or not is arguable.  I say that as while it makes the test reporter more resilient it leaves one problem unhandled. That is to say this:

```
  63 passing (1m)
  2 failing

  1) issue81 should have the correct output:

      Uncaught AssertionError: output.txt is different between actual and expected
      + expected - actual

               Asset     Size  Chunks             Chunk Names
           bundle.js  2.42 kB       0  [emitted]  main
      -bundle.js.map  2.54 kB       0  [emitted]  main
      +bundle.js.map  2.53 kB       0  [emitted]  main
       chunk    {0} bundle.js, bundle.js.map (main) 896 bytes [rendered]
           [0] ./.test/issue81/a.ts 896 bytes {0} [built]

      at C:\source\ts-loader\test\run.js:243:28
      at Array.forEach (native)
      at Watching.handler (C:\source\ts-loader\test\run.js:239:39)
      at Watching._done (C:\source\ts-loader\node_modules\webpack\lib\Compiler.js:81:7)
      at Watching.<anonymous> (C:\source\ts-loader\node_modules\webpack\lib\Compiler.js:61:18)
      at Compiler.emitRecords (C:\source\ts-loader\node_modules\webpack\lib\Compiler.js:282:37)
      at Watching.<anonymous> (C:\source\ts-loader\node_modules\webpack\lib\Compiler.js:58:19)
      at C:\source\ts-loader\node_modules\webpack\lib\Compiler.js:275:11
      at Compiler.applyPluginsAsync (C:\source\ts-loader\node_modules\tapable\lib\Tapable.js:60:69)
      at Compiler.afterEmit (C:\source\ts-loader\node_modules\webpack\lib\Compiler.js:272:8)
      at Compiler.<anonymous> (C:\source\ts-loader\node_modules\webpack\lib\Compiler.js:267:14)
      at C:\source\ts-loader\node_modules\async\lib\async.js:52:16
      at done (C:\source\ts-loader\node_modules\async\lib\async.js:246:17)
      at C:\source\ts-loader\node_modules\async\lib\async.js:44:16
      at C:\source\ts-loader\node_modules\graceful-fs\graceful-fs.js:42:10
      at C:\Source\ts-loader\node_modules\mocha\node_modules\graceful-fs\graceful-fs.js:104:5
      at FSReqWrap.oncomplete (fs.js:82:15)

  2) issue81 should work with transpile:

[...]

PR #142, opened 2016-01-21
escaped and unescaped windows line endings normalised

johnnyreilly:
This is a continuation of https://github.com/TypeStrong/ts-loader/pull/138 but caught up with the latest master.  Hopefully the tests pass now....

### DIFF
--- a/test/run.js
+++ b/test/run.js
@@ -48,6 +48,18 @@ fs.readdirSync(__dirname).forEach(function(test) {
     }
 });
 
+function getFileContents(outputLocation, file) {
+    var contents, fileLocation;
+    try {
+        fileLocation = path.join(outputLocation, file);
+        contents = fs.readFileSync(fileLocation, 'utf8')
+            .replace(/\\r\\n/g, '\\n') // Handle escaped windows end of lines
+            .replace(/\r\n/g, '\n');   // and un-escaped windows end of lines
+    }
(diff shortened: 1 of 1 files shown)


jbrantly:
Is this still failing for you personally due to the filesize issues?

johnnyreilly:
Yes but only due to the filesize issue I mentioned:

```
      -bundle.js.map  2.54 kB       0  [emitted]  main
      +bundle.js.map  2.53 kB       0  [emitted]  main
```

I haven't looked into it but I assume that the one of the webpack options [here](https://webpack.github.io/docs/node.js-api.html#stats-tojson) could be used to prevent filesize being emitted.  Or some other jiggery pokery could be done to exclude this by hand.  

Either way I wanted to get your input on that.  What do you think is a sensible way forward that you would be happy with?  

If you're ok with excluding the file size from the tests so you get a standard test run regardless of line endings I'd be up for looking into this some more.  Your call!

jbrantly:
Looking at the test that's failing (issue81) and all of the other tests, I believe the unique thing about issue81 is that it uses both babel-loader and sourcemaps. None of the other tests do that (some use babel but no sourcemaps, others use sourcemaps but no babel). I'm wondering then if it's a difference in the babel version. AppVeyer is running babel-loader 5.4.0.

johnnyreilly:
Same babel version.....   Hmmmmm

johnnyreilly:
Closing as I think this was handled elsewhere

Issue #108, opened 2015-11-18
webpack exit code is equal to 0, if typescript build fails


pepe79:
When the build fails (i.e. at least one typescript compile error occured), the ts-loader emits an incomplete javascript file and does not propagate the build failure to webpack. When using webpack with the --bail command line option an compile error, should result in an webpack exit code not equal to zero.  

Is there another way or option to check if a compile error occured.

mpseidel:
+1 - this would be really helpful

@pepe79 theoretically you could use the --noEmitOnError typescript option to get an exit code > 0 but then you only get something like this as an error message <code>ModuleBuildError: Module build failed: Error: Typescript emitted no output for .../file.ts</code>

pepe79:
yes, i tried this already before, but this error message is not really useful. 

https://github.com/denvned/webpack-typescript does fail with exit code > 0 on compile error with webpack command line option --bail

But this loader has other issues, so that's not an alternative for us and i would like to go on with ts-loader. But this bug is a blocker for our automated builds.

jbrantly:
Yea, definitely know the need is there. Have you tried `noEmitOnError` (even though there are possibly issues with that - #91)?

jbrantly:
Wow. I didn't get a lot of sleep and missed your comment on `noEmitOnError`. Sorry.

denvned:
@pepe79, I would be happy to resolve any issues you have with **webpack-typescript:** https://github.com/denvned/webpack-typescript/issues

MarkusKramer:
+1 would be really great if this was fixed. as incomplete error messages are obviously a big problem.

EvAlex:
:+1: unbeleivable! That's a blocker issue for using with build server!

EvAlex:
@denvned, your loader has the same issue

EvAlex:
Oh yeah, I've managed to fix it. I will create Pull Request and attach explaination there in 6-8 hours.

raybooysen:
ir @EvAlex @denvned @jbrantly combine to have a single loader, so we don't have all these disparate loaders. :)

jbrantly:
Always open to PRs. FWIW, I think the correct solution to this is to fix #91 to be more useful and then combine that with `--bail` for the webpack exit code, but since I haven't actually dug in and tried it yet I don't know if there are problems with that approach.

vladimir-rovensky:
[...]

PR #143, opened 2016-01-21
Feature/108 exit code

EvAlex:
This provides a workaround for #108. The issue is very simple to reproduce:
1. Add some mess to TypeScript code to get compilation error(s).
2. Run `webpack` with ts-loader
3. Check exit code. It is 0. 

That's a major problem, because build server executing such build will report success when there are errors in TypeScript code.

Workaround is simple and consists of 2 steps:
1. Add plugin that executes on `done` and exits process with non-zero exit code if error(s) present
2. Add logging to ts-loader in order to provide detailed info on compilation error(s)

The solution works fine on my project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ messages are emitted via webpack which is not affected by this flag.
 ##### ignoreDiagnostics *(number[]) (default=[])*
 
 You can squelch certain TypeScript errors by specifying an array of diagnostic
-codes to ignore. 
+codes to ignore.
 
 ##### compiler *(string) (default='typescript')*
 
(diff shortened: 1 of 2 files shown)


jbrantly:
Thank you for taking the time to put together code and submit a PR. However, I'm not sure I want to go down this path quite yet. I still believe the right way to do this is to use `noEmitOnError` (broken right now) and `--bail`. It's possible that `--bail` might not work how it's supposed to (I have some vague recollection of reading about that in a webpack issue) but even if it doesn't I would rather add a new ts-loader config option than document a way to handle process.exit through a plugin.

I'm going to hold off on merging this until I've explored the other approach.

jbrantly:
Closing in favor of #172

Issue #117, opened 2015-12-11
Invalid file references in tsconfig.json lead to misleading ts-loader behaviour


johnnyreilly:
Hey,

Having just burned a good amount of time on something that turned out to be nothing I thought I'd share my learnings.

Consider the following tsconfig.json:

```
{
    "compileOnSave": false,
    "filesGlob": [
        "**/*.{ts,tsx}",
        "../typings/**/*.*.ts"
    ],
    "compilerOptions": {
        "target": "es6",
        "module": "commonjs",
        "noImplicitAny": false,
        "suppressImplicitAnyIndexErrors": true,
        "removeComments": false,
        "preserveConstEnums": true,
        "sourceMap": true
    },
    "files": [
        "i-am-a-file-what-does-not-exist.ts",
        "demo/components/Module.tests.tsx",
        "lib/index.tests.ts",
        "../typings/flux/flux.d.ts",
        "../typings/jasmine/jasmine.d.ts",
        "../typings/node/node.d.ts",
        "../typings/react-router/history.d.ts",
        "../typings/react-router/react-router.d.ts",
        "../typings/react/react-addons-test-utils.d.ts",
        "../typings/react/react-dom.d.ts",
        "../typings/react/react.d.ts",
        "../typings/tsd.d.ts"
    ],
    "exclude": [],
    "atom": {
        "rewriteTsconfig": true
    }
}
```

Spot the invalid file reference?  I didn't either.  For hours.  This is for my tests and Karma hooks up webpack / TS loader.  The following errors were then thrown:

```
11 12 2015 06:14:50.221:ERROR [karma]: { [Error: no such file or directory]
  code: 'ENOENT',
  errno: 34,
  message: 'no such file or directory',
  path: '/_karma_webpack_/test/lib/index.tests.ts' }
Error: no such file or directory
    at MemoryFileSystem.readFileSync (C:\Source\globalize-so-what-cha-want\node_modules\memory-fs\lib\MemoryFileSystem.j
s:107:10)
    at MemoryFileSystem.readFile (C:\Source\globalize-so-what-cha-want\node_modules\memory-fs\lib\MemoryFileSystem.js:29
7:21)
    at doRead (C:\Source\globalize-so-what-cha-want\node_modules\karma-webpack\index.js:156:26)
    at Plugin.readFile (C:\Source\globalize-so-what-cha-want\node_modules\karma-webpack\index.js:160:3)
    at doNTCallback0 (node.js:419:9)
[...]

PR #145, opened 2016-01-24
Make invalid files in tsconfig.json super obvious when ts-loader errors

johnnyreilly:
Hey @jbrantly 

This PR is to mitigate the problems of https://github.com/TypeStrong/ts-loader/issues/117 - invalid files in `tsconfig.json` will be very obvious as a result of the change to error message.  Hope you like it; happy to tweak further.  I've added a test as well.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -297,13 +297,24 @@ function ensureTypeScriptInstance(loaderOptions: LoaderOptions, loader: any): { 
     }
 
     // Load initial files (core lib files, any files specified in tsconfig.json)
-    filesToLoad.forEach(filePath => {
-        filePath = path.normalize(filePath);
-        files[filePath] = {
-            text: fs.readFileSync(filePath, 'utf-8'),
-            version: 0
-        }
-    });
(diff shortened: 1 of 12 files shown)


jbrantly:
Beautiful! Thanks for this. The test is currently failing on TS 1.8 (cause of the changes in TS 1.8 to add "use strict" to output). As a quick note, you can could either update manually or....

```
npm install typescript@next
npm test -- --save-output
```

That will update the test cases with the actual output. Then you just need to review the output using git diff and you're good to go.

As far as actual functionality, :+1:

johnnyreilly:
Thanks chap - I submitted the PR and then turned off my computer (foolishly assuming the build would pass :smile:).  I'll fix and commit when I get a moment.

jbrantly:
Thanks again!

johnnyreilly:
My pleasure!

JKillian:
:+1:  Great change! This was throwing me for a loop today

PR #150, opened 2016-02-02
:memo: prefer exclude

basarat:


basarat:
Read here : https://github.com/TypeStrong/ts-loader/blob/memo/exclude/README.md#configuration If okay please merge and delete :rose:

PR #165, opened 2016-03-02
Implement new option instancePerProject

MortenHoustonLudvigsen:
I have a project with references to node modules containing only pure typescript files and a `tsconfig.json` file in each node module. When I run webpack ts-loader fails to resolve typings (*.d.ts files) in these node modules.
It seems that this is because the `tsconfig.json` file in the main project is used.
To solve this, I have introduced a new option: `instancePerProject`.
If `instancePerProject` is `true` _and_ `instance` has not been set (ie. `instance` === 'default') the instance is set to the path of `tsconfig.json` (actually option `configFileName`) for each file to be compiled.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -28,6 +28,7 @@ function hasOwnProperty(obj, property) {
 interface LoaderOptions {
     silent: boolean;
     instance: string;
+    instancePerProject: boolean;
     compiler: string;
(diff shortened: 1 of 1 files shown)


johnnyreilly:
Closing as stale

Issue #156, opened 2016-02-16
Interface change doesn't trigger change in webpack


AndreyUtka:
For example:

1. clone this starter kit https://github.com/AndreyUtka/angular-typescript-starter-kit
2. run web pack dev server
3. Make some change in interface 

```
export interface IHelloScope {
    displayName: string;
    style: Object;
}
```

->

```
export interface IHelloScope {
    displayName: boolean;
    style: Object;
}
```

In this case, it does not trigger update process. 

The main question is why it does not trigger an update when I made changes only in interfaces? 

Thanks.

kpesanka:
I am experiencing the same issue as well (interface changes not triggering webpack) with a different starter kit:

https://github.com/preboot/angular2-webpack

jbrantly:
@chadaustin reported an issue via twitter that I believe is related: https://twitter.com/chadaustin/status/705542582189715456

Sorry for the delay on this. I'm planning on taking a look this weekend.

johnnyreilly:
I think I may be experiencing this too but I'm not sure.  Certainly I'm breaking things, fixing them and then having to re-kick off my watcher.  No great hardship but it would be nice if I didn't have to do that every now and then.

jbrantly:
There's a test for this kind of thing at https://github.com/TypeStrong/ts-loader/tree/master/test/typeSystemWatch but there must be some case that is being missed.

raybooysen:
Interesting, I see the same thing.  Often the loader will stop seeing
changes to files, especially after you wrote bad code and there were
errors.  I have to restart webpack to get it going again.  Guess it's time
to build a specific test.

AndreyUtka:
I've tried https://www.npmjs.com/package/awesome-typescript-loader  and it handles this case perfectly

jbrantly:
Released in v0.8.2

PR #168, opened 2016-03-05
Create test for issue #156

jbrantly:
This is a test for issue #156. I'm creating this separately because there is some unrelated issue in the test runner that's causing it to fail. I'll update and merge once the test runner issue is fixed.


johnnyreilly:
Due to merge conflicts we couldn't take this as is.  I've resubmitted it as a PR here: https://github.com/TypeStrong/ts-loader/pull/334

PR #169, opened 2016-03-05
Watch for changes on elided imports. (#156)

### LINKED ISSUE #156: Interface change doesn't trigger change in webpack

jbrantly:
Currently elided imports (imports that only contain type information) are not watched by webpack. This was actually a [known issue](https://github.com/TypeStrong/ts-loader/pull/52). This PR fixes so that all imports are explicitly watched.

Resolves #156


AndreyUtka:
Hi @jbrantly ,any progress for it? We are looking forward to this fix)

jbrantly:
Current plan is to cut a release this weekend.

johnnyreilly:
Awesome - this will be super useful on my current project!

AndreyUtka:
thanks a lot)

PR #172, opened 2016-03-19
Update README.md to include fail build details

johnnyreilly:
Added docs as suggested here: https://github.com/TypeStrong/ts-loader/issues/108


johnnyreilly:
PS the build failed last time it ran; I'm presuming it will fail now as well but please note this is strictly a docs change and so should be unrelated.

jbrantly:
Thanks! Been meaning to do this forever ago.

PR #173, opened 2016-03-22
Change compiler Name

davincho:
I don't know if it is just a typo, but in my case a had to change the name of the ts compiler from `ntypescript` to `typescript`, as webpack threw an error stating that `ntypescript` package was not found.
Feel free to close this PR if my changes do not make any sense at all ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,12 +97,12 @@ module.exports = {
   module: {
     loaders: [
       // specify option using query
-      { test: /\.tsx?$/, loader: 'ts-loader?compiler=ntypescript' }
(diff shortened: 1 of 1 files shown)


jbrantly:
This was an example of how you can specify alternative TypeScript compilers using the `compiler` flag. If you simply leave off the `compiler` option then it defaults to `typescript`. I can see how there could be confusion here but I'm not sure this change is the right way to address it. Maybe a different option could be set instead of `compiler`, like `ignoreDiagnostics` or something else.

davincho:
Ok my fault - I did not read the whole `Options` section (especially https://github.com/TypeStrong/ts-loader#compiler-string-defaulttypescript)

PR #194, opened 2016-05-11
update html-webpack-plugin to latest version to fix 2 tests

johnnyreilly:
Also udated associated test results to reflect incremented lodash version.  This should fix 2 of the webpack tests that have been failing for a while.  The ^ meant that the version used was always the latest and when it changed the tests broke.  This will likely happen again at some point by the way; but I guess that's intentional?

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "escape-string-regexp": "^1.0.3",
     "fs-extra": "^0.22.1",
     "glob": "^6.0.3",
-    "html-webpack-plugin": "^2.8.1",
+    "html-webpack-plugin": "^2.16.1",
     "mkdirp": "^0.5.1",
     "mocha": "^2.1.0",
     "rimraf": "^2.4.2",
--- a/test/html-webpack-plugin/expectedOutput-1.6/output.txt
+++ b/test/html-webpack-plugin/expectedOutput-1.6/output.txt
@@ -5,8 +5,8 @@ chunk    {0} bundle.js (main) 22 bytes [rendered]
     [0] ./.test/html-webpack-plugin/app.ts 22 bytes {0} [built]
 Child html-webpack-plugin for "index.html":
          Asset    Size  Chunks       Chunk Names
-    index.html  519 kB       0       
-    chunk    {0} index.html 502 kB [rendered]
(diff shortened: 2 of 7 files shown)


johnnyreilly:
Looks good I think - we only have 5 failing tests instead of 7 :smile:

@jbrantly do you want to make the call on whether you'd like to merge?  BTW it might be an idea to make use of [Greenkeeper](https://greenkeeper.io/).  This should allow us good visibility on when an npm dependendency change is breaking one of our tests.

jbrantly:
:+1: from me

I've had something like this locally for a while but haven't gotten around to pushing. I hope to get back to development on ts-loader soon.

Also :+1: on Greenkeeper

jbrantly:
Oh, also re: intentional. I'm not sure. It's really annoying how some innocuous changes like this cause the tests to fail, but then again sometimes it's an important change that really does break things.

I personally would be OK with pinning the version of something like html-webpack-plugin and then just periodically updating (like with Greenkeeper).

raybooysen:
I strongly suggest the use of a shrinkwrap file to prevent these exact
cases plus saving exact versions in package.json

johnnyreilly:
Awesome - thanks @jbrantly.  I haven't done anything with greenkeeper before but I'll do a little reading and see if I can get it plugged in when I get some time.

> I hope to get back to development on ts-loader soon.

Fab! :smile:

johnnyreilly:
@raybooysen - I haven't looked into shrinkwrap.  Perhaps I will :smile:

raybooysen:
In our current project, we save exact versions in project.json.  So:

dependencies: {
 "foo": "1.0.0"
}

This keeps our top level dependencies fixed, but doesn't help with
dependencies of foo that may use semver.  A shrinkwrap file will fix all
these versions in place.  Which gives you the deterministic npm installs.

johnnyreilly:
`package.json` rather than `project.json` I presume?

raybooysen:
@johnnyreilly correct. :)

johnnyreilly:
What with the recent changes to asp.net core it seems `project.json` has ceased to be a thing 😭

PR #198, opened 2016-05-18
Fix some "error" tests

johnnyreilly:
The only change was in the type of error message used and so it seemed like a safe fix.


johnnyreilly:
didn't work in all cases :cry:

Issue #115, opened 2015-12-04
configFileName can only be in root?


flahertyb:
My use case for `configFileName` is that I have many Webpack apps, some may want to use various typescript files, so I need different tsconfigs for each Webpack app to avoid compiling every .ts file for each. So I was psyched to find this as an option for ts-loader! But this property seems to not accept a path, and seems to not search in child directories for the given filename either. Having a variety of `appname_tsconfig.json` files in the root of the repo is not great. The typescript certainly accepts a path for config on the command line - is there a plan to support this in ts-loader?

jbrantly:
The property doesn't accept paths, that's true. However, the search for a tsconfig file starts at the entrypoint. You can use this to segregate your apps by using directories.

- app1
  - main.ts
  - tsconfig.json
- app2
  - main.ts
  - tsconfig.json

You may also need to use the "instance" option so that each app is separated into its own TypeScript instance, but you may not need to depending on your webpack configuration.

All of that said, ts-loader should have feature parity with `tsc` so we should probably accept a path. There's also a related discussion in https://github.com/TypeStrong/meta/issues/1

flahertyb:
What are your thoughts on being able to specify `files` or `filesGlob` directly in the webpack config? It seems like the most low-friction way for an existing Webpack app to casually start using/depending on a couple ts modules that themselves depend, say, on `React` or `jQuery` typings.

jbrantly:
In an earlier version of ts-loader you could do just that. However, once tsconfig.json was released I removed support for it. tsconfig.json is a much better place for that file list because all of your TS tooling knows how to read it, not just webpack.

flahertyb:
Ah, I see. Well, for sure, tsconfig.json absolutely makes a lot of sense for me and the typescript app I'm writing. atom-typescript, tsc, ts-loader can all use the same thing. 

I would just point out that, to someone else who maintains an es6 webpack app, who knows nothing about typescript, but may want to make use of a library React/Typescript component that I've written - to them, it may seem pretty unhelpful that they need an addition json config file just to tell TypeScript where to find the typings for React. For them, a files array in the webpack config is probably easiest.

Course, I understand the absolute easiest way to start importing ts files, for people who don't know or care about typescript, may not be the goal here. Just pointing out this one perspective.

Thank you for help with everything!

Delagen:
I think it must be rewritten. If all config params are supplied through loader query and config file not found in search path, error occured due path.dirname does not accept `undefined`

[...]

PR #202, opened 2016-05-19
Make ts-loader compatible with node v6

Blechhirn:
Node 6+ enforces the parameter to path.dirname to be a string (https://github.com/nodejs/node/commit/08085c49b6e49ecfffd638c6cd46f4de639ae4f4). The empty string results in the previous behavior.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -250,13 +250,13 @@ function ensureTypeScriptInstance(loaderOptions: LoaderOptions, loader: any): { 
         configParseResult = (<TSCompatibleCompiler><any>compiler).parseJsonConfigFileContent(
             configFile.config,
             compiler.sys,
-            path.dirname(configFilePath)
+            path.dirname(configFilePath || '')
         );
     } else {
         configParseResult = (<TSCompatibleCompiler><any>compiler).parseConfigFile(
             configFile.config,
             compiler.sys,
-            path.dirname(configFilePath)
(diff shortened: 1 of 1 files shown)


johnnyreilly:
I'll leave this to @jbrantly to merge but I wanted to be clear as to the problem this solves.  Does ts-loader not work with node 6?  I haven't tried it myself...

Blechhirn:
That's right - at least in my context where that config file parameter was not used.

Blechhirn:
I have been upgrading several web pack loaders in the last week after switching to node 6 in my projects.   All of them had very similar issues. Btw: seems like some other user also reported this (last comment here: https://github.com/TypeStrong/ts-loader/issues/115).

jhchen:
I've been using ts-loader in [quill](https://github.com/quilljs/quill) and looks like node v6 with ts-loader is [causing build errors](https://github.com/quilljs/quill/issues/732). The code changes in the PR fixes the issue.

johnnyreilly:
I think we should merge this; it's minor and backwards compatible .@jbrantly are you happy with this?

bezreyhan:
any update on this issue?

johnnyreilly:
Merged - sorry about the delay. I'm not sure when @jbrantly next hopes to publish but it'll be in the next release.

Issue #185, opened 2016-04-11
webpack hangs when declaration:true without creating d.ts files


yaroslavya:
I cloned this repo: `https://github.com/adrianhall/example-typescript-webpack`
then `npm install` and if I make `npm run prepublish` it generates the js files normally, but if I change the declaration to `declaration:true` in `tsconfig.json` then `npm run prepublish` just hangs, right after it generates the `.js` files. For now whatever I try the ts-loader doesn't generate the d.ts file for my bundle and always hangs after it generated the js files.

I saw the similar issues were fixed some time ago, like with this pull request: `https://github.com/TypeStrong/ts-loader/pull/175`, but in spite the version was 0.8.2 it doesn't work for me at all

jbrantly:
You said "it generates the .js files" (I assume you mean corresponding .js files for each .ts file, not the final bundle) which seems strange to me. Typically when using ts-loader you don't want those files and ts-loader itself doesn't generate them. So here's my theory of what is happening to you:

1. You're using a TypeScript aware IDE where when you press save it generates all of the .js files for each .ts file
2. The example project you linked has the ordering of the extensions wrong. It has `[ '', '.js', '.ts', '.jsx', '.tsx' ]` which means that .js files will be resolved before .ts, so ts-loader is trying to compile the .js files instead.

First you need to fix the ordering of the extenions (.ts stuff should go before .js stuff). Secondly, if you don't want the intermediate .js files from your IDE you can set `noEmit` to true in your tsconfig.json and then [turn it back on for ts-loader specifically using the webpack config override](https://github.com/TypeStrong/ts-loader/issues/161#issuecomment-188906356).

yaroslavya:
I tried the following config settings:
tsconfig.json:
   `"declaration": true,
    "module": "commonjs",
    "noImplicitAny": true,
    "preserveConstEnums": true,
    "removeComments": false,
    "sourceMap": true,
    "target": "es5",
    "noEmit": true`

webpack.config.js:
`module: {
    loaders: [
      { test: /\.tsx?$/, loader: 'ts', exclude: /node_modules/ }
    ]
  }`
`resolve: {
    root: path.resolve('./src'),
    extensions: ['.ts', '.js', '']
  }`
`ts:{
    compilerOptions:{
      noEmit: false
    }
  }`

If I change the declaration to false in tsconfig.json it works, but of course doesn't produce a d.ts file.
[...]

PR #214, opened 2016-05-27
Use a relative path for the declaration assets

17cupsofcoffee:
This is intended as a fix for issue #185 - a full explanation of what the problem was can be found in [this comment](https://github.com/TypeStrong/ts-loader/issues/185#issuecomment-222118959), but what it boiled down to is that the language service returns absolute paths for generated declaration files, and these were being used directly as keys in `compilation.assets`, causing Webpack to hang during the emit stage. Resolving the path relative to the current working directory beforehand fixes this issue. According to @t246246's comment on the issue, this only affects Windows, but as I don't have access to a Linux or OSX device, I can't confirm that.

One caveat to this fix is that the output file structure for the declaration files follows that of the source code - i.e, if the source looks like this:

```
root/
    src/
        dir1/
        dir2/
```

Then the declarations get output like so:

```
root/
    dist/
        src/
            dir1/
            dir2/
    src/
        dir1/
        dir2/
```

This might well be a breaking change or not be the 'correct' way of handling it - I can't tell, as this functionality never worked for me in the first place! There's a [currently pending pull request](https://github.com/s-panferov/awesome-typescript-loader/pull/136) on awesome-typescript-loader's repository that demonstrates a possible way of making it follow Webpack's context instead - if the current way files are output isn't acceptable, I can have a go at implementing something along those lines here instead? Let me know, I can make some more changes if needed. This should be a good start though, at the very least.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -496,7 +496,8 @@ function ensureTypeScriptInstance(loaderOptions: LoaderOptions, loader: any): { 
                 let output = languageService.getEmitOutput(filePath);
                 let declarationFile = output.outputFiles.filter(filePath => !!filePath.name.match(/\.d.ts$/)).pop();
                 if (declarationFile) {
-                    compilation.assets[declarationFile.name] = {
+                    let assetPath = path.relative(process.cwd(), declarationFile.name);
+                    compilation.assets[assetPath] = {
                         source: () => declarationFile.text,
                         size: () => declarationFile.text.length
                     };


17cupsofcoffee:
Ah, this doesn't seem to be playing nicely with the `declarationOutput` test (probably because the working directory is the same when running the test suite?). Might need to look into that.

lbar:
Hi. Is there some news about this PR?

17cupsofcoffee:
@lbar: Nothing on my end, I'm holding off on working any further on it until someone more experienced lets me know if I'm going in the right direction :p

lbar:
OK thank you for your answer :)

flyon:
@jbrantly @johnnyreilly or any other maintainer, can you please look at this PR fixing #185, people are leaving to awesome-typescript-loader because of this issue, and I'd hate to join them!

johnnyreilly:
Okay - running the test just hangs on windows.  Manually testing the PR does change the emit path but it also allows the test to run; this is a better outcome that at present in my book.

johnnyreilly:
Okay - I've pondered it; I can't see any reason not to merge.  @jbrantly if there's any objections then please do pitch in; it's a 1 line change and can easily be amended if there's abetter approach possible.  But since this is not a working thing (at least on Windows) then I think it's probably worth going with this for now.

Just as a by the by I put some `console.log`s in to test various things out:

```
                    console.log("declarationFile", JSON.stringify(declarationFile))
                    console.log("process.cwd()", process.cwd())
                    let assetPath = path.relative(process.cwd(), declarationFile.name);
                    let assetPath2 = path.normalize(declarationFile.name);
                    console.log("assetPath2", assetPath2)
                    console.log("assetPath", assetPath)
```


```
  declarationOutput
declarationFile {"name":"C:/source/ts-loader/.test/declarationOutput/app.d.ts","text":"import dep = require('./sub/dep')
;\ndeclare class Test extends dep {\n    doSomething(): void;\n}\nexport = Test;\n"}
process.cwd() C:\source\ts-loader
assetPath2 C:\source\ts-loader\.test\declarationOutput\app.d.ts
assetPath .test\declarationOutput\app.d.ts

declarationFile {"name":"C:/source/ts-loader/.test/declarationOutput/sub/dep.d.ts","text":"declare class Test {\n    doSomething(): void;\n}\nexport = Test;\n"}
process.cwd() C:\source\ts-loader
assetPath2 C:\source\ts-loader\.test\declarationOutput\sub\dep.d.ts
assetPath .test\declarationOutput\sub\dep.d.ts
```

Do notice the difference in paths "/" vs "\".  Just didn't want to lose this information.

johnnyreilly:
Okay merged.  Now to try and get us back to a "passing tests state" 😄

dizel3d:
This fix also gave us a broken change. d.ts files paths should be related to webpack context, not to current process workdir. They are not always the same. My webpack config looks like:
```
module.exports = {
    context: path.join(__dirname, 'src')
    ...
}
```

johnnyreilly:
Hi @dizel3d,

Thanks for reporting; PRs to improve the behaviour would be gratefully received.  Over to you 👍

dizel3d:
@johnnyreilly, I did it, see #307

17cupsofcoffee:
@dizel3d - Apologies for the breaking PR, this was my first one outside of updating docs and I wasn't too familiar with what the expected behavior was! Thanks for improving it <3

johnnyreilly:
@17cupsofcoffee - don't let it frighten you off submitting more PRs!

PR #220, opened 2016-05-31
minor fix to test

johnnyreilly:
maybe if I do this I'll tempt @jbrantly back :smile:


johnnyreilly:
Only 5 failing tests in the typescript@next build now instead of 7.

PR #221, opened 2016-05-31
Fix html-webpack-plugin tests

johnnyreilly:
Fix html-webpack-plugin tests


johnnyreilly:
Down to 3 failing tests for the highest 3 builds (1.7, 1.8 and next)

Issue #255, opened 2016-08-02
Error Checks block increases incremental build time


andyzg:
During `ts-loader`'s `after-plugin` callback, the error checks take a ton of time per incremental build. In the code base I'm currently working with, error checks suck up over 6 seconds of incremental build  time.

All typescript files by default are currently checked for errors. I'm proposing that we add a regex to exclude files solely for error checks for incremental build time optimization.

My current build time went from 8s to 2s on average through a prototype of this change.

andyzg:
Duplicate of #78

PR #256, opened 2016-08-02
Enable loader to exclude files matching regex for error checks

andyzg:
Error checks are a huge bottleneck to the ts-loader.  Refer to #255


andyzg:
I've just realized after digging deeper that this ties into issue #78 . My current alternative is to mute errors from `ts-loader` and run `tsc --noEmit` to check for any errors, which brought my incremental build time from 9s to 1.5s

andyzg:
Closing this. Realized `transpileOnly: true` does the exact same.

Issue #268, opened 2016-08-22
multiple instances, errors only shown of single instance


DatenMetzgerX:
Hy

I'm working on a project that targets node and the browser (using web workers). The project consists of a code base that is shared between node and the browser, and other parts that are browser, node or even web worker specific. To benefit the most from type script, multiple type script instances should be used with different libraries included (only es5, es5+dom, es5+web worker...).

**Setup**
I created a sample project with a similar structure in [DatenMetzgerX/ts-loader-multiple-instances](https://github.com/DatenMetzgerX/ts-loader-multiple-instances). Files located in src/common are es5 only and can be used by node or in the browser. At the other hand, files located in src/browser are specific for the browser environment. Different tsconfig.json files are used for each directory to configure the different libraries (es only, es + dom...).  In each directory (common and browser) a single file is located that contains some trivial typescript errors (browser-defaults: a is a string, but a number is assigned, simple.js: n is a number, but a string is assigned).

The webpack configuration contains 2 configured ts-loader instances pointing todifferent tsconfig.json's and are distinguished by their name. 

**Actual Result**
If webpack is executed, only errors of one of the ts-loader instances are shown in detail: 

```
[micha@mbp ~/git/ts-loader-mutliple-instances (master)]$ webpack
ts-loader: Using typescript@2.0.0 and /Users/micha/git/ts-loader-mutliple-instances/src/common/tsconfig.json
ts-loader: Using typescript@2.0.0 and /Users/micha/git/ts-loader-mutliple-instances/src/browser/tsconfig.json
Hash: 98e778fd3041495f122c
Version: webpack 1.13.2
Time: 1798ms
         Asset     Size  Chunks             Chunk Names
parallel-es.js  4.78 kB       0  [emitted]  main
   [0] ./src/common/simple.ts 154 bytes {0} [built] [3 errors]
   [1] ./src/browser/browser-defaults.ts 123 bytes {0} [built] [2 errors]

ERROR in ./src/browser/browser-defaults.ts
(5,7): error TS2322: Type 'number' is not assignable to type 'string'.

ERROR in ./src/browser/browser-defaults.ts
(5,7): error TS6133: 'a' is declared but never used.
```

But the output also shows... that simple.ts contains 3 errors, but the error details are not printed to the console (why so ever).

**Expected Result**
Ts-loader should also output the errors in simple.ts in detail and not only summarize the number of errors behind the asset. 

```
[...]

PR #269, opened 2016-08-22
Errors of multiple instances

DatenMetzgerX:
See #268 for more details


johnnyreilly:
I'm afraid that merging some other PRs has left this PR in a state which means we can't take this as it stands.  We are quite interested in this functionality though.  Would you be willing to resubmit (and also supply a test to cover this as well)?  🌷

PR #274, opened 2016-09-02
Add link to typescript docs on .d.ts files

hawkrives:
To save people from the extreme effort of Googling for it 😉

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ of your code.
 ### Loading other resources and code splitting
 
 Loading css and other resources is possible but you will need to make sure that
-you have defined the `require` function in a declaration file.
+you have defined the `require` function in a [declaration file](https://www.typescriptlang.org/docs/handbook/writing-declaration-files.html).
 
 ```typescript
 declare var require: {


hawkrives:
Maybe nevermind; those docs aren't terribly helpful for this case.

johnnyreilly:
Thanks chap!  I wonder if it would be worth providing an example of what that definition might look like as well?  Something like this:

```ts
declare var require: (filename: string) => any;
```

hawkrives:
I'm more looking for an example of where to put the declaration so it's available to all the typescript files? Flow has the `flow-typed` folder for project-scope definitions, but I'm not sure where Typescript's equivalent is.

johnnyreilly:
As long as it sits within the same folder structure as your tsconfig.json it will be detected.

hawkrives:
I do intend to update this PR, just as soon as I figure out how convince my project to accept the example.

Ah, I figured out my problem. I'm using the DefinitelyTyped `require` already, and I'm not sure how to "extend" that definition with `require.ensure`. Should I just replace it with the sample from the docs (mostly included below)?

(I'm sorry that we're getting off the original topic of this PR, but I can submit a PR for docs on how to do this if I can ever figure it out.)

```ts
declare var require: {
    <T>(path: string): T
    (paths: string[], callback: (...modules: any[]) => void): void
    ensure: (paths: string[], callback: (require: <T>(path: string) => T) => void, chunkName?: string) => void
}
```

hawkrives:
So after finding [this SO answer](http://stackoverflow.com/questions/12719529/using-jquery-plugin-in-typescript), I came up with this method of extending the DefinitelyTyped `require` definition:

```ts
interface Require {
    ensure: (paths: string[], callback: (require: <T>(path: string) => T) => void, chunkName?: string) => void
}
```

It seems to work! Are there any major problems with it? If not, then I might submit a PR to rework that chunk of the README.

johnnyreilly:
I think that the require you are referring to is related to RequireJS.  Despite the similar names it's a completely different thing.  Webpack's require is commonjs and requirejs is amd...

hawkrives:
you know, that makes a lot of sense. Well, I've successfully mixed commonjs and requirejs types, then! time to get rid of the requirejs ones. Thanks!

PR #293, opened 2016-10-05
Fail fast if expectedOutput does not exist

johnnyreilly:
I realised that the tests were failing because there is no expected output.  I made the tests check at the beginning if there is an output to assert against and fail immediately if not.  I hope this makes the test pack faster and the reason for failure more obvious.

We should probably get rid of the 1.9 expected output by the way; there is no TS 1.9 and will not be.  We should likely adjust the build matrices to remove typescript@next since we don't have valid output for that either.


johnnyreilly:
Believe me I've been putting this off hoping the problem would fix itself. It didn't :smile: 

Using WebPack at work so I need it to work

PR #294, opened 2016-10-06
Ensure test data can be regenerated 

johnnyreilly:
when the test data doesn't exist in the first place and is being regenerated with `--save-output`

Also enable the running of a single test using `npm test -- --single-test testname` (documented in `CONTRIBUTING.md`)


johnnyreilly:
i'll try again

PR #318, opened 2016-10-16
Add types support

johnnyreilly:
Added support for @types resolution.  This is covered by a new test called `2.0.3_typesResolution`.  This test only runs for the 2.x + versions of TypeScript.  It is an execution test and relies upon @types/jasmine.  It fails if the resolution of @types/jasmine fails. 

Also: 
Moved node_modules to root dir in tests


johnnyreilly:
PS @basarat - wondering if you spot anything in `index.ts` :smile:

basarat:
🌹

PR #356, opened 2016-11-03
Add fileName as prefix to error message to support Visual Studio

gamli:
This allows Visual Studio to search the output of ts-loader for errors and show them in its error list.
See https://msdn.microsoft.com/en-us/library/yxkt8b26.aspx for a description of the format.
The path is normalized so the platform's path separator is used (Visual Studio doesn't handle those forward slashes very well - i.e. jumping to the line number does not work).

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,10 @@ set to the NPM name of the compiler, eg [`ntypescript`](https://github.com/basar
 
 Allows you to specify a custom configuration file.
 
+##### visualStudioErrorFormat *(boolean) (default=false)*
+
+If `true`, the TypeScript compiler output for an error or a warning, e.g. `(3,14): error TS4711: you did something very wrong`, in file `myFile` will instead be `myFile(3,14): error TS4711: you did something very wrong` (notice the file name at the beginning). This way Visual Studio will interpret this line and show any errors or warnings in the *error list*. This enables navigation to the file/line/column through double click.
+
 ##### compilerOptions *(object) (default={})*
 
 Allows overriding TypeScript options. Should be specified in the same format
--- a/src/index.ts
+++ b/src/index.ts
@@ -57,6 +57,7 @@ function makeOptions(loader: interfaces.Webpack) {
         compiler: 'typescript',
         configFileName: 'tsconfig.json',
         transpileOnly: false,
+        visualStudioErrorFormat: false,
         compilerOptions: {},
     }, configFileOptions, queryOptions);
     options.ignoreDiagnostics = arrify(options.ignoreDiagnostics).map(Number);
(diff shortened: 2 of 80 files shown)


jbrantly:
I wonder if this should be behind a flag (not advocating that it should be, just thinking out loud). The existing format is really meant to be compatible with other webpack output, and having the file name twice seems kind of repetitive. On the other hand, having the output display in Visual Studio seems useful for those who use it.

gamli:
Yes, this would be better if someone attempts to parse the exact output. But I am not sure how the tests could be modified to fit this behaviour.

johnnyreilly:
Ah - yes I think I misunderstood your original issue. I'd say that I agree with @jbrantly that if this goes in it should definitely behind a flag as for the general use case as we'd want ts-loader to remain consistent with WebPack.

Would you be willing to amend your PR to put this behind a flag called something like `visualStudioErrorFormat` or something? We'd want a single new test which used this new flag and thus expected error text in the VS format. I'd advocate cloning one of the existing error tests.

johnnyreilly:
If you're after a good candidate `aliasResolution` would probably do the job if you removed the `patch0` and `commonComponents` subdirectories.

gamli:
Yes sure - I will be making that change tomorrow morning. It feels a lot better to just add a test/feature than breaking existsing stuff.

Just for giggles:
Today I spent about 2 hours changing the 63 files with expected output by hand - struggling with invisible ESC characters for changing colors and lack of knowledge of the inner workings of npm colors.
That was, until I realized that there is the option "--save-output" :)

johnnyreilly:
> Today I spent about 2 hours changing the 63 files with expected output by hand - struggling with invisible ESC characters for changing colors and lack of knowledge of the inner workings of npm colors.
That was, until I realized that there is the option "--save-output" :)

OUCH 😄

gamli:
Sorry for the delay, but I got sick on friday.

I implemented the changes:

* Turn on/off Visual Studio error output with the flag `visualStudioErrorFormat`
* Add test `visualStudioErrorFormat` (based on `aliasResolution` but without patch0)

I am not quite sure if the tests with `output.transpiled.txt` are necessary.

johnnyreilly:
Cool - could you update the documentation in the readme with the new option please?

gamli:
Done.

johnnyreilly:
Thanks!

gamli:
One more question: is it possible to reference the latest version through npm or do i have to reference my fork?

johnnyreilly:
You can reference ts-loader on GitHub now your fork has been merged in. It's not been published to npm yet.

gamli:
Ok, thanks!

Issue #376, opened 2016-11-15
Re-exported const enums break emit in watch mode


smphhh:
Given

bar.ts:
```typescript
export const enum BarEnum {
    Bar = 1
};
```
foo.ts:
```typescript
export { BarEnum } from './bar';
```
app.ts:
```typescript
import { BarEnum } from './foo';
console.log(BarEnum.Bar);
```
the emit for app.ts in the initial build looks like the following:
```javascript
"use strict";
console.log(1 /* Bar */);
```
After changing bar.ts to
```typescript
export const enum BarEnum {
    Bar = 2
};
```
no modules will be rebuild. The expected output after the change would be
```javascript
"use strict";
console.log(2 /* Bar */);
```

This is because ts-loader currently only adds the direct dependencies of modules as dependencies to webpack. Because bar.ts is not a direct dependency of app.ts, app.ts will not be rebuild when bar.ts changes.

For semantic errors the deeper dependencies are handled in the after-compile hook, but this doesn't help in cases like const enums where the actual emit can depend on deeper dependencies.

I believe awesome-typescript-loader does add all actual dependencies of a module as webpack dependencies recursively, which I think is the only way to handle these (relatively rare) cases correctly.

I'll post a pull-request with a test showcasing this error.

johnnyreilly:
Thanks for the issue and the PR illustrating the issue. I think you're right - it's a problem. If you get the time to take a look at this that'd be awesome. I'm not sure when I'd get time to look myself.

smphhh:
I can probably at least play around with this.

johnnyreilly:
Awesome - every little helps :wink:

johnnyreilly:
Fixed with https://github.com/TypeStrong/ts-loader/pull/377 - thanks @smphhh!

PR #377, opened 2016-11-15
Add test for handling re-exported const enums (issue #376)

smphhh:
Add a test for issue #376.

I'm not sure if the transpile-mode expected output is correct, it looks like const enums are always preserved in that mode.


johnnyreilly:
This looks really good!  Will take a proper look at this soon.  Thanks!

johnnyreilly:
This is awesome! Thanks!

Issue #379, opened 2016-11-16
typescript.sys should be compiler.sys


johnnyreilly:
Taken from @jbrantly comments here: https://github.com/TypeStrong/ts-loader/pull/362

> Regarding the current use of typescript.sys in servicesHost, I think it should actually be compiler.sys. It doesn't look like typescript.sys was accessed in the original code.

PR #380, opened 2016-11-16
typescript.sys should be compiler.sys

johnnyreilly:
https://github.com/TypeStrong/ts-loader/issues/379

Made change to `servicesHost.ts`.  Raising PR now to trigger CI builds and see which tests are failing - will address those failures in a later commit.


johnnyreilly:
It looks like the test data doesn't need regenerating; will probably merge this later on.

Issue #388, opened 2016-11-21
Cannot use ag-grid with ts-loader >= 0.9.3


Wykks:
When using ag-grid and ts-loader >= 0.9.3 with allowJs set to true, I get lots of error like this : 
```
Module '"[...]/node_modules/ag-grid/main"' has no exported member '[...]'.
```
It's fine with ts-loader 0.9.2

Maybe it's related to this issue: https://github.com/TypeStrong/ts-loader/issues/250
Since ag-grid use the typings field as well https://github.com/ceolter/ag-grid/blob/master/package.json#L9

johnnyreilly:
Since 0.9.3 is when we added support for `allowJs` I would guess it's related to that.  Have you tried setting `allowJs` to `false` in your `tsconfig.json`?

Wykks:
Unfortunately, I can't do that, because my project is too big, and rely on this option for now.

johnnyreilly:
Do you want to try something for me?  If you tweak [this line](https://github.com/TypeStrong/ts-loader/blob/5359d6334746e8374c8f559d0d814ef4b604ae82/src/instances.ts#L92):

from:

```ts
    // if allowJs is set then we should accept js(x) files
    const scriptRegex = configFile.config.compilerOptions.allowJs
        ? /\.tsx?$|\.jsx?$/i
        : /\.tsx?$/i;
```

to: 

```ts
    const scriptRegex = /\.tsx?$/i;
```

does it resolve your issue?

Wykks:
Yes, it works fine with this change (using v1.2.2)

johnnyreilly:
OK.  So that's a little frustrating.  Essentially, when `allowJs` is set to true we allow both both `js(x)` and `ts(x)` files to be processed. For most people this works in the way they would hope.  Not in your case though; I'm not sure why.

I'd possibly be open to supporting this behaviour in some way in ts-loader but there's a couple of things we'd need first:

1. A **simple** repro of the issue.  That repro should be turned into an integration test that becomes part of our testpack.  With the tweak in place the test passes, without it, it fails.  
2. We then need to think of how this is wisely implemented.  It would be possible to just stick this behind a flag and have done but I'm not keen.  I want to understand what the issue is.  I'd really rather have `allowJs` just work.  If we need to add an extra flag to cater for this we need to have a good understanding of why and, crucially, a clear name for it so other people picking over the issue will "fall into the pit of success".

Over to you!

[...]

PR #397, opened 2016-12-02
Test for allowJS issue (see issue #388)

Wykks:


johnnyreilly:
Thanks for this!  Is there a need for the `npm link`?

Wykks:
I just figured out that no lib is needed to reproduce the issue.

Test updated !

PR #412, opened 2016-12-11
minor perf optimisation for loaderoptions

johnnyreilly:
new node type definitions
started documenting the codebase some more


johnnyreilly:
Will revert `test` -> `match` changes; actually seem to be less performant with later versions of v8 http://stackoverflow.com/a/10940138/761388

PR #415, opened 2016-12-14
Fix bug when "extend"ing a tsconfig that specifies "allowJs"

cspotcode:
When ts-loader checks the `allowJs` flag of your tsconfig, it should be checking the fully loaded config, including all "extend"ed tsconfig files.

I also fixed a spelling mistake in the README that I noticed while working.


johnnyreilly:
Thanks for the PR! Seems reasonable - assuming tests pass I'll merge. Thanks again!

PR #475, opened 2017-02-21
move to using loader-utils 1.0.0

johnnyreilly:


johnnyreilly:
Issues with loader-utils discussed here: https://github.com/webpack/loader-utils/issues/56

PR #480, opened 2017-02-22
Protect from the mtimes property being undefined.

mredbishop:


johnnyreilly:
Great minds right? https://github.com/TypeStrong/ts-loader/pull/481

Thanks for this - will merge assuming tests pass. It looks like something has changed in webpack. I don't know what though.

johnnyreilly:
Okay this was a breaking change in webpack. ts-loader depends on an undocumented part of the API. That is being added back to webpack until we  can better work around this.

mredbishop:
Yep 🙂
at-loader has the same issue...

johnnyreilly:
https://github.com/webpack/watchpack/pull/48

johnnyreilly:
Thanks for your help - just released: https://github.com/TypeStrong/ts-loader/releases/tag/v2.0.1

mredbishop:
Great, thanks John!

PR #500, opened 2017-03-14
Track timestamps to filter changed files.

WearyMonkey:
`getTimes()` returns a map with all files. This breaks version tracking as all file versions are incremented.

This PR adds manual timestamp tracking to filter out actually changed files.


johnnyreilly:
Thanks! Let's see how the tests do.  Also @jbrantly would you be able to comment on this?  Seems like a reasonable change to me.

@WearyMonkey I'm probably about to go offline - ping me tomorrow if it looks like I've forgotten about this.

johnnyreilly:
Okay the failing tests are `issue441` and `importsWatch` - i think the failures are reasonable and expected given the changed behaviour this will introduce. Do you want to regenerate the 2.2 test output for these tests? You might need to delete the existing output manually before you do this; it doesn't always clean up entirely. If memory serves you'll need to do something like `npm run comparison-tests -- --save-output --single-test issue441` etc

johnnyreilly:
Just had a go at regenerating the test data myself.  I didn't realise I could work directly on a supplied PR before I'd landed it.  Pleased to see I can :smile:

johnnyreilly:
Merged - will look to cut a release tonight if possible.

PR #524, opened 2017-04-19
fix webpack example to include '-loader'

jraller:
Updated for latest webpack syntax. leaving off the '-loader' throws an error


johnnyreilly:
Thanks!

Issue #537, opened 2017-05-06
Faster type checking with forked process


piotr-oles:
Hi everyone!
I've created webpack plugin: [fork-ts-checker-webpack-plugin](https://github.com/Realytics/fork-ts-checker-webpack-plugin) that plays nicely with `ts-loader`. The idea is to compile project with `transpileOnly: true` and check types on separate process (async). With this approach, webpack build is not blocked by type checker and we have semantic check with fast incremental build. More info on github repo :)
So if you like it and you think it would be good to add some info in README.md about this plugin, I would be greatful.
Thanks :)

johnnyreilly:
Thanks for letting us know! This sounds really useful / interesting. Assuming there's no gotchas then I think adding documentation about this to ts-loader could be a good idea.

When I get some time I'd like to have a play with this and reply back to you. Thanks!

johnnyreilly:
Would you  be interested in submitting a PR which contains:

1. An update to the readme mentioning the fork-ts-checker-webpack-plugin and
2. An example setup to live here: https://github.com/TypeStrong/ts-loader/tree/master/examples

?

I think this could be useful to people. Thanks for your work!

piotr-oles:
Of course, I will submit it soon ;)

johnnyreilly:
Merged.  Thanks!

PR #540, opened 2017-05-11
Add fork-ts-checker-webpack-plugin info and examples

piotr-oles:
As a response to [TypeStrong/ts-loader#537](https://github.com/TypeStrong/ts-loader/issues/537) - adds informations about fork-ts-checker-webpack-plugin and one example of usage.


johnnyreilly:
Thanks!

PR #547, opened 2017-05-21
Fixes #336. Add happypack compatibility mode

### LINKED ISSUE #336: HappyPack with ts-loader.
I'm trying to use HappyPack with ts-loader, and it seems to be working fine in `transpileOnly` mode if I remove lines:

https://github.com/TypeStrong/ts-loader/blob/v0.9.5/index.ts#L346
https://github.com/TypeStrong/ts-loader/blob/v0.9.5/index.ts#L718
https://github.com/TypeStrong/ts-loader/blob/v0.9.5/index.ts#L763

For me it is fine to remove L346 and L718 because I have tsc running in the background in typecheck mode. I wonder what are the implication of L763.

Also do you think it is possible to do some change to make it work out of the box without local patching of ts-loader.

see also https://github.com/amireh/happypack/issues/33

aindlq:


johnnyreilly:
This is brilliant! Thanks​! Do you know why the test is failing?

aindlq:
forgot to add happypack as a dependency, should be ok with the next run

johnnyreilly:
Cool - thanks!  Just doing some tweaks

johnnyreilly:
This looks good - let's merge!

Issue #568, opened 2017-06-29
2.2.1  = npm error "Could not find file:" ...


timothyallan:
Just started getting this npm build error with 2.2.1, not a ton of other details I can find yet.

```
./node_modules/typescript/lib/typescript.js:89078
throw new Error("Could not find file: '" + fileName + "'.");
```
using it like this
```
 loaders: [{
      test: /\.ts(x?)$/,
      exclude: /node_modules/,
      use: [
        {
          loader: 'babel-loader',
          options: {
            presets: [['es2015'], 'stage-0', 'react'].concat(DEBUG ? ['react-hmre'] : []),
          },
        },
        {
          loader: 'ts-loader',
        },
      ],
    },
```
If I revert to 2.2.0, everything works fine. 
Using Typescript 2.4.1, NPM 5.0.3 on macOs

johnnyreilly:
That's strange... Our builds are passing. Do you have any more details?

johnnyreilly:
Would you be able to test the build I created here?: https://github.com/johnnyreilly/ts-loader

It contains a speculative fix for your issue...

timothyallan:
Hey John, sorry no go on that one. Just compiled it, dropped it in and am getting this:
```
/Users/ta/development/src/node_modules/ts-loader/node_modules/typescript/lib/typescript.js:89078
                throw new Error("Could not find file: '" + fileName + "'.");
                ^

Error: Could not find file: '/Users/ta/development/src/resources/assets/js/app.js'.
    at getValidSourceFile (/Users/ta/development/src/node_modules/ts-loader/node_modules/typescript/lib/typescript.js:89078:23)
    at Object.getSyntacticDiagnostics (/Users/ta/development/src/node_modules/ts-loader/node_modules/typescript/lib/typescript.js:89291:52)
    at /Users/ta/development/src/node_modules/ts-loader/dist/after-compile.js:91:38
    at Array.forEach (native)
    at provideErrorsToWebpack (/Users/ta/development/src/node_modules/ts-loader/dist/after-compile.js:90:10)
    at Compiler.<anonymous> (/Users/ta/development/src/node_modules/ts-loader/dist/after-compile.js:21:9)
    at next (/Users/ta/development/src/node_modules/webpack/node_modules/tapable/lib/Tapable.js:140:14)
    at Compiler.<anonymous> (/Users/ta/development/src/node_modules/webpack/lib/CachePlugin.js:62:5)
[...]

PR #569, opened 2017-06-30
Speculative fix for #568

johnnyreilly:


johnnyreilly:
@donaldpipowitch I've included the change we discussed about dropping this piece of code:

```js
        // do any necessary config massaging		
        if (loaderOptions.transpileOnly) {		
            configFile.config.compilerOptions.isolatedModules = true;		
        }
```

This can still be set via `tsconfig.json` if desired.  It just won't be set by default anymore (since it's not useful with fork-ts-checker-webpack-plugin in the mix anyway)

donaldpipowitch:
Thank you for pinging 👍

PR #579, opened 2017-07-06
refactor: Use chalk instead of colors

### LINKED ISSUE #62: Replace `colors` usage with `chalk`
We can get rid of those globals.

develar:
* to reduce dependency tree, because most of the modern packages uses chalk, not colors
* Doesn't extend String.prototype and https://github.com/chalk/chalk#highlights

Close #62

Please note — `yarn.lock` is added since Yarn is much better than npm in terms of speed&reliability (#369). 
In general, I recommend to remove `package.lock` from this repo and use only yarn, since no reasons to use npm, but anyway for now this PR contains modified package.lock (because it is npm, file changes is very big due to npm 5.1 release).


develar:
CI failures are not related to my changes, as far I see.

johnnyreilly:
Hi @develar,

First of all, thanks for sending this!  I won't have be able to look at this immediately but I will get to it.  

The CI failures are related to your change.  Essentially the comparison test pack is kind of brutal and literally compares output to output.  The result of your changes has changed the line number of stack traces; see below:

```shell
  1) nodeModulesMeaningfulErrorWhenImportingTs should have the correct output:
      Uncaught AssertionError: output.txt is different between actual and expected
      + expected - actual
       Module build failed: Error: Typescript emitted no output for node_modules/a/index.ts.
       You should not need to recompile .ts files in node_modules.
       Please contact the package author to advise them to use --declaration --outDir.
       More https://github.com/Microsoft/TypeScript/issues/12358
      -    at Object.loader (dist/index.js:31:15)
      +    at Object.loader (dist/index.js:32:15)
```

The line number is now `32` rather than `31`.  Arguably the test pack should ignore this; it's not that relevant.  We ignore other differences such as number of bytes etc.  

If you'd like to either regenerate the output for those 3 tests / patch it for new line numbers or make the comparison test pack ignore the line numbers in the stack trace then we'll be green.

johnnyreilly:
Hi @develar,

I did some work at the weekend to get the tests passing.  I'm pretty happy with this; I just want to do some local experimentation and then I'll look to merge.  Thanks again!

develar:
@johnnyreilly Thanks for you work, sorry, that I didn't start to fix it on weekend, was busy on another open source project.

johnnyreilly:
Merging - thanks for your work! This will go out with the next release of ts-loader.

PR #587, opened 2017-07-17
Use compiler.fileTimestamps cause watcher may be null

### LINKED ISSUE #585: Undefined watcher in watch-run causes error...
Might not be a ts-loader issue but I thought it might be worth linking it until I figure out the cause:  https://github.com/zinserjan/mocha-webpack/issues/159

zinserjan:
fixes #585

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v2.3.1 - NOT RELEASED YET
+
+- [Fix undefined watcher in watch-run causes error](https://github.com/TypeStrong/ts-loader/pull/587) (#585) - thanks @zinserjan and @sokra! 
+
 ## v2.3.0
 
 - [add appendTsxSuffixTo option to support using tsx with Vue](https://github.com/TypeStrong/ts-loader/pull/581) - lots of discussion went into this PR. Thanks to @vhqtvn (author) and @HerringtonDarkholme, @johnnyreilly, @jbrantly, @octref, @rhyek and others for helping us land on our final implementation.
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -127,6 +127,7 @@ export interface WebpackCompiler {
     isChild(): boolean;
     context: string; // a guess
     watchFileSystem: WebpackNodeWatchFileSystem;
+    fileTimestamps: {[key: string]: number};
 }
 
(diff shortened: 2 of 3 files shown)


johnnyreilly:
Thanks for contributing! Do you know if this an official part of the API? I ask as we've had that happen before; see here: 

https://blog.johnnyreilly.com/2017/02/under-duck-afternoon-in-open-source.html

Could you check please? @sokra may know..

johnnyreilly:
Does `fileTimestamps` produce the same value as `getTimes`? Or does it contain more / less? To what extent? I'm considering the merits of having `compilerOptions` in there as a fallback

zinserjan:
> Does fileTimestamps produce the same value as getTimes?  Or does it contain more / less? 

It's atm the same, but it depends on the implementation of the watcher. The results of watcher.getTimes() are stored in fileTimestamps. See [here](https://github.com/webpack/webpack/blob/master/lib/node/NodeWatchFileSystem.js#L44) and [here](https://github.com/webpack/webpack/blob/master/lib/Compiler.js#L120).

In case of the WatchIgnorePlugin it contains more data, as WatchIgnorePlugin [enhances the timestamps with all ignored files](https://github.com/webpack/webpack/blob/master/lib/WatchIgnorePlugin.js#L38-L46). 

Performance isn't really a topic.
* without WatchIgnorePlugin it should be faster but this in the lower ms range.
* with WatchIgnorePlugin I would expect that it's a bit slower, cause fileTimestamps is bigger but maybe the effect of not collecting the data compensates this..

johnnyreilly:
Thanks @zinserjan - that's really helpful.  I've been asking in the webpack slack channel to see if there's any reason not to go with this.  No-one has advised that's the case so far.  If no-one has given a compelling reason why we shouldn't take this in the next 24 hours then I reckon we're good to merge.  Thanks for the contribution by the way, we really appreciate it! 🌻

johnnyreilly:
@sokra (the creator of webpack) has just confirmed this should be safe so I plan to merge.  Thanks for the PR!

zinserjan:
@johnnyreilly awesome!

johnnyreilly:
Released with v2.3.1 - thanks again! https://github.com/TypeStrong/ts-loader/releases/tag/v2.3.1

Issue #597, opened 2017-08-09
The output of getCustomTransformers makes compiling fail when tsconfig.json contains rootDir.


zheeeng:
versions:
```
"typescript": "~2.4.0",
"ts-loader": "^2.2.1",
```

The test code:
```
// webpack.config.js
// path: module.rules

      {
        test: /\.(ts|tsx)$/,
        include: paths.appSrc,
        loader: require.resolve('ts-loader'),
        options: {
          transpileOnly: true,
          getCustomTransformers: () => ({
            before: [ content => node => node ] // Issue caused by here!!!!!!!!!!
          }),
        },
      }
```

complier print:
```
error TS6059: File 'index.tsx' is not under 'rootDir' '/Users/Zheeeng/Workspace/project/src'. 'rootDir' is expected to contain all source files.
```

The tsconfig.json:
```
"compilerOptions": {
    "rootDir": "src"
}
```

Does the getCustomTransformers saved the temp content to other place and fails the compiling?

I encountered this issue while I'm using ts-import-plugin, and reported it at https://github.com/Brooooooklyn/ts-import-plugin/issues/5. After a simple test, I find it's something wrong with ts-loader.

johnnyreilly:
Do you have any ideas @longlho or @igorbek?

Brooooooklyn:
This issue is actually cased by `transpileOnly`  option.
if this option was passed to `ts-loader`, it would use .

Because https://github.com/Microsoft/TypeScript/blob/master/src/services/transpile.ts#L92 and https://github.com/Microsoft/TypeScript/blob/master/src/compiler/program.ts#L1797
we shouldn't pass `rootDir` to `compiler.transpileModule`.

I will fix it later with a pull request.

zheeeng:
@Brooooooklyn Thx for your work.

PR #598, opened 2017-08-09
fix(tsconfig): stop pass rootDir option to TypeScript compiler

Brooooooklyn:
fix https://github.com/TypeStrong/ts-loader/issues/597


johnnyreilly:
Thanks! Just waiting for the tests to go green and I'll merge. I'm on holiday right now and so won't be able to get a release out immediately but will do so as soon as I can. Thanks again!

PR #605, opened 2017-08-16
Fix findConfigFile in Windows

### LINKED ISSUE #604: Should "configFileName" be used as filename only or with path?
I am using follow configuration:
```
{
  configFileName: path.join(__dirname, 'tsconfig.json'),
}
```

It works as expected in POSIX file systems. But in windows, 'ts-loader' cannot find the config file. Below function in config.js is broken in windows: https://github.com/TypeStrong/ts-loader/blob/master/src/config.ts#L76
```javascript
function findConfigFile(compiler, searchPath, configFileName) {
    while (true) {
        var fileName = path.join(searchPath, configFileName);
        if (compiler.sys.fileExists(fileName)) {
            return fileName;
        }
        var parentPath = path.dirname(searchPath);
        if (parentPath === searchPath) {
            break;
        }
        searchPath = parentPath;
    }
    return undefined;
}
```
 
I just want to discuss that should we use configFileName as simple filename or path. If we use it as filename, then it's fine. But if we can use it as path, we should fixed this bug in windows file system.

mengxy:
If configFileName is a full filepath, findConfigFile return nothing.

Modify configFileName to fix windows compatibility.

Closes #604

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -74,6 +74,9 @@ export function getConfigFile(
  * and continuing up the parent directory chain.
  */
 function findConfigFile(compiler: typeof typescript, searchPath: string, configFileName: string): string | undefined {
+    // Remove path.root from configFileName to fix windows specific issue. #604
+    configFileName = configFileName.substr(path.parse(configFileName).root.length)
+
     while (true) {
         const fileName = path.join(searchPath, configFileName);
         if (compiler.sys.fileExists(fileName)) {


johnnyreilly:
Thanks! Will look to merge this tomorrow I hope

johnnyreilly:
Shipped with 2.3.3

Issue #603, opened 2017-08-15
LoaderOptions: configFileContent


hdennen:
We're specifying different tsconfig settings for different builds and I'd like to be able to just generate one and pass it in without writing anything to disk. Is anyone working on anything like this? I'm happy to take a shot at it when I get the time.

johnnyreilly:
Not at present but we welcome contributions - go for it!

hdennen:
Will do 👌

Loilo:
We could extend the scope of the possible new `configFile` option to achieve this (see #607). Naming-wise, it could definitely go in there.

Currently, the option takes file names and path, so strings only. We could passing a JavaScript object interpret as "passing the config directly". What do you think, @hdennen?

hdennen:
@Loilo Honestly I don't think it matters. My knee jerk response is to put options 1 to 1 with what they do but adding more functionality to the configFile option seems fine.

PR #607, opened 2017-08-25
Add `configFile` option

Loilo:
This deprecates the `configFileName` option and renames it to `configFile`.
It keeps its user-facing behaviour, improving some path resolutions under the hood.
Using the `configFileName` option will now raise a `logWarning` in the console.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -24,7 +24,7 @@ export function getConfigFile(
     log: logger.Logger,
     compilerDetailsLogMessage: string
 ) {
-    const configFilePath = findConfigFile(compiler, path.dirname(loader.resourcePath), loaderOptions.configFileName);
+    const configFilePath = findConfigFile(compiler, path.dirname(loader.resourcePath), loaderOptions.configFile);
     let configFileError: WebpackError | undefined;
     let configFile: ConfigFile;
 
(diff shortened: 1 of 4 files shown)


johnnyreilly:
Thanks! I'll try and take a look this weekend. Tests pass!

johnnyreilly:
Brilliant - thanks!

johnnyreilly:
shipped with v2.3.4 - thanks for your help!

Loilo:
My pleasure. 🙃

develar:
So cool, electron-webpack uses ts-loader in the user projects and in any case have to find tsconfig (and set to fork-ts-checker-webpack-plugin if need, option `tsconfig`) So, a little bit more reliable (because don't rely on ts-loader logic anymore) and faster. Thanks a lot.

Loilo:
@develar Glad that it helped.

I'm still wondering if this option could need an additional per-entry-point configurability, something like this:

```javascript
configFile: {
  '*': 'tsconfig.json', // ← default value
  'named-entry-point-a': '/var/www/tsconfig.json',
  'named-entry-point-b': 'wildly-different-name.json'
}
```

Before possibly opening an issue about this, what do you @develar as an experienced `ts-loader` user think about this?

develar:
@Loilo I think it is not required. Currently, we use one tsconfig.json for both parts (Electron app consists of two parts — code for renderer process and code for main process). I cannot imagine when it can be required :)

Loilo:
Okay, thanks for your feedback – I guess then I could go deeper in thinking about [this](https://github.com/TypeStrong/ts-loader/issues/603#issuecomment-324963843) instead.

PR #621, opened 2017-09-05
Make TypeScript `basePath` configurable, fixes #619

### LINKED ISSUE #619: Config files don't work in arbitrary places
Related to #607.

If the `configFile` option is set to a path and points to a `tsconfig.json` outside the respective entry point's parent directory chain, the config file will be found and used, but the webpack process will throw errors. (Usually that's a `TS18003: No inputs were found in config file`).

I assume this is TypeScript not being able to handle the "unusual" location of the `tsconfig.json`.

I'll investigate this and see what we can do about this.

Loilo:
Enables the behaviour suggested in #619 to modify the `basePath` provided to the TS compiler.

Added:

- [x] Implementation
- [x] Comparison test
- [x] Hint in README


johnnyreilly:
Great! I'll look to include this in the next release. Thanks!

johnnyreilly:
Released with https://github.com/TypeStrong/ts-loader/releases/tag/v2.3.5

pettyalex:
This seems to break the default tsconfig.json as distributed with [create-react-app-typescript](https://github.com/wmonk/create-react-app-typescript). 

I'm seeing failures with 
```
TS6059: File '/app/src/App.test.tsx' is not under 'rootDir' '/app/src/src'. 'rootDir' is expected to contain all source files.
```

pettyalex:
The quick and dirty workaround is to change rootDir to '.' rather than 'src', but I'm not sure whether create react app's tsconfig is wrong, or some other part of the build pipeline is causing the issue.

Loilo:
That's weird. As far as I can tell, the provided `basePath` should be correct (it does not contain `src/src`, just a single `src`).

I'm probably going to have to look into how exactly TypeScript handles the `basePath`, this smells like some errornous assumptions from my side.

johnnyreilly:
If you could look into this I'd appreciate that @Loilo. @pettyalex let's see what @loilo comes up with. If all else fails we can revert this change and push out a new version. For now I would pin your version to 2.3.4 though.

Loilo:
I think that's currently the best solution. I created a quick repro repo where I couldn't nail it down yet, gonna get to work now but going to look at the problem from there. I suggest to move every further communication over to [wmonk/create-react-app-typescript#156](https://github.com/wmonk/create-react-app-typescript/issues/156).

PR #622, opened 2017-09-05
Add an additional check for js files before reusing isExternalLibaryImport (#620)

### LINKED ISSUE #620: Typescript files aliased to node_modules's module won't be compiled
Hi ts-loaders,

Got a new issue for you - happy to provide any additional information you guys need (and also to help out adding the fix if given a bit of guidance).

### Versions
- webpack@3.5.5
- typescript@2.4.2
- ts-loader@2.3.4

### Some background:
This is proprietary code so apologies for being slightly vague - I'll work on adding a little repro project if this isn't enough info.

We're using effectively an interface module - just typings - shipped from our local npm server. For production this gets bundled with a real implementation. For local testing however we use webpack to alias a simple local implementation of the interface that we can instrument as we see fit. This local implementation is (important in the context of this bug) written in typescript and needs to be compiled with the rest of the library for testing.

### Problem:
The local implementation file is not being compiled resulting in webpack failing to find a valid implementation. Example output:
```
WARNING in ./test/MockAPI.ts
Module build failed: Error: Typescript emitted no output for D:\dev\project\test\MockAPI.ts.
    at successLoader (D:\dev\project\node_modules\ts-loader\dist\index.js:47:15)
    at Object.loader (D:\dev\project\node_modules\ts-loader\dist\index.js:29:12)
 @ ./test \.tsx?$
```

### Investigation:
I'm pretty sure was introduced by pull #590 in response to issue #586. This logic, which appears to have been added to support typings bundled with js files, is also hit when you have a local .ts file that still needs to be compiled. In our case the new regex matches the typings interface in node_modules, meaning the local implementation is marked as being an "externalLibraryImport" and therefore is not compiled.

The code in question:
https://github.com/TypeStrong/ts-loader/blob/254718854259233c2ee3696fafbd6088f0b175aa/src/servicesHost.ts#L155-L158

### Possible fix
A simple fix would be just adding a further check to ensure it's already JS:

```typescript
if (resolutionResult.resolvedFileName === tsResolutionResult.resolvedFileName ||
    (/node_modules(\\|\/).*\.d\.ts$/.test(tsResolutionResult.resolvedFileName) && resolutionResult.resolvedFileName.endsWith('js')) {
    ...
}
```
This doesn't seem the cleanest to me but I do think it would fix both the original issue and our issue (tested locally for us and it works - will have to clone the repo and get it tested for the original issue).

Thanks!

WillMartin:
Fixes #620 by only reusing Typscript's isExternalLibraryImport designation if the loader-resolved module is a JS file and not a TS file.

Added a comparison-test rather than an execution-test as this really is just a question of "does it compile" rather than changing any behavior related to the generated JS.

Note that this also locks typescript down to 2.4.2 because right now building from a clean clone fails. Since typescript is currently ^2.4.0 it upgraded locally to the latest - 2.5.2 - and since 2.5.x updates some general typings (in this case System.readFile) that causes a build failure in config.ts. Not sure how you guys want to deal with that - it seemed outside the scope of my change to fix that but it did look fairly simple.

Please let me know what you're thinking!

-Will

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "mkdirp": "^0.5.1",
     "mocha": "^3.1.0",
     "rimraf": "^2.4.2",
-    "typescript": "2.4.2",
+    "typescript": "^2.4.2",
     "typings": "^2.0.0",
     "webpack": "^2.2.0"
   }
--- a/src/servicesHost.ts
+++ b/src/servicesHost.ts
@@ -112,6 +112,14 @@ function resolveModuleNames(
     return resolvedModules;
 }
 
+function isJsImplementationOfTypings(
+    resolvedModule: ResolvedModule,
+    tsResolution: ResolvedModule
+) {
+    return resolvedModule.resolvedFileName.endsWith('js') && 
+            /node_modules(\\|\/).*\.d\.ts$/.test(tsResolution.resolvedFileName);
+}
+
 function resolveModuleName(
     resolveSync: ResolveSync,
     moduleResolutionHost: ModuleResolutionHost,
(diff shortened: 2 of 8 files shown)


johnnyreilly:
Thanks - this looks great! Don't worry, I'll sort out the test stuff. Hopefully I'll get time to merge this and a couple of other things this week and then put out a release.

johnnyreilly:
Thanks!

WillMartin:
Thanks for being so responsive! Looking forward to the release

johnnyreilly:
Released with https://github.com/TypeStrong/ts-loader/releases/tag/v2.3.5

PR #633, opened 2017-09-11
Test display change

mblandfo:
This PR does the following:

1) Refactor mocha reporter options into common location test/reporterOptions.js
2) Change mocha report options from success: 'bgGreen', to success: 'green'
3) better path finding for mocha and karma (running tests didn't find them locally for me, I didn't have either installed globally)


johnnyreilly:
Heya - I'm afraid you've broken Windows with your change :sob: 

```
node_modules' is not recognized as an internal or external command,
operable program or batch file.
'node_modules' is not recognized as an internal or external command,
operable program or batch file.
'node_modules' is not recognized as an internal or external command,
operable program or batch file.
'node_modules' is not recognized as an internal or external command,
operable program or batch file.
'node_modules' is not recognized as an internal or external command,
operable program or batch file.
'node_modules' is not recognized as an internal or external command,
```

FWIW I don't believe the existing mechanism relies on mocha / karma being installed globally.  I don't and they run on my machine...

johnnyreilly:
BTW did you remember to `npm run build` before `npm test`?  I often forget.  In fact I'm planning to add a `pretest` script `npm run build`

mblandfo:
Oh, right - they probably run fine on your machine because you are using "npm run ..." whereas I wanted to run --single-test, and did not see a npm script for that, so I was running it directly. Perhaps a better solution would be to add a npm script for single test.

mblandfo:
Maybe npm run test should just call npm run build. The build is super fast anyway.

johnnyreilly:
> Maybe npm run test should just call npm run build. The build is super fast anyway.

I'll make that change

johnnyreilly:
BTW to run single tests:

To run a single comparison test use:

```
npm run comparison-tests -- --single-test nameOfTest
```

Documented here: https://github.com/TypeStrong/ts-loader/blob/master/test/comparison-tests/README.md

To run a single execution test:

```
npm run execution-tests -- --single-test nameOfTest
```

Documented here: https://github.com/TypeStrong/ts-loader/blob/master/test/execution-tests/README.md

Issue #263, opened 2016-08-10
Combined type definitions (.d.ts) output


mblandfo:
I'm writing a typescript library for use with other typescript projects. Running ts-loader I got a .d.ts file (in the wrong directory, but that's already a tracked issue).

This .d.ts file contained import statements to import classes from local library files (these would not be shipped) instead of the actual class declarations I needed.

I found a npm plugin that fixes this: https://www.npmjs.com/package/declaration-bundler-webpack-plugin. This plugin has a bug due to a webpack change, but I have pasted a fixed version here: http://pastebin.com/ypqed086

I think that should be the default declarations file behavior.

JasonKleban:
what is the bug that it fixes?  Is it:

```
            var lines = data.split("\n");
                            ^

TypeError: Cannot read property 'split' of undefined
    at DeclarationBundlerPlugin.generateCombinedDeclaration (...node_modules\decl
aration-bundler-webpack-plugin\plugin.js:44:29)
```

mblandfo:
Yes, that is the bug.

It appears the change I made was to change this line:

`var data = declarationFile._value;`

into this:

`var data = declarationFile._value || declarationFile.source();`

So presumably webpack changed from using _value to using .source(). I don't think declaration-bundler-plugin has a github or anything (it's just on npm?) I did email the author back in 2016 but he never responded. Can ts-loader be fixed to not need this plugin? It'd be really nice if ts-loader just did this and didn't have to rely on declaration-bundler-plugin at all!

Otherwise I guess someone can make a declaration-bundler-plugin2 package or something.

johnnyreilly:
@mblandfo - if you'd like to submit a PR I'll take a look

mblandfo:
@johnnyreilly it looks like this wouldn't be hard to do, like in https://github.com/TypeStrong/ts-loader/blob/master/src/instances.ts at the bottom plugins are getting added, so the fixed version of declaration-bundler-plugin could probably be added there.

I started to write a PR but get errors running tests. I did this:

```
git clone https://github.com/TypeStrong/ts-loader.git
cd ts-loader
npm install
npm run build
npm run test
```

and got test failures such as this:

```
simpleDependency
    1) should have the correct output
    2) should work with transpile


  0 passing (6s)
  2 failing

  1) simpleDependency should have the correct output:

      Uncaught AssertionError: patch0/output.txt is different between actual and expected
      + expected - actual
```

[...]

PR #634, opened 2017-09-11
Declaration bundler plugin

mblandfo:
Add option to bundle files into a single .d.ts file

#263


johnnyreilly:
I'll take a look at this once we've sorted out the other PR :smile:

mblandfo:
I'll need to add more documentation for the limitations here - typescript's compilerOptions.target must be "AMD" or "System". This is the same limitation typescript has on combining to a single outFile. Also, I think file export names must match file names.

johnnyreilly:
First of all - well done for your work.  

However, I've been thinking about this some more and I'm not sure it makes sense for this functionality to be part of ts-loader itself.  I think it makes more sense for this to be published to npm as a separate project and maintained separately.  

I've a couple of reasons which I hope you'll understand:

1. It's not core ts-loader functionality; ts-loader is basically about taking typescript, turning into javascript and feeding it to webpack.  The plugin here is not part of that functionality and so it doesn't feel to me like it "belongs" in ts-loader.

2. There is more than one typescript webpack loader (one day I hope we'll all join together but that hasn't happened yet).  The functionality in this plugin could / should be used by anyone using any loader ideally.  If it becomes part of ts-loader then that won't be possible.

I'd be perfectly happy to have documentation in ts-loader that links to this plugin (just as we have docs that link to the `fork-ts-checker-webpack-plugin` etc)  But I don't think this functionalty should land in ts-loader directly for the reasons given.

I hope you understand my reasoning.  What do you think?

mblandfo:
Ok, no worries, thanks for looking at this.

PR #642, opened 2017-09-24
Add vanilla-jsts example

nojvek:
Show how you can use ts-loader as a hybrid javascript and typescript loader with some js files having //@ts-check


johnnyreilly:
Thanks!

piec:
Hi, this is for checking js files with tsc as well right?
This seems to require `"checkJs": true` with ts-loader. The ` //@ts-check ` annotation at the top of js files does not seem to be taken into account

Issue #267, opened 2016-08-20
Tries to compile sources that are not supposed to be loaded


whitecolor:
Webpack 2 and TS2 beta
ts-loader tries to compile even source that should not be loaded by webpack. Is it expected behaviour?

blakeembrey:
Can you give more information? I can't tell you without seeing an example, but most like it's not compiling it - it's using it for type information and that would be expected.

whitecolor:
I serve files mostly in `/client` directory (by `ts-loader`), and I have ts file say `server-somthing.ts` in server (that is loaded using ts-node) this file is not shared with client part.

And say there is a mistake in `server-somthing.ts`.

But when webpack lanunches it error with compile error in `server-somthing.ts`. To get rid of it I have to limit files in `tsconfig` - if I use https://github.com/s-panferov/awesome-typescript-loader instead of `ts-loader` (just replace `ts-config` for it in webpack.confg) it works ok (doesn't try to load) to compile `server-somthing.ts`.

Vloz:
Same things for me: 
ts-loader is giving error from file not used in the entry, but are part of the file tree.
I have 5 different ts entry path in the same filetree, and it looks like they are parsing each-other's files... Wonder if it doesnt overslow the build speed... @_@

timocov:
@blakeembrey Hi, guys! This issue prevents us from using ts-loader. We've made a simple example so you can reproduce it.
This example contains tests folder with additional dependencies in package.json which are not to be installed to build the base project.
just unpack it, and run
```
npm install
./node_modules/.bin/webpack
```
You'll get an error in `tests/src/test.ts` which is not required to build the base project:
`(1,25): error TS2307: Cannot find module 'jquery'.`
[ts-loader-fail.zip](https://github.com/TypeStrong/ts-loader/files/523866/ts-loader-poc.zip)

johnnyreilly:
Hi @timocov,

Thanks for the repro - that's very helpful and we're always grateful when someone can provide a clear example of something not working as they expect.

We're open to a receiving a PR that will resolve this 🌷

EugeneZee:
@johnnyreilly Hi, this issue is solved in [awesome-typescript-loader](https://github.com/s-panferov/awesome-typescript-loader).
The problem is that `after-compile` in ts-loader looks for errors in all files from tsconfig.json.
Look at how it is done in `awesome-typescript-loader`:
[...]

PR #671, opened 2017-10-25
Compile bundled files and typing files

maier49:
This PR adds a flag that modifies the behavior of `ts-loader` so that it only loads files that are bundled by webpack and any `.d.ts` files that match the config in `tsconfig.json`. This is an attempt to implement the behavior described in #267, or at least a starting point. 

The approach is very simple and I can't help but feel like there's some obvious issue with it that I'm not thinking of, but it seems to work fine in the test case I created at least. The reason for compiling all `.d.ts` files as well as bundled files is that they may include types that are used in files that are bundled without being explicitly imported. Attempting to determine exactly which files are needed seems like it may be quite difficult and potentially slow.


johnnyreilly:
Awesome - thanks for this! I'll try and take a look later on. BTW we only need comparison test output for the latest released version of TypeScript (2.5 right now). I really should make that clearer in the docs!

src/index.ts
```diff
@@ -110,7 +110,7 @@ function getLoaderOptions(loader: Webpack) {
 }
 
 type ValidLoaderOptions = keyof LoaderOptions;
-const validLoaderOptions: ValidLoaderOptions[] = ['silent', 'logLevel', 'logInfoToStdOut', 'instance', 'compiler', 'configFile', 'transpileOnly', 'ignoreDiagnostics', 'errorFormatter', 'colors', 'compilerOptions', 'appendTsSuffixTo', 'appendTsxSuffixTo', 'entryFileCannotBeJs' /* DEPRECATED */, 'happyPackMode', 'getCustomTransformers'];
+const validLoaderOptions: ValidLoaderOptions[] = ['silent', 'logLevel', 'logInfoToStdOut', 'instance', 'compiler', 'configFile', 'transpileOnly', 'ignoreDiagnostics', 'errorFormatter', 'colors', 'compilerOptions', 'appendTsSuffixTo', 'appendTsxSuffixTo', 'entryFileCannotBeJs', 'onlyCompileBundledFiles' /* DEPRECATED */, 'happyPackMode', 'getCustomTransformers'];
```
johnnyreilly: Please can you move `/* DEPRECATED */` back to after `'entryFileCannotBeJs`?

johnnyreilly:
Generally this looks good - could you remove comparison test output apart from 2.5?  Also could you add something to the `README` about the new option?

maier49:
@johnnyreilly Thanks for the quick review. I made some updates to address your comments.

johnnyreilly:
Sweet - I'll try and take a look tomorrow. Thanks!

johnnyreilly:
Perfect!

johnnyreilly:
Hope to ship this with ts-loader 3.1.0

PR #674, opened 2017-10-26
Fix none-node_module imports of built ts files

### LINKED ISSUE #667: Error importing buildt ts files with allowJs
We have a subproject that is built separately so that it produces `.js` and `.d.ts` files. With `"allowJs": true` I get the following error when importing a module from it:
`TS2306: File /path/to/file.js is not a module.`

This can be fixed by setting `entryFileCannotBeJs: true` but as that is going away I think this issue must be fixed first.

I have not looked deeper into it but the feeling I get is that `ts-loader` prefer the `.js` file over the `.d.ts` and thus errors. Plain TypeScript have not errors with this setup.

Pajn:
fixes #667

### DIFF
--- a/src/servicesHost.ts
+++ b/src/servicesHost.ts
@@ -152,7 +152,7 @@ function isJsImplementationOfTypings(
     tsResolution: ResolvedModule
 ) {
     return resolvedModule.resolvedFileName.endsWith('js') &&
-        /node_modules(\\|\/).*\.d\.ts$/.test(tsResolution.resolvedFileName);
+        /\.d\.ts$/.test(tsResolution.resolvedFileName);
 }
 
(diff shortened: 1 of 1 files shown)


johnnyreilly:
This looks good!  It's a minor change and the existing tests still pass with it which is good. 

Generally it's good to have a test to cover changed behaviour to prevent regressions in the future. 
 Unfortunately it looks tricky to create a test for this will be hard given the difficulties creating a repro.

I think it makes sense to merge and when I release this I'll keep my eyes open for anyone having issues.  Hopefully there won't be any.

Thanks so much for drilling into this and coming up with a solution!

johnnyreilly:
Published with [3.1.1](https://github.com/TypeStrong/ts-loader/releases/tag/v3.1.1) - thanks so much!

PR #676, opened 2017-11-01
Fix default logLevel in README

gsteacy:
Looks like this was missed in #643.


johnnyreilly:
Whoops - yup, totally right. Thanks so much!

PR #678, opened 2017-11-04
upgraded to building with ts 2.6

johnnyreilly:


piotrwitek:
Hello @johnnyreilly :)
Just a quick question, is this PR added a support for a new Faster --watch mode announced here: (https://blogs.msdn.microsoft.com/typescript/2017/10/31/announcing-typescript-2-6/) ?
Or is it not possible to leverage currently with ts-loader and it will need this new API they mentioned there?

johnnyreilly:
AFAIK we will need the API that is due to ship with 2.7. That said, I haven't had chance to look at that yet. In the meantime if you want to speed up your builds then I'd recommend you give this a read:  

https://medium.com/webpack/typescript-webpack-super-pursuit-mode-83cc568dea79

piotrwitek:
@johnnyreilly Good stuff, thanks for a quick response :)
BTW: I read the article already :) 
Cheers!

PR #701, opened 2017-12-23
Report diagnostics only on certain files [WIP]

### LINKED ISSUE #700: Allow suppressing diagnostics for certain files
### Expected Behaviour

I'd like to be able to ignore typescript errors/diagnostics for certain files (or only reporting errors for certain files).

This feature is available in [at-loader](https://github.com/s-panferov/awesome-typescript-loader#reportfiles-string) as `reportFiles`. This can be useful when some type definitions contain errors that do not otherwise break the application. In this case it can be useful to squelch them while waiting for upstream to fix the issue.

### Actual Behaviour

ts-loader reports all errors and we can only filter out some specific diagnostics but this is global.

freeman:
Fixes #700 

This is a first implementation heavily borrowed from at-loader.

I had to add an explicit `micromatch` dependency (even though we have it indirectly, I felt it was cleaner to have a dependency on it if we rely on it). Is that ok ?

Also currently the pattern is applied against the fully qualified path to the file. I think we should match against the `context` relative path. Unfortunately it is not available in the `formatErrors` function.

I think it is reachable in every call site. Is adding a new argument to `formatErrors` OK ? (note that I already took a first step by making the last argument non optional as it was provided at every call site)


johnnyreilly:
Initially this looks good to me. I'm on my mobile and so it's hard to do a proper review. Couple of things:

> I had to add an explicit micromatch dependency (even though we have it indirectly, I felt it was cleaner to have a dependency on it if we rely on it). Is that ok ?

Yeah that's probably fine and if we're using something I definitely want to be explicit. How did we have it indirectly in the first place?

> I think it is reachable in every call site. Is adding a new argument to formatErrors OK ? (note that I already took a first step by making the last argument non optional as it was provided at every call site)

Yup absolutely fine.

src/utils.ts
```diff
@@ -49,12 +50,21 @@ export function formatErrors(
     loaderOptions: LoaderOptions,
     colors: Chalk,
     compiler: typeof typescript,
-    merge?: { file?: string; module?: WebpackModule }
+    merge: { file?: string; module?: WebpackModule }
 ): WebpackError[] {
 
     return diagnostics
         ? diagnostics
             .filter(diagnostic => loaderOptions.ignoreDiagnostics.indexOf(diagnostic.code) === -1)
```
johnnyreilly: Could we merge the 2 filter statements in `utils` please? I'd only like to iterate the collection a single time if possible.

freeman: sure will do

freeman:
> Yeah that's probably fine and if we're using something I definitely want to be explicit. How did we have it indirectly in the first place?

It is karma that is installing it (so only a dev dependency which makes it even more important to add it to the direct dependencies).

freeman:
Seems complete. But I am having second thoughts about the option name. 

`reportFiles` is nice because it is the same as at-loader (so moving between loader is easier). But I feel like `ignoreFiles` would be more coherent with `ignoreDiagnostics`. I am happy to change it if you prefer the `ignoreFiles` list approach.

johnnyreilly:
There's a long standing plan (which may or may not happen) to merge ts-loader and ATL. As part of that we'd discussed standardising on the same API as we moved towards merging. For that reason I think going with the same option name as ATL is probably a good idea.

johnnyreilly:
I've fixed (I think) the broken test with this: https://github.com/freeman/ts-loader/pull/1

README.md
```diff
@@ -245,6 +245,10 @@ messages are emitted via webpack which is not affected by this flag.
 You can squelch certain TypeScript errors by specifying an array of diagnostic
 codes to ignore.
 
+#### reportFiles *(string[]) (default=[])*
+
+Only report errors on files matching these glob patterns.
```
johnnyreilly: Could you add an example here please?  Examples rule the world when it comes to docs in my experience.

src/utils.ts
```diff
@@ -49,12 +50,28 @@ export function formatErrors(
...
-    merge?: { file?: string; module?: WebpackModule }
+    merge: { file?: string; module?: WebpackModule },
+    context: string
 ): WebpackError[] {
 
     return diagnostics
         ? diagnostics
-            .filter(diagnostic => loaderOptions.ignoreDiagnostics.indexOf(diagnostic.code) === -1)
+            .filter(diagnostic => {
+
```
johnnyreilly: Could we lose this blank line please? Reminds me - must add prettier to ts-loader

johnnyreilly:
Could we remove the comparison test you've implemented and replace it with an execution test please? https://github.com/TypeStrong/ts-loader/blob/master/test/execution-tests/README.md

Execution tests are run against all versions of TypeScript that are supported and tend to be less flaky than comparison tests (which we only run against the latest supported version of TypeScript)

freeman:
I will try the execution test approach

freeman:
I don't think the execution tests approach can work because reportFiles (and ignoreDiagnostics) do not work at the typescript compiler level. 

They only filter errors from webpack reporting but if noEmitOnError is true, nothing will be emitted anyway. Not sure if it is possible to go one level deeper ...

johnnyreilly:
Yup - I think you're right. We could only use an execution test if we changed the approach so that the files being supplied to the Typescript compiler were varied by `reportFiles`. I think that would likely make things more complicated than they need be (and might introduce other issues) so a comparison test is fine.

src/utils.ts
freeman: done

README.md
freeman: I have added an example

johnnyreilly:
This all looks good - thanks for your work @freeman!  I'm currently shepherding through a very complex PR from one of the TypeScript team: https://github.com/TypeStrong/ts-loader/pull/685

That's a biggie so I'm holding off merging other PRs until that's in.   Once that's in I'll look to merge and release.  Watch this space!

johnnyreilly:
Tests aren't passing after other PRs were merged; I'm going to merge this and fix up afterwards

johnnyreilly:
Will look to release this with 3.3

PR #715, opened 2018-01-23
Fixes to support builder api for compiling https://github.com/johnnyreilly/typescript-ts-loader-watch-api-illustration correctly

sheetalkamat:
With this the error reporting happens correctly.


johnnyreilly:
I've pushed this out with a new release of ts-loader (v3.3.1): https://github.com/TypeStrong/ts-loader/releases/tag/v3.3.1

I'm just heading off to work - I'll try and test the watch changes tonight.  If anyone else is tracking this then do feel free to give v3.3.1 a spin and report back - every little helps!

PR #738, opened 2018-03-07
Remove cross-env from react-hot-boilerplate

nil4:
With the [upgrade to Webpack 4](https://github.com/TypeStrong/ts-loader/commit/c122c9ffbf24eca99ef9297e9241d6176542010d), passing `NODE_ENV` is no longer necessary, and neither  is the `cross-env` dependency. 

Removing `cross-env` as suggested in https://github.com/TypeStrong/ts-loader/commit/c122c9ffbf24eca99ef9297e9241d6176542010d#r27961370.


johnnyreilly:
Awesome - thanks!

Issue #748, opened 2018-03-19
TypeError: Cannot set property 'tsLoaderFileVersion' of undefined


baywet:
### Expected Behaviour
Build to work when happypackmode is not set
### Actual Behaviour
Transpilation fails with following error message
```
ERROR in ./somefile.ts
Module build failed: TypeError: Cannot set property 'tsLoaderFileVersion' of undefined
    at successLoader (C:\sources\gitttl\gsk\MeetingWorkspace\node_modules\ts-loader\dist\index.js:47:54)
    at Object.loader (C:\sources\gitttl\gsk\MeetingWorkspace\node_modules\ts-loader\dist\index.js:21:12)
```
### Steps to Reproduce the Problem
Have a project with webpack 3, ts-loader@latest, and happypackmode not set.

hrgui:
Same here, I had to revert back to `v3.3.1` in order to get this resolved.

johnnyreilly:
Yep - ts-loader 4 supports webpack 4. If you're on webpack 3 then run with ts-loader 3.x. I'm going to leave this open for visibility for a while, but there is no issue with ts-loader.

baywet:
does this mean I should update this PR to also include a comprehensive error message? https://github.com/TypeStrong/ts-loader/pull/749

baywet:
well one way would be to read the version of the package.json of the webpack module (the one installed, not the package.json of the current project)
maybe there's something like that in the webpack api, but I don't know it well enough.
If you look at my PR, anything without the property I was adding defensive programming for it webpack <4 so maybe instead of version checking only a capability checking could be implemented?

johnnyreilly:
If there's a simple (non-hacky) way to reliably test the webpack version I'm open to adding a "your webpack version is not supported - use webpack 4 or and older version of ts-loader". Otherwise I think it's best left as is.

This was intentionally a breaking changes release and TBH people should take a look at release notes if they are hoping to make a major version bump. Not trying to be difficult. Just trying to be a good semver citizen.

baywet:
agree 200% on the semver, I didn't do my homework properly before upgrading (lots of deps to upgrade).
I do believe however that's it's always better to have meaningful error messages.
Do you consider the "reading webpack's package.json version" option to be hacky?

johnnyreilly:
Uh.... It should work reliably. But it slightly gives me an icky feeling. Hmmmm. Not totally sure to be honest.

Rahulkr204:
[...]

PR #749, opened 2018-03-19
fixed a undefined error when using webpack 3

baywet:
Fixes https://github.com/TypeStrong/ts-loader/issues/748

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -87,7 +87,7 @@ function successLoader(
   );
 
   // _module.meta is not available inside happypack
-  if (!options.happyPackMode) {
+  if (!options.happyPackMode && loader._module && loader._module.buildMeta) {
     // Make sure webpack is aware that even though the emitted JavaScript may be the same as
     // a previously cached version the TypeScript may be different and therefore should be
     // treated as new


baywet:
before I checked the release notes https://github.com/TypeStrong/ts-loader/releases/tag/v4.0.0 I didn't know that this version couldn't be used with webpack 3.
I currently have a project with ts-loader 4, webpack 3 and for some reason npm is not complaining about dependency unmet. It's in happy pack mode and looks like it's working fine.
before reverting ts-loader or falling forward on webpack 4, I'm curious about the "level of incompatibility". What exactly makes it incompatible?
Thanks!

johnnyreilly:
> I'm curious about the "level of incompatibility". What exactly makes it incompatible?

webpack 4 uses more ES6 features; particularly using `Map`s where it previously used objects.

webpack has massively changed for v4 API-wise. ts-loader is exposed to those API changes; hence breaking changes release.  It would have been theoretically possible to have a release that catered for old and new but would have involved more work than I felt was justified.

My advice: fall back on ts-loader 3.5 - it's very similar (nearly functionally identical) to 4.1.

PR #756, opened 2018-04-06
Pass `context' to error formatters

gustavderdrache:
This PR adds the `context` variable to the `ErrorInfo` interface. The primary use case is here is to allow custom error formatters to recover the full path to the file in which an error occurred, since I wanted to read the file from disk to use `'babel-code-frame'` to go along with the message.

I couldn't find tests that exercise `errorFormatter` functionality, but I'm happy to write some extras if you'd like.

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -14,6 +14,7 @@ export interface ErrorInfo {
   file: string;
   line: number;
   character: number;
+  context: string;
 }
 
 export interface AsyncCallback {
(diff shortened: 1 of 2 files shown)


johnnyreilly:
Thanks for this!  Let me have a ponder and report back.

johnnyreilly:
Yeah - I'm happy to take this.  I'll merge as is and fix the tests myself.  Thanks for contributing!

PR #759, opened 2018-04-09
Example of using ts-loader at scale with fast incremental builds

kenotron:
As a followup of an article I wrote (https://medium.com/@kenneth_chau/speeding-up-webpack-typescript-incremental-builds-by-7x-3912ba4c1d15), adding an example for ts-loader's own repo so others can start off with something that is going to scale to ~9000 modules with ts-loader and still keep its blazing fast rebuilds.


johnnyreilly:
Thanks so much!!

examples/fast-incremental-builds/webpack.production.config.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
+
+module.exports = {
+    devtool: 'inline-source-map',
+    entry: './src/index.ts',
+    output: {
+        pathinfo: false
```
sokra: That's `false` in production mode anyway. I guess you can omit it.

johnnyreilly: Thanks @sokra

Issue #761, opened 2018-04-15
Hot Module Replacement Doesn't Work


NTARelix:
### Expected Behaviour
HMR should behave consistently with plain-old-js HMR (no `ts-loader`)

### Actual Behaviour
With HMR enabled and `ts-loader` used, HMR fails to recognize some changes as HMR, resulting in a full-page reload. Other times the HMR seems to succeed, but the replaced modules' values are not updated.

Also doesn't work in `awesome-typescript-loader`: [https://github.com/s-panferov/awesome-typescript-loader/issues/576](https://github.com/s-panferov/awesome-typescript-loader/issues/576)

### Steps to Reproduce the Problem
See README of demo repo linked below. Here's a sample of the HMR code that should work, but doesn't:
```
// ./src/index.ts
///<reference types="webpack-env" />
import { stringToLog } from './exports-string';

console.log('Initial value:', stringToLog);

if (module.hot) {
  module.hot.accept('./exports-string', () => {
    console.log('New value:', stringToLog);
  });
}
```

```
// ./src/exports-string.ts
export const stringToLog = 'initial value';
```

### Location of a Minimal Repository that Demonstrates the Issue.
[https://github.com/NTARelix/hmr-fail](https://github.com/NTARelix/hmr-fail)

NTARelix:
Ah, that does it. Thanks!

NTARelix:
Although it's kind of unfortunate HMR doesn't work out of the box.

PR #762, opened 2018-04-15
Add minimal HMR example

NTARelix:
Documenting HMR findings through https://github.com/TypeStrong/ts-loader/issues/761

I found another inconsistency with HMR compared to vanilla webpack, and have documented it in the example's README:

> You must re-require the hot replaced module

I'd probably call it a bug because it's still inconsistent with the way HMR works in vanilla webpack, but HMR is still achievable using `ts-loader`. Anyone else have opinions on this behavior?

### DIFF
--- a/README.md
+++ b/README.md
@@ -542,6 +542,17 @@ Because TS will generate .js and .d.ts files, you should ignore these files, oth
 
 It's worth noting that use of the `LoaderOptionsPlugin` is [only supposed to be a stopgap measure](https://webpack.js.org/plugins/loader-options-plugin/). You may want to look at removing it entirely.
 
+### Hot Module replacement
+
+To enable `webpack-dev-server` HMR, you need to follow the official [Webpack HMR guide](https://webpack.js.org/guides/hot-module-replacement/), then tweak a few config options for `ts-loader`. The required configuration is as follows:
+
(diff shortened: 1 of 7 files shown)


NTARelix:
I'm glad to contribute to great projects like this, and thank you for building and maintaining `ts-loader`!

The React example covers the same exact thing. However, this example is more bare-boned and calls out some `ts-loader` HMR caveats. I just want to make sure users know exactly how to set up HMR, with or without React, and I feel like the React example has a lot of boilerplate that might make it difficult to reason about why HMR doesn't work out of the box.

Maybe instead of an example it would make sense to provide some sort of callout to HMR config in the README? If you think this example is worth keeping then I'll update it to use `fork-ts-checker-webpack-plugin`. Otherwise we could reopen that HMR bug I made and try to fix it without needing this extra config.

Thoughts?

johnnyreilly:
> Maybe instead of an example it would make sense to provide some sort of callout to HMR config in the README?

If we haven't explicitly called out hmr in the readme that's a bit of a gap. Would you be up for adding a small bit of text about hmr to the readme that links to your example and to the existing one. And add the `fork-ts-checker-webpack-plugin` to your example?

I think that it might be good to have both.

johnnyreilly:
Thanks!

PR #766, opened 2018-04-23
fix(index.js): create only one instance per-build

eddyw:
The idea behind the PR is to create only one TypeScript instance.

The loader is imported/required by webpack only once, so we can initialize and keep only one instance of TypeScript instead of creating an instance for every file that needs to be transpiled.

This will have an effect on improving build time after the first file/source that passes through the loader that initializes the instance of TS.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,11 +13,14 @@ import {
   TSFile,
   TSInstance,
   Webpack,
+  WebpackError,
   LogLevel
 } from './interfaces';
 
 const webpackInstances: Compiler[] = [];
 const loaderOptionsCache: LoaderOptionsCache = {};
+
+let instanceOrError: { instance?: TSInstance; error?: WebpackError };
 
 /**
  * The entry point for ts-loader
@@ -26,7 +29,8 @@ function loader(this: Webpack, contents: string) {
   this.cacheable && this.cacheable();
   const callback = this.async();
   const options = getLoaderOptions(this);
-  const instanceOrError = getTypeScriptInstance(options, this);
+
+  instanceOrError = instanceOrError || getTypeScriptInstance(options, this);
 
   if (instanceOrError.error !== undefined) {
     callback(instanceOrError.error);


johnnyreilly:
Thanks for contributing!

Could you elaborate on how this improves upon the existing instance caching inside the body of `getTypeScriptInstance`? As is we should be reusing instances rather than recreating. If your change is offering something not already covered could you clarify please?

eddyw:
Yes :)

`getTypeScriptInstance` calls `getCompiler` which in line 15 imports/requires typescript:
```javascript
export function getCompiler(loaderOptions: LoaderOptions, log: logger.Logger) {
  let compiler: typeof typescript | undefined;
  let errorMessage: string | undefined;
  let compilerDetailsLogMessage: string | undefined;
  let compilerCompatible = false;

  try {
    compiler = require(loaderOptions.compiler); // << here
```
What it's currently doing is creating an instance for every file that is build. However, in simplistic words how TS works in my head is:
- Every time we call `transpileModule`, it generates an AST for the file and loads the needed types, etc
- If you call `transpileModule` a second time for a different file using the same instance, it will re-use previously generated AST and types (if both files use a library in common for instance)

What `ts-loader` is doing with `getTypeScriptInstance` is create a new instance for every content that passes through the loader instead of using one single instance. This is why it's slow with type checking since the required types cannot be re-used and are loaded every time for every file.

In fact, I had made a previous attempt to address this issue on my own with:
https://github.com/eddyw/owlpkg-typescript-loader
But it's kind of ugly, however it works faster than `ts-loader` and has similar results as using `HappyPack` (vanilla, without HappePack). Instead of maintaining my own package, I want to bring those ideas to `ts-loader`.

This is very noticeable in watch mode, since a single instance can detect if new types need to be loaded or not for a file that changed (because they are internally cached)

eddyw:
To add, if it explains better, in `instances.ts` if `loaderOptions.instance` is provided, then the instance is re-used. Otherwise, it creates an instance for every file which this PR tries to solve by having one global instance instead of creating one per file :)

johnnyreilly:
Thanks for the detail!  I really appreciate that.

However, I think that you may have missed something.  Namely that `instance` always has a value.  Or at least I think it does.  You can track it through [here](https://github.com/TypeStrong/ts-loader/blob/ec679eda55ec40b38f025a2a5814f69a0280c3b4/src/index.ts#L115):

```
function getLoaderOptions(loader: Webpack) {
  // differentiate the TypeScript instance based on the webpack instance
  let webpackIndex = webpackInstances.indexOf(loader._compiler);
  if (webpackIndex === -1) {
    webpackIndex = webpackInstances.push(loader._compiler) - 1;
  }

  const loaderOptions =
    loaderUtils.getOptions<LoaderOptions>(loader) || ({} as LoaderOptions);

  // NAME IS CREATED HERE
  const instanceName =
    webpackIndex + '_' + (loaderOptions.instance || 'default');

  if (loaderOptionsCache.hasOwnProperty(instanceName)) {
    return loaderOptionsCache[instanceName];
  }

  validateLoaderOptions(loaderOptions);

  // NAME IS PASSED AND INITIALISED HERE
  const options = makeLoaderOptions(instanceName, loaderOptions);

  loaderOptionsCache[instanceName] = options;

  return options;
}
```

So when [this code](https://github.com/TypeStrong/ts-loader/blob/ec679eda55ec40b38f025a2a5814f69a0280c3b4/src/instances.ts#L54) executes:

```
export function getTypeScriptInstance(
  loaderOptions: LoaderOptions,
  loader: Webpack
): { instance?: TSInstance; error?: WebpackError } {
  if (instances.hasOwnProperty(loaderOptions.instance)) {
    const instance = instances[loaderOptions.instance];
    ensureProgram(instance);
    return { instance: instances[loaderOptions.instance] };
  }

```

It should already be reusing the cached instance.  Do you want to check whether I'm right or wrong here?  I could well be wrong!

eddyw:
You're right. I missed that 😅

PR #365, opened 2016-11-09
node modules subdir

basarat:
refs https://github.com/TypeStrong/ts-loader/issues/278

Run as `npm run comparison-tests -- --single-test node_modules-subdir` 

Or cd into test dir, edit `webpack.config.js` in the test dir to have an additional `..` and run as `webpack` in the test dir (with `npm install webpack -g`)

PR #773, opened 2018-05-02
Add allowTsInNodeModules option for importing .ts files from node_modules.

aelawson:
Summary:

This PR adds an `allowTsInNodeModules` option that will allow importing of `.ts` modules from a `node_modules` subdirectory / dependency. At the time of writing, this will function as a "whitelist" - the user must manually add the desired files(s) or module(s) to the `files` or `include` block of their `tsconfig.json`, respectively. Otherwise, they will receive an error that nothing has been emitted as before with additional text suggesting this option if they really need it.

Changes:

- Added `allowTsInNodeModules` option.
- Updated error message when a module is imported from `node_modules` without this option enabled.
- Updated the `nodeModulesMeaningfulErrorWhenImportingTs` test outputs and have added a new `allowTsInNodeModules` test (with the option enabled) for TS 2.8.
- Updated `README` to include this option.

As most recently discussed here:
https://github.com/TypeStrong/ts-loader/issues/278

And also here:
https://github.com/TypeStrong/ts-loader/pull/365
https://github.com/Microsoft/TypeScript/issues/12358


johnnyreilly:
Thanks for the thorough PR @aelawson! I'm getting failing tests on the build servers. It occurs to me that an execution test would likely work just as well  here. Given that the comparison tests can be a world of pain, do you fancy trying putting together an execution test instead? I think that you can probably clone this one:

https://github.com/TypeStrong/ts-loader/tree/master/test/execution-tests/nodeResolution

And quickly tweak it into an execution test. Want to give it a whirl?

aelawson:
@johnnyreilly Sure I'll change the test to be an execution test :) Yeah, looks like one of the failures was me forgetting to update the test output after a small change, d'oh.

aelawson:
@johnnyreilly I have updated the test to be an execution test - looks like builds are passing now. Let me know if you've got any other concerns!

johnnyreilly:
Nice - thanks for the hard work! I'll set aside some time to review it very soon!

README.md
```diff
@@ -479,6 +479,34 @@ loading only those files that are actually bundled by webpack, as well as any `.
...
 compilation without being explicitly imported, and therefore not picked up by webpack.
 
+#### allowTsInNodeModules _(boolean) (default=false)_
+
+By default, ts-loader will not compile `.ts` files in `node_modules`.
+You should not need to recompile `.ts` files there, but if you really want to, use this option.
+Note that this option acts as a *whitelist* - any modules you desire to import must be included in
+the `"files"` or `"include"` block of your project's `tsconfig.json`.
+
+See: https://github.com/Microsoft/TypeScript/issues/12358
```
johnnyreilly: Can we linkify this please? i.e. `[https://github.com/Microsoft/TypeScript/issues/12358](https://github.com/Microsoft/TypeScript/issues/12358)` I think

src/index.ts
```diff
@@ -69,10 +69,17 @@ function successLoader(
     : getEmit(rawFilePath, filePath, instance, loader);
 
   if (outputText === null || outputText === undefined) {
-    const additionalGuidance =
-      filePath.indexOf('node_modules') !== -1
-        ? '\nYou should not need to recompile .ts files in node_modules.\nPlease contact the package author to advise them to use --declaration --outDir.\nMore https://github.com/Microsoft/TypeScript/issues/12358'
-        : '';
+    let additionalGuidance: string;
```
johnnyreilly: Can we turn this into a ternary please?  Something like:

```
const additionalGuidance = (!options.allowTsInNodeModules && filePath.indexOf('node_modules') !== -1)
    ? " By default, ts-loader will not compile .ts files in node_modules.\n" +
      "You should not need to recompile .ts files there, but if you really want to, use the allowTsInNodeModules option.\n" +
      "See: https://github.com/Microsoft/TypeScript/issues/12358"
    : "";
```

johnnyreilly:
Also, can we lose the comparison test now we have an execution test please?

aelawson:
@johnnyreilly Done. Just to double check on one thing: the remaining comparison test is the existing `nodeModulesMeaningfulErrorWhenImportingTs` test that checks for a meaningful error if a user hasn't enabled this option - you're okay with getting rid of that?

johnnyreilly:
Hey @aelawson,

I want to keep `nodeModulesMeaningfulErrorWhenImportingTs ` please; that's still useful.  Just wanted to get rid of the new comparison test since you've been able to come up with an execution test instead.  I prefer execution tests but they're not always appropriate.

aelawson:
@johnnyreilly That comparison test has been deleted :P The only changes to comparison tests left in the PR are for the `nodeModulesMeaningfulErrorWhenImportingTs` and `validateLoaderOptions` tests :)

Issue #775, opened 2018-05-09
add peerDependecies


yibuyisheng:
### Expected Behaviour

Should display some warn messages when installed with a mismatched webpack.

### Actual Behaviour

Display nothing.

### Steps to Reproduce the Problem

```
mkdir my-project
cd my-project
npm init
npm i webpack@3 -D
npm i ts-loader@4 -D
```

yibuyisheng:
> the problem is there's no nice way to determine webpack version.

I guess you have thought about the `peerDependencies` of `package.json`. Since the `webpack@4` was put into the `package.json`'s 'devDependencies' config, it means that the test codes runs well with `webpack@4`. So I want to know what the specific problem is while using `peerDependencies`.

johnnyreilly:
Can a package exist in `devDependencies` and `peerDependencies` at the same time?

yibuyisheng:
Yes. This is how [babel-preset-env](https://github.com/babel/babel/blob/master/packages/babel-preset-env/package.json#L54) does.

The function and behavior of `devDependencies` can be found [here](https://docs.npmjs.com/files/package.json#peerdependencies).

johnnyreilly:
Makes sense. Would you like to submit a PR?

PR #776, opened 2018-05-11
add `"webpack": "^4.5.0"` to `peerDependencies`

yibuyisheng:
fix https://github.com/TypeStrong/ts-loader/issues/775

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,9 @@
     "micromatch": "^3.1.4",
     "semver": "^5.0.1"
   },
+  "peerDependencies": {
+    "webpack": "^4.5.0"
+  },
   "devDependencies": {
     "@types/micromatch": "^3.1.0",
     "@types/node": "^9.6.2",
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "semver": "^5.0.1"
   },
   "peerDependencies": {
-    "webpack": "^4.5.0"
+    "webpack": "^4.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


johnnyreilly:
Great! Can we change this to `4.0.0`. As far as I know it works with 4 as a minimum?

yibuyisheng:
The webpack version in `peerDependencies` must be the same as the one in `devDependencies`, because the tests are run with the webpack version in `devDependencies`.

johnnyreilly:
Hmmm. I don't think that's going to work then. We use a 4.x version of webpack to run the tests. Unless we switch the devDependency to be 4.0.0 as well. We could do that I guess, just rely on the yarn.lock file to enforce this. 

Actually that seems reasonable. Do you want to make the devDependency change to 4.0.0?

johnnyreilly:
Hmmm.. some `yarn.lock` trickiness I think.

yibuyisheng:
The test codes are breakable because it seems that they are only run well with `webpack@4.5.0`. When I changed the webpack to `4.0.0` or `4.8.3`, tests failed.

I think we should find a way to make the test codes passed with `webpack@^4.0.0`.

yibuyisheng:
Maybe we should learn something from [babel-loader](https://github.com/babel/babel-loader/blob/master/test/loader.test.js#L56);

johnnyreilly:
Leave it with me. When I get a moment I'll take a look

johnnyreilly:
Closing as would now conflict with yarn PNP support

PR #792, opened 2018-06-12
fix(loader): new Error to webpack when errors occured in the loader f…

linxiaowu66:
…unction

When using `callback(instanceOrError.error)` to throw error, webpack will print :

```
NonErrorEmittedError: (Emitted value instead of an instance of Error) [object Object]
    at runLoaders (/Users/linxiaowu/SourceCode/eno/node_modules/webpack/lib/NormalModule.js:300:13)
    at /Users/linxiaowu/SourceCode/eno/node_modules/loader-runner/lib/LoaderRunner.js:364:11
    at /Users/linxiaowu/SourceCode/eno/node_modules/loader-runner/lib/LoaderRunner.js:230:18
    at context.callback (/Users/linxiaowu/SourceCode/eno/node_modules/loader-runner/lib/LoaderRunner.js:111:13)
    at Object.loader (/Users/linxiaowu/SourceCode/eno/node_modules/ts-loader/dist/index.js:19:9)
```

This way can not show the error message explicitly. So I change the returning value when error.

`new Error(instanceOrError.error.message)`

PR #801, opened 2018-06-26
typecheck the error on index.ts and do the necessary convertions

systemmetaphor:
My proposal to fix #792

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -13,11 +13,20 @@ import {
   TSFile,
   TSInstance,
   Webpack,
-  LogLevel
+  LogLevel,
+  WebpackError
 } from './interfaces';
 
 const webpackInstances: Compiler[] = [];
 const loaderOptionsCache: LoaderOptionsCache = {};
+
+function isError(value: any): value is Error {
+  return 'message' in value && 'name' in value;
+}
+
+function isWebPackError(value: any): value is WebpackError {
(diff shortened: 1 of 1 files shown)


systemmetaphor:
It will print something meaningful even if ts-loader catch’s a strange object as an error. I think that we can’t really trust on the error shape because some plugin might be messing things around. It will also keep the existing behavior for the expected error objects.

johnnyreilly:
I'm not sure if I want to go to this level of error interrogation or not.  I need to think about it a little I think.

@linxiaowu66 do you have a view on this? Your approach is simpler which is initially appealing to me.  Would this PR deal with your concerns?  Does it have any advantages / disadvantages that you're aware of?

systemmetaphor:
@johnnyreilly sure :), both options do the job, I was just guessing if it broke once due to receiving some unexpected shape it might break again on something similar

johnnyreilly:
Thanks @systemmetaphor - I'll have a think.  I'm going to go with one of these approaches.  Just need to decide which 😄

johnnyreilly:
Thanks for your efforts @systemmetaphor.  I've decided to go with the initial PR for simplicities sake.  I hope that's okay with you - and even though your PR hasn't landed I hope you understand why and aren't too discouraged! Much ❤️

systemmetaphor:
:) no problem

PR #817, opened 2018-08-10
Support project references

### LINKED ISSUE #815: Support for TS 3 project references
We have a monorepo when one of the packages is compiled with webpack, but references other package from the monorepo. We are using project references (new in TS 3) to compile only the packages, that are needed (the changed one and other that have reference to it). Is there a way to make it work with webpack? Currently we are compiling it with tsc and then bundling compiled js with webpack, but it is not an ideal solution, since many features doesn't work so well (e.g. source maps and custom loaders such as sass-loader) and it is quite slow.

andrewbranch:
Closes #815 

**TODO:**
- [x] support project references for `transpileOnly`
- [x] support project references for `experimentalWatchApi`
- [x] support project references for running with neither of the above
- [x] resolve loader output to built JS files for TS files that were part of a project reference
- [x] tests (~~for each of the three modes above,~~ one test for a successful build and one that fails because the dependent project isn’t built)
- [x] get Webpack to watch the built JS files instead of TS files in composite projects
- [x] manually test on a real-life big project
- [x] write execution tests


johnnyreilly:
This is PR "1" of the potential 2 I guess? You've just reminded me that I need to update the comparison test pack for TS 3.0. I'll try and do that this weekend so you don't have to.

I'll message here when it's done.

johnnyreilly:
I've upgraded the comparison test pack to 3.0 - feel free to merge that on to your pr

src/index.ts
```diff
@@ -63,11 +68,81 @@ function successLoader(
...
+        `Could not find output JavaScript file for input ${filePath} ` +
+          `(looked at ${jsFileName}).\nThe input file is part of a project ` +
+          `reference located at ${referencedProject.sourceFile.fileName}, ` +
+          'so ts-loader is looking for the project’s pre-built output on ' +
+          'disk. Try running `tsc --build` to build project references.'
+      );
+    }
 
+    if (!instance.compiler.sys.fileExists(mapFileName)) {
+      loader.emitWarning(
```
andrewbranch: It might be obnoxious to do this for every file. Maybe I can memoize by `referencedProject.sourceFile.fileName`? Or would it make more sense to validate each project reference while first setting up the program/languageService/watch?

src/utils.ts
```diff
@@ -227,3 +228,51 @@ export function arrify<T>(val: T | T[]) {
 
   return Array.isArray(val) ? val : [val];
 }
+
+export function ensureProgram(instance: TSInstance) {
```
andrewbranch: I'm not confident I understood the purpose of this function correctly. I moved it to utils because it seemed it was a way to get the `typescript.Program` out of the watch or the language service, but then noticed it potentially has side effects—I'm not sure what `watchHost.updateRootFileNames()` does and if it's harmful to call it more often than it was previously.

src/index.ts
johnnyreilly: That seems reasonable; presumably that's not something that changes often at all so should be a completely reasonable choice

src/utils.ts
johnnyreilly: Since it's only called conditionally on `instance.changedFilesList` then hopefully behaviour would be unchanged?

andrewbranch:
@johnnyreilly can the comparison test’s "patch" feature be used to patch the test's `webpack.config.js` itself?

I'm thinking about the scenarios that need to be tested, and they include all three modes of creating a TS program—`transpileOnly`, `experimentalWatchApi`, and the normal language service route. Would you create a different comparison test for each even though it shouldn't theoretically change the output at all? Or is that something that's more appropriate for an execution test?

test/comparison-tests/projectReferencesNoSourceMap/lib/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "composite": true
+  },
+  "files": [
+    "./index.ts"
```
andrewbranch: For some reason, without `files` or `include`, `program.getProjectReferences()[0].commandLine.fileNames` was empty. I thought it should have resolved to all `*.ts` files by default?

johnnyreilly: > To generate outputs for older TS versions, do I need to manually install each older version and run again?

No - we only run comparison tests for the latest TypeScript version. So 3 alone is fine

andrewbranch: When I was testing ad-hoc, the `reference.commandLine.fileNames` was working correctly, but I think I was using typescript@next... I'll have to investigate more.

andrewbranch: Ah, I was mistaken:

> All implementation files must be matched by an include pattern or listed in the files array. If this constraint is violated, tsc will inform you which files weren’t specified

https://www.typescriptlang.org/docs/handbook/project-references.html#composite

It seems it's working as intended, but via our programmatic route we don't get that warning. Maybe need to add our own.

johnnyreilly: >  can the comparison test’s "patch" feature be used to patch the test's webpack.config.js itself?

No I'm afraid not; the tests rely upon an unchanged webpack config between patches. Sorry

andrewbranch: > It seems it's working as intended, but via our programmatic route we don't get that warning. Maybe need to add our own.

I'm not sure if this can be done reliably and efficiently... without actually doing a build of the dependent projects, it's hard to tell if a given file was intended to be included by one of them, or was just intended to be processed by the parent project (i.e. in our case, run through the loader). Implementing a warning from ts-loader side while _consuming_ project references might be out of scope—hopefully it's something users will notice while actually running `tsc` on those projects, since pre-building the referenced projects is currently required, and `tsc` will emit the error if `files` or `include` don't cover all the files.

andrewbranch:
@johnnyreilly status update:

### Implementation
- Waiting on https://github.com/Microsoft/TypeScript/pull/26421 (no movement there since opening 😕)
- Anything else? Feel free to do a closer review now; I think it's pretty much done!

### Tests
- Is it a problem that the `.transpiled` one didn't show up for the `projectReferences` and `projectReferencesNoSourceMap` comparison tests?
- Should I copy any of the three comparison tests I added and run with `experimentalWatchApi`?
- Any other comments/feedback on tests?

johnnyreilly:
Sorry for the delay in responding @andrewbranch - I'm away and just have my phone and poor connectivity. In fact I'm typing this surrounded by 🐑 whilst lost rambling!

> Is it a problem that the .transpiled one didn't show up for the projectReferences and projectReferencesNoSourceMap comparison tests?

I'm not sure it is; my guess is that people will only use projectReferences with non-transpile mode. Either way this is fine for now in my view.

> Should I copy any of the three comparison tests I added and run with experimentalWatchApi?

Thanks but no. What I actually mean to do at some point is tweak the comparison test pack so it runs in 3 modes: standard, transpileOnly and watchApi. So in the longer term separate tests for the watchApi shouldn't exist. It's just another part of the test matrix if you will.

> Any other comments/feedback on tests?

They're not passing 😂

Jokes aside, I wouldn't worry about the tests for now. First get to an implementation that you're content with, let's review that together and then we can work together on getting some tests in place that cover this new functionality. 

Thanks again for all the work you're doing; I really appreciate it. Much ♥️🌻

I should be back online in a week or so. Assuming I find the way out of this field 😉

andrewbranch:
Thanks! Hope you are enjoying your vacation.

The build is failing because I wrote a line that assumes Microsoft/TypeScript#26421 will be merged. I think maybe we just need to press on without that for now. I'll give it a few more days but have something mergeable by the time you're back.

src/utils.ts
```diff
@@ -227,3 +229,78 @@ export function arrify<T>(val: T | T[]) {
...
+  filePath: string,
+  instance: TSInstance
+) {
+  if (isUsingProjectReferences(instance)) {
+    const program = ensureProgram(instance);
+    return (
+      program &&
+      program
+        .getProjectReferences()!
+        .find(ref => ref!.commandLine.fileNames.indexOf(filePath) > -1)
```
ajafff: do you handle missing referenced projects somewhere else or how can you be sure `ref` is always defined?
`ref` will be `undefined` if a referenced project cannot be resolved, i.e. missing folder or missing tsconfig.json.

src/index.ts
```diff
@@ -63,11 +69,80 @@ function successLoader(
...
+          referencedProject.sourceFile.fileName
+        } is using the ` +
+          `outFile' option, which is not supported with ts-loader.`
+      );
+    }
 
-  const { outputText, sourceMapText } = options.transpileOnly
-    ? getTranspilationEmit(filePath, contents, instance, loader)
-    : getEmit(rawFilePath, filePath, instance, loader);
+    const jsFileName = instance.compiler.getOutputJavaScriptFileName(
```
ajafff: given that the whole tsbuild API is now internal, this uses an internal function? I couldn't find a custom implementation if it in this PR

src/utils.ts
andrewbranch: It will be a holey array? I didn't realize that; it feels a bit unintuitive, but I guess the types agree. Let me do something here. Thanks for the feedback!

andrewbranch:
I feel like this is done. I'm going to try it out to split up a massive real-life project we have at work and see what happens! Will report back soon.

johnnyreilly:
Thanks chap! Looking forward to hear how it goes 😀

andrewbranch:
@johnnyreilly I found a bit of an issue that I'm not sure how to solve. Things seem to work with the compilation, but Webpack is watching the “wrong” files:

|Event|Expected Behavior|Actual Behavior|
|-|------------------|----------------|
|***.ts in a composite project changes**|Nothing|Webpack recompiles|
|***.js output files are rebuilt**|Webpack recompiles|Nothing|

johnnyreilly:
Questions:

- is this an issue in non-experimentalWatchApi mode also?
- is webpack actually watching the *.ts in a composite project changes? Or is it as a result of the file dependency graph in ts-loader. If it's the latter we can work around this; essentially make ts-loader depend upon *.js output files instead of *.TS files in the case of composite projects. If webpack directly is watching files (i.e. the former) then we can probably test and continue or not when ts-loader is triggered (I think). Or have a smarter regex in our webpack.config.js

Thoughts?

andrewbranch:
> is this an issue in non-experimentalWatchApi mode also?
Yeah, actually I only used the normal watch in this case.
> is webpack actually watching the *.ts in a composite project changes? Or is it as a result of the file dependency graph in ts-loader

Ok, did some debugging and if I hijack the dependencyGraph to swap the TS file for the JS file (and consequently what gets passed to `loader.addDependency`), the watch works as expected.

I'm not sure the best / most correct place to make that swap, though. What exactly is the language service's [`resolveModuleNames` hook](https://github.com/TypeStrong/ts-loader/blob/d8096aac0061cdd0ef9228fee5a0c2d137eca34f/src/servicesHost.ts#L124) for? I'm guessing that what it's returning to the language service is correct, but maybe where it calls `populateDependencyGraph` we should be putting the JS files in the graph instead of the TS files... but I don't totally understand everything that's going on there.

johnnyreilly:
> Ok, did some debugging and if I hijack the dependencyGraph to swap the TS file for the JS file (and consequently what gets passed to loader.addDependency), the watch works as expected

Awesome

> I'm not sure the best / most correct place to make that swap, though. What exactly is the language service's resolveModuleNames hook for? I'm guessing that what it's returning to the language service is correct, but maybe where it calls populateDependencyGraph we should be putting the JS files in the graph instead of the TS files... but I don't totally understand everything that's going on there.

I think it's a judgement call really - and remember; it's always something we can remedy later if the initial choice isn't sound.

I don't feel strongly personally. Putting the `js` files in the dependency graph could be fine I guess; I mean they're always going to be js files (the result of compilation) so why not? Might be a minimal perf improvement.

Do we lose anything that way though? Perhaps the ts metadata is useful? Go with your gut I say 😁

andrewbranch:
I think I got it. We can't very well put the JS files in the dependency graph because we won't yet have a TS program the first time the dependency graph gets built, and we need the TS program to know whether/where to look for a JS file. So, the dependency graph always points to the plain TS files still, but when doing `loader.addDependency`, I'm checking for project references there. I also added some project reference info to the `instance.files` cache since

- there are now two separate places that need the info,
- figuring out the project reference for a given file is linearly expensive with the size of the project, and
- I'm 99% sure project reference resolution can't change for a file during the life a program.

Take a look at the last few commits and make sure they make sense to you? Thanks for the continued input! 👍

andrewbranch:
I'm not sure what’s up with the test failures around `sourceMappingURL`—everything passes for me locally now.

Also tested the latest changes against the real life work project, and everything seems to work!

src/index.ts
```diff
@@ -90,7 +173,7 @@ function successLoader(
   );
 
   // _module.meta is not available inside happypack
-  if (!options.happyPackMode) {
+  if (!options.happyPackMode && loader._module.buildMeta) {
```
andrewbranch: Oh, I didn't actually mean to commit this change, but it was necessary for testing with our work project. Possibly because we're on Webpack 3.6? Once I added this, everything worked fine.

johnnyreilly:
> I'm not sure what’s up with the test failures around sourceMappingURL—everything passes for me locally now.

I'll see if I can look at the test when I'm at a machine; this may just be part of the comparison test pack quirkiness.

> Oh, I didn't actually mean to commit this change, but it was necessary for testing with our work project. Possibly because we're on Webpack 3.6? Once I added this, everything worked fine.

Question: how are you managing to test given that ts-loader is webpack 4 minimum? (We dropped webpack 3 support when webpack 4 shipped because it was a more breaking release than normal.)

pragatisureka:
Hi,

Great work on this feature. We are also trying to use project reference to manage our large and fragmented typescript codebase.  Our typescrit project configuration is similar to [project-reference-example by Ryan Cavanaugh](https://github.com/RyanCavanaugh/project-references-outfile/blob/master/tsconfig.json) but without the out file.

In the last few commens it is decided to disable watching changes in referenced projects, this will not work with proposed setup which creates a "final" `tsconfig.json` that referencs all the other packages. 

First will the ts-loader even work with proposed setup?

If yes then, is it possible to add the as a config option - enable watching referenced project files?

I will be happy to contribute to this  PR if required.

johnnyreilly:
Thanks for commenting @pragatisureka 

@andrewbranch could you respond? Thanks!

andrewbranch:
> Question: how are you managing to test given that ts-loader is webpack 4 minimum?

¯\\_(ツ)_/¯ Dunno, with the one line I changed, it works!

Regarding the tests, I have regenerated the expected output and it always comes back without that line for me, and running the test against that expected output also passes. Something strange is different between my machine and Travis.

@pragatisureka if you read through #815, there's some interest in having ts-loader actually act like `tsc --build`, but this PR doesn't aim to do that. This PR just avoids recompiling what's already been compiled in upstream projects, which is the exact same thing `tsc` will do if it has `references`.

Imagine that you've already run `tsc --build` once on your root tsconfig file, then you edit some TS files in some of the dependent projects. Now, if you run `tsc --build` again, it will of course rebuild the dependent projects that you edited. But if you just run `tsc`, it will _not_ rebuild the dependent projects, it will just build the root project again with the _old source_ of the dependent projects that was already written to disk. That's exactly what this PR is: `tsc`, and support for `--build` will hopefully come later.

The one sort of strange consequence of this is that if, like your example, you have basically 100% of your source code in dependent projects, `ts-loader` has virtually nothing to do with TypeScript, because TypeScript already just built your entire codebase and wrote it to disk. In that case, the only thing `ts-loader` needs to do is look up your various tsconfig files to figure out where the output is stored, and then read that output from disk. So it may well be true that this scenario is kind of a waste of `ts.createProgram()`, but I'm _assuming_ a TypeScript program from entirely project references is a lot cheaper to make than one from a bunch of unbuilt source code... but I'm not totally sure about that.

The point of the last few commits was basically this. Imagine this project structure:

```
tsconfig.json (with reference to lib)
app.ts
  lib/
    - lib.ts
    - lib.d.ts (built output)
    - lib.js   (built output)
    - tsconfig.json (composite)
```

When ts-loader sees a request for `lib/lib.ts`, it does the following:
 - Are you using project references? If yes,
 - Is there a referenced project that contains `lib/lib.ts`? If yes,
 - Find `lib/lib.js` and pass its contents to Webpack.

So basically, because `lib.js` already exists, ts-loader doesn't actually use `lib.ts` at all. (This should be a nice speed optimization.)

Here's what was happening _before_ the last few commits with this structure:

1. You start `webpack --watch`
2. ts-loader goes through the process I described above
3. You make edits to `lib/lib.ts`
4. Webpack recompiles, going through the process I described above again

See the problem? You edited `lib/lib.ts` which caused Webpack to recompile, but the compilation _still uses the JS file, which hasn't changed_. So this is already a nuisance, but let's continue:

5. You run `tsc --build` in another terminal tab
6. `lib/lib.js` is updated
7. Webpack does nothing

So this time, a source file that ts-loader _does_ depend on changed, but it didn't trigger a recompile. Not good.

TL;DR: The behavior we were getting before the commits you mentioned was _never_ right or desirable for anyone. I think the behavior you _want_ is to emulate `tsc --build`, where ts-loader would be able to rebuild upstream projects when they change. That's also something we want, I think, but not as part of _this_ PR, because this PR is already pretty big.

Concrete advice for your scenario once this PR lands: use `transpileOnly` with ts-loader, and then run `tsc --build --watch` in another process. That will keep ts-loader doing very minimal work while `tsc --build` does all the hard stuff. And because of the latest few commits you mentioned, doing that combo will cause webpack to recompile after `tsc` recompiles the upstream sources.

johnnyreilly:
@andrewbranch first of all well done on a stellar explanation just now. That's tremendously clear and helpful!

Regards the test failures it's kind of strange; on AppVeyor (Windows) and Travis (Linux) you have 2 test failures:

- projectReferences
- projectReferencesOutDir

But you say they pass locally? If so, maybe I'll take a whirl on my own machine and see what happens. I'm Windows 10; let's see what happens.

It could be good to have an execution test for project references. They can only really test the positive case "this thing compiles TypeScript and executes real JavaScript". You can't use them to test things like output directory etc. But they're much more reliable and consequently we run them on all historic versions of TypeScript that are supported (bounding the minimum with a minimum TypeScript version number where applicable).

So whilst we can only support projectReferences for TypeScript 3.0.x, when TypeScript 3.1 etc ships, if we have an execution test in place than that will continue to cover basic projectReferences support in the test pack. 

Would you be up for implementing one? I feel like you could probably clone https://github.com/TypeStrong/ts-loader/tree/master/test/execution-tests/basic and take the code for the existing projectReferences comparison test and augment the cloned test with that.

You can read more about execution tests here: https://github.com/TypeStrong/ts-loader/blob/master/test/execution-tests/README.md

andrewbranch:
@johnnyreilly I would be happy to! However, I'm not 100% sure when I'll be able to get to it, as normal work stuff outside ts-loader is about to get pretty crazy for a couple weeks. Hopefully I can do it one night this week but I'll have to play it by ear a bit.

johnnyreilly:
> However, I'm not 100% sure when I'll be able to get to it, as normal work stuff outside ts-loader is about to get pretty crazy for a couple weeks. 

Same here! Don't worry about timing. I'm invested in this and I feel we're pretty close to finished. 

I'll report back when I've had chance to test the comparison tests locally

pragatisureka:
@andrewbranch Thank you for explaining it so clearly. I think I will go with the solution you suggested. Running ts-loader with `transpileOnly` and using `tsc --build --watch`. 

In our setup we have multiple Typescript projects each with its own `ourDir` and any project can have dependency on any other project. Each project is bundled separately and loaded only if that particular feature is used in the UI. 

Hopefully I will find some time to contribute `tsc --build` like feature to ts-loader.

Kudos @andrewbranch and @johnnyreilly for great work on ts-loader

johnnyreilly:
Hey @andrewbranch,

So I took the code for a test drive on my own machine; all of the projectReferences tests are failing and from the looks of it legitimately so.

I regenerated the output of each of the tests like so:

```
yarn run comparison-tests -- --save-output --single-test projectReferences
yarn run comparison-tests -- --save-output --single-test projectReferencesNoSourceMap
yarn run comparison-tests -- --save-output --single-test projectReferencesNotBuilt
yarn run comparison-tests -- --save-output --single-test projectReferencesOutDir
```

This created all the output files for the comparison tests (including the missing `.transpiled` files - which is cool).  This allowed easy viewing of the difference between what's currently in the test output and what we expect. 

Each `output.txt` file featured an error along these lines:

```
[./lib/index.ts] 348 bytes {main} [built] [failed] [1 error]

ERROR in ./lib/index.ts
Module build failed: Error: TypeScript emitted no output for lib\index.ts.
    at makeSourceMapAndFinish (dist\index.js:71:15)
    at successLoader (dist\index.js:61:9)
    at Object.loader (dist\index.js:21:12)
 @ ./app.ts 3:12-28
```

So I think somewhere along the way something got broken... Any ideas?

johnnyreilly:
BTW I've merged upstream changes into your branch - remember to `git pull`!

andrewbranch:
It looks like there may be some extra work to do due to https://github.com/Microsoft/TypeScript/pull/27082 😐

andrewbranch:
> So I took the code for a test drive on my own machine; all of the projectReferences tests are failing and from the looks of it legitimately so.

@johnnyreilly are you on a Windows machine? My latest push looks like it's going to succeed on Travis, but the AppVeyor run is showing the same thing you got locally. Also, I still can’t get the transpiled output to be generated. Seems like there’s some x-plat fishiness going on. I have a Windows machine I can try this out on a little later.

Life is back to a normal pace now so I should have time to finish this up this week, if all goes well 🤞

johnnyreilly:
Yup I'm Windows. 

> Update: well, one of the Travis runs failed, but it seems unrelated? Maybe just flaked out?

Yeah - that happens. Requeued it now...

> Life is back to a normal pace now so I should have time to finish this up this week, if all goes well 🤞

Awesome!

johnnyreilly:
> It looks like there may be some extra work to do due to Microsoft/TypeScript#27082 😐

GAH!

andrewbranch:
capture

😑 😑 😑

src/utils.ts
```diff
@@ -294,7 +294,11 @@ function getProjectReferenceForFile(filePath: string, instance: TSInstance) {
         .getProjectReferences()!
         .find(
           ref =>
-            (ref && ref.commandLine.fileNames.indexOf(filePath) > -1) || false
+            (ref &&
+              ref.commandLine.fileNames.some(
+                file => path.normalize(file) === filePath
```
andrewbranch: Running `typescript.normalizePath` on `filePath` would be more straightforward, but that function seems to be private.

andrewbranch:
Ok, after fixing the path comparison that was _actually_ broken on Windows, it looks like Windows was also treating a couple paths in error messages differently than *nix, so I changed every error message I introduced to use `path.relative`.

However, there’s one error message that comes from TypeScript itself, so there’s nothing I can do about it:

```diff
-      TS6305: Output file '/.test/projectReferencesNotBuilt/lib/index.d.ts' has not been built from source file '/.test/projectReferencesNotBuilt/lib/index.ts'.
+      TS6305: Output file 'lib/index.d.ts' has not been built from source file 'lib/index.ts'.
```

johnnyreilly:
So that's a pain. But it's fine I think. We just need to tweak the comparison test pack to ignore the start of the path and only check the end.

This is not my first rodeo: https://github.com/TypeStrong/ts-loader/blob/master/test/comparison-tests/create-and-execute-test.js#L381

Regrettably the test pack has to do this a *lot*. Just one more case to cater for... Regex the way to glory!

andrewbranch:
@johnnyreilly could you see if you can figure out what's going on with the one remaining comparison test failing on Windows only? I'm starting to lose my mind a little with it. The bundle itself is perfect, but on Windows the `output.txt` includes that diagnostics message from the TS compiler. Which I'm guessing means that on Windows, with project references, with an `outDir`, type checking isn't working via the loader. Every other combination of every option and every operating system is now working and I don't understand 😓

johnnyreilly:
So it looks like the comparison test is fine. I'm wondering if this might actually be a TypeScript bug. I'd like to try upgrading to TypeScript 3.0.3 to see if it was fixed in the patch releases. Failing that it'd be good to see what the inputs are on Windows vs Linux for this test. If they're the same then my guess is it's a compiler issue.

I'll try these today and report back.

johnnyreilly:
One more thought: whilst this PR has been in progress another PR specifically related to `outDir` was merged: https://github.com/TypeStrong/ts-loader/pull/822

I wonder if that may have some bearing on this.

johnnyreilly:
I've forked your branch, upgraded to TypeScript 3.0.3 and tested locally.  Status:

```
yarn run comparison-tests -- --single-test projectReferences # passes!
yarn run comparison-tests -- --single-test projectReferencesNoSourceMap # passes!
yarn run comparison-tests -- --single-test projectReferencesNotBuilt # passes!
yarn run comparison-tests -- --single-test projectReferencesOutDir # fails!
```

Now begins the long dark tea-time of the soul 😉

johnnyreilly:
So I've hacked some logging in - see my fork of your branch [here](https://github.com/andrewbranch/ts-loader/pull/1).

With that in place I've done a `yarn run comparison-tests -- --single-test projectReferencesOutDir` and this was the output:

```
  projectReferencesOutDir
-----------------
filePath C:\work\ts-loader\.test\projectReferencesOutDir\app.ts
setting file.projectReference.project to  undefined
projectReference undefined

-----------------
filePath C:\work\ts-loader\.test\projectReferencesOutDir\lib\index.ts
setting file.projectReference.project to  { commandLine:
   { options:
      { composite: true,
        sourceMap: true,
        outDir: 'C:/work/ts-loader/.test/projectReferencesOutDir/lib/out',
        configFilePath:
         'C:/work/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json' },
     fileNames:
      [ 'C:/work/ts-loader/.test/projectReferencesOutDir/lib/index.ts' ],
     projectReferences: undefined,
     typeAcquisition: { enable: false, include: [], exclude: [] },
     raw: { compilerOptions: [Object], files: [Array] },
     errors: [],
     wildcardDirectories: {},
     compileOnSave: false,
     configFileSpecs:
      { filesSpecs: [Array],
        referencesSpecs: undefined,
        includeSpecs: undefined,
        excludeSpecs: [Array],
        validatedIncludeSpecs: undefined,
        validatedExcludeSpecs: [Array],
        wildcardDirectories: {} } },
  sourceFile:
   SourceFileObject {
     pos: 0,
     end: 142,
     flags: 16842752,
     transformFlags: undefined,
     parent: undefined,
     kind: 277,
     text:
      '{\r\n  "compilerOptions": {\r\n    "composite": true,\r\n    "sourceMap": true,\r\n    "outDir": "./out"\r\n  },\r\n  "files": [\r\n    "./index.ts"\r\n  ]\r\n}',
     bindDiagnostics: [],
     bindSuggestionDiagnostics: undefined,
     languageVersion: 2,
     fileName:
      'C:/work/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json',
     languageVariant: 1,
     isDeclarationFile: false,
     scriptKind: 6,
     statements: [ [NodeObject], pos: 0, end: 142 ],
     endOfFileToken:
      TokenObject { pos: 142, end: 142, flags: 16842752, parent: undefined, kind: 1 },
     parseDiagnostics: [],
     version: '0',
     scriptSnapshot:
      StringScriptSnapshot {
        text:
         '{\r\n  "compilerOptions": {\r\n    "composite": true,\r\n    "sourceMap": true,\r\n    "outDir": "./out"\r\n  },\r\n  "files": [\r\n    "./index.ts"\r\n  ]\r\n}' } } }
projectReference { commandLine:
   { options:
      { composite: true,
        sourceMap: true,
        outDir: 'C:/work/ts-loader/.test/projectReferencesOutDir/lib/out',
        configFilePath:
         'C:/work/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json' },
     fileNames:
      [ 'C:/work/ts-loader/.test/projectReferencesOutDir/lib/index.ts' ],
     projectReferences: undefined,
     typeAcquisition: { enable: false, include: [], exclude: [] },
     raw: { compilerOptions: [Object], files: [Array] },
     errors: [],
     wildcardDirectories: {},
     compileOnSave: false,
     configFileSpecs:
      { filesSpecs: [Array],
        referencesSpecs: undefined,
        includeSpecs: undefined,
        excludeSpecs: [Array],
        validatedIncludeSpecs: undefined,
        validatedExcludeSpecs: [Array],
        wildcardDirectories: {} } },
  sourceFile:
   SourceFileObject {
     pos: 0,
     end: 142,
     flags: 16842752,
     transformFlags: undefined,
     parent: undefined,
     kind: 277,
     text:
      '{\r\n  "compilerOptions": {\r\n    "composite": true,\r\n    "sourceMap": true,\r\n    "outDir": "./out"\r\n  },\r\n  "files": [\r\n    "./index.ts"\r\n  ]\r\n}',
     bindDiagnostics: [],
     bindSuggestionDiagnostics: undefined,
     languageVersion: 2,
     fileName:
      'C:/work/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json',
     languageVariant: 1,
     isDeclarationFile: false,
     scriptKind: 6,
     statements: [ [NodeObject], pos: 0, end: 142 ],
     endOfFileToken:
      TokenObject { pos: 142, end: 142, flags: 16842752, parent: undefined, kind: 1 },
     parseDiagnostics: [],
     version: '0',
     scriptSnapshot:
      StringScriptSnapshot {
        text:
         '{\r\n  "compilerOptions": {\r\n    "composite": true,\r\n    "sourceMap": true,\r\n    "outDir": "./out"\r\n  },\r\n  "files": [\r\n    "./index.ts"\r\n  ]\r\n}' } } }

-----------------
options.outDir C:/work/ts-loader/.test/projectReferencesOutDir/lib/out
projectDirectory C:/work/ts-loader/.test/projectReferencesOutDir/lib
relativePath index.ts
newExtension .js
outputJavaScriptFileName C:\work\ts-loader\.test\projectReferencesOutDir\lib\out\index.js

-----------------
filePath C:\work\ts-loader\.test\projectReferencesOutDir\lib\index.ts
file.projectReference.project { commandLine:
   { options:
      { composite: true,
        sourceMap: true,
        outDir: 'C:/work/ts-loader/.test/projectReferencesOutDir/lib/out',
        configFilePath:
         'C:/work/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json' },
     fileNames:
      [ 'C:/work/ts-loader/.test/projectReferencesOutDir/lib/index.ts' ],
     projectReferences: undefined,
     typeAcquisition: { enable: false, include: [], exclude: [] },
     raw: { compilerOptions: [Object], files: [Array] },
     errors: [],
     wildcardDirectories: {},
     compileOnSave: false,
     configFileSpecs:
      { filesSpecs: [Array],
        referencesSpecs: undefined,
        includeSpecs: undefined,
        excludeSpecs: [Array],
        validatedIncludeSpecs: undefined,
        validatedExcludeSpecs: [Array],
        wildcardDirectories: {} } },
  sourceFile:
   SourceFileObject {
     pos: 0,
     end: 142,
     flags: 16842752,
     transformFlags: undefined,
     parent: undefined,
     kind: 277,
     text:
      '{\r\n  "compilerOptions": {\r\n    "composite": true,\r\n    "sourceMap": true,\r\n    "outDir": "./out"\r\n  },\r\n  "files": [\r\n    "./index.ts"\r\n  ]\r\n}',
     bindDiagnostics: [],
     bindSuggestionDiagnostics: undefined,
     languageVersion: 2,
     fileName:
      'C:/work/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json',
     languageVariant: 1,
     isDeclarationFile: false,
     scriptKind: 6,
     statements: [ [NodeObject], pos: 0, end: 142 ],
     endOfFileToken:
      TokenObject { pos: 142, end: 142, flags: 16842752, parent: undefined, kind: 1 },
     parseDiagnostics: [],
     version: '0',
     scriptSnapshot:
      StringScriptSnapshot {
        text:
         '{\r\n  "compilerOptions": {\r\n    "composite": true,\r\n    "sourceMap": true,\r\n    "outDir": "./out"\r\n  },\r\n  "files": [\r\n    "./index.ts"\r\n  ]\r\n}' } } }
-----------------
referencedProject.sourceFile.fileName C:/work/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json
loader.rootContext C:\work\ts-loader\.test\projectReferencesOutDir\lib\index.ts

```

I'd be interested in the difference between this and what the operation produces on a Mac. `setting file.projectReference.project to  undefined` is particularly curious to me....

johnnyreilly:
tried it in wsl; nothing useful to report.

```
  projectReferencesOutDir
-----------------
filePath /mnt/c/source/ts-loader/.test/projectReferencesOutDir/app.ts
setting file.projectReference.project to  undefined
projectReference undefined

-----------------
filePath /mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/index.ts
setting file.projectReference.project to  { commandLine: 
   { options: 
      { composite: true,
        sourceMap: true,
        outDir: '/mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/out',
        configFilePath: '/mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json' },
     fileNames: 
      [ '/mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/index.ts' ],
     projectReferences: undefined,
     typeAcquisition: { enable: false, include: [], exclude: [] },
     raw: { compilerOptions: [Object], files: [Array] },
     errors: [],
     wildcardDirectories: {},
     compileOnSave: false,
     configFileSpecs: 
      { filesSpecs: [Array],
        referencesSpecs: undefined,
        includeSpecs: undefined,
        excludeSpecs: [Array],
        validatedIncludeSpecs: undefined,
        validatedExcludeSpecs: [Array],
        wildcardDirectories: {} } },
  sourceFile: 
   SourceFileObject {
     pos: 0,
     end: 133,
     flags: 16842752,
     transformFlags: undefined,
     parent: undefined,
     kind: 277,
     text: '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "./out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}',
     bindDiagnostics: [],
     bindSuggestionDiagnostics: undefined,
     languageVersion: 2,
     fileName: '/mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json',
     languageVariant: 1,
     isDeclarationFile: false,
     scriptKind: 6,
     statements: [ [Object], pos: 0, end: 133 ],
     endOfFileToken: TokenObject { pos: 133, end: 133, flags: 16842752, parent: undefined, kind: 1 },
     parseDiagnostics: [],
     version: '0',
     scriptSnapshot: 
      StringScriptSnapshot {
        text: '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "./out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}' } } }
projectReference { commandLine: 
   { options: 
      { composite: true,
        sourceMap: true,
        outDir: '/mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/out',
        configFilePath: '/mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json' },
     fileNames: 
      [ '/mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/index.ts' ],
     projectReferences: undefined,
     typeAcquisition: { enable: false, include: [], exclude: [] },
     raw: { compilerOptions: [Object], files: [Array] },
     errors: [],
     wildcardDirectories: {},
     compileOnSave: false,
     configFileSpecs: 
      { filesSpecs: [Array],
        referencesSpecs: undefined,
        includeSpecs: undefined,
        excludeSpecs: [Array],
        validatedIncludeSpecs: undefined,
        validatedExcludeSpecs: [Array],
        wildcardDirectories: {} } },
  sourceFile: 
   SourceFileObject {
     pos: 0,
     end: 133,
     flags: 16842752,
     transformFlags: undefined,
     parent: undefined,
     kind: 277,
     text: '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "./out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}',
     bindDiagnostics: [],
     bindSuggestionDiagnostics: undefined,
     languageVersion: 2,
     fileName: '/mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json',
     languageVariant: 1,
     isDeclarationFile: false,
     scriptKind: 6,
     statements: [ [Object], pos: 0, end: 133 ],
     endOfFileToken: TokenObject { pos: 133, end: 133, flags: 16842752, parent: undefined, kind: 1 },
     parseDiagnostics: [],
     version: '0',
     scriptSnapshot: 
      StringScriptSnapshot {
        text: '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "./out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}' } } }

-----------------
options.outDir /mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/out
projectDirectory /mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib
relativePath index.ts
newExtension .js
outputJavaScriptFileName /mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/out/index.js

-----------------
filePath /mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/index.ts
file.projectReference.project { commandLine: 
   { options: 
      { composite: true,
        sourceMap: true,
        outDir: '/mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/out',
        configFilePath: '/mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json' },
     fileNames: 
      [ '/mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/index.ts' ],
     projectReferences: undefined,
     typeAcquisition: { enable: false, include: [], exclude: [] },
     raw: { compilerOptions: [Object], files: [Array] },
     errors: [],
     wildcardDirectories: {},
     compileOnSave: false,
     configFileSpecs: 
      { filesSpecs: [Array],
        referencesSpecs: undefined,
        includeSpecs: undefined,
        excludeSpecs: [Array],
        validatedIncludeSpecs: undefined,
        validatedExcludeSpecs: [Array],
        wildcardDirectories: {} } },
  sourceFile: 
   SourceFileObject {
     pos: 0,
     end: 133,
     flags: 16842752,
     transformFlags: undefined,
     parent: undefined,
     kind: 277,
     text: '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "./out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}',
     bindDiagnostics: [],
     bindSuggestionDiagnostics: undefined,
     languageVersion: 2,
     fileName: '/mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json',
     languageVariant: 1,
     isDeclarationFile: false,
     scriptKind: 6,
     statements: [ [Object], pos: 0, end: 133 ],
     endOfFileToken: TokenObject { pos: 133, end: 133, flags: 16842752, parent: undefined, kind: 1 },
     parseDiagnostics: [],
     version: '0',
     scriptSnapshot: 
      StringScriptSnapshot {
        text: '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "./out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}' } } }
-----------------
referencedProject.sourceFile.fileName /mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json
loader.rootContext /mnt/c/source/ts-loader/.test/projectReferencesOutDir/lib/index.ts
```

andrewbranch:
macOS:

```
  projectReferencesOutDir
-----------------
filePath /Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/app.ts
setting file.projectReference.project to  undefined
projectReference undefined

-----------------
filePath /Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/index.ts
setting file.projectReference.project to  { commandLine: 
   { options: 
      { composite: true,
        sourceMap: true,
        outDir: '/Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/out',
        configFilePath: '/Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json' },
     fileNames: 
      [ '/Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/index.ts' ],
     projectReferences: undefined,
     typeAcquisition: { enable: false, include: [], exclude: [] },
     raw: { compilerOptions: [Object], files: [Array] },
     errors: [],
     wildcardDirectories: {},
     compileOnSave: false,
     configFileSpecs: 
      { filesSpecs: [Array],
        referencesSpecs: undefined,
        includeSpecs: undefined,
        excludeSpecs: [Array],
        validatedIncludeSpecs: undefined,
        validatedExcludeSpecs: [Array],
        wildcardDirectories: {} } },
  sourceFile: 
   SourceFileObject {
     pos: 0,
     end: 133,
     flags: 16842752,
     transformFlags: undefined,
     parent: undefined,
     kind: 277,
     text: '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "./out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}',
     bindDiagnostics: [],
     bindSuggestionDiagnostics: undefined,
     languageVersion: 2,
     fileName: '/Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json',
     languageVariant: 1,
     isDeclarationFile: false,
     scriptKind: 6,
     statements: [ [Object], pos: 0, end: 133 ],
     endOfFileToken: TokenObject { pos: 133, end: 133, flags: 16842752, parent: undefined, kind: 1 },
     parseDiagnostics: [],
     version: '0',
     scriptSnapshot: 
      StringScriptSnapshot {
        text: '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "./out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}' } } }
projectReference { commandLine: 
   { options: 
      { composite: true,
        sourceMap: true,
        outDir: '/Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/out',
        configFilePath: '/Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json' },
     fileNames: 
      [ '/Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/index.ts' ],
     projectReferences: undefined,
     typeAcquisition: { enable: false, include: [], exclude: [] },
     raw: { compilerOptions: [Object], files: [Array] },
     errors: [],
     wildcardDirectories: {},
     compileOnSave: false,
     configFileSpecs: 
      { filesSpecs: [Array],
        referencesSpecs: undefined,
        includeSpecs: undefined,
        excludeSpecs: [Array],
        validatedIncludeSpecs: undefined,
        validatedExcludeSpecs: [Array],
        wildcardDirectories: {} } },
  sourceFile: 
   SourceFileObject {
     pos: 0,
     end: 133,
     flags: 16842752,
     transformFlags: undefined,
     parent: undefined,
     kind: 277,
     text: '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "./out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}',
     bindDiagnostics: [],
     bindSuggestionDiagnostics: undefined,
     languageVersion: 2,
     fileName: '/Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json',
     languageVariant: 1,
     isDeclarationFile: false,
     scriptKind: 6,
     statements: [ [Object], pos: 0, end: 133 ],
     endOfFileToken: TokenObject { pos: 133, end: 133, flags: 16842752, parent: undefined, kind: 1 },
     parseDiagnostics: [],
     version: '0',
     scriptSnapshot: 
      StringScriptSnapshot {
        text: '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "./out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}' } } }

-----------------
options.outDir /Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/out
projectDirectory /Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib
relativePath index.ts
newExtension .js
outputJavaScriptFileName /Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/out/index.js

-----------------
filePath /Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/index.ts
file.projectReference.project { commandLine: 
   { options: 
      { composite: true,
        sourceMap: true,
        outDir: '/Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/out',
        configFilePath: '/Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json' },
     fileNames: 
      [ '/Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/index.ts' ],
     projectReferences: undefined,
     typeAcquisition: { enable: false, include: [], exclude: [] },
     raw: { compilerOptions: [Object], files: [Array] },
     errors: [],
     wildcardDirectories: {},
     compileOnSave: false,
     configFileSpecs: 
      { filesSpecs: [Array],
        referencesSpecs: undefined,
        includeSpecs: undefined,
        excludeSpecs: [Array],
        validatedIncludeSpecs: undefined,
        validatedExcludeSpecs: [Array],
        wildcardDirectories: {} } },
  sourceFile: 
   SourceFileObject {
     pos: 0,
     end: 133,
     flags: 16842752,
     transformFlags: undefined,
     parent: undefined,
     kind: 277,
     text: '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "./out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}',
     bindDiagnostics: [],
     bindSuggestionDiagnostics: undefined,
     languageVersion: 2,
     fileName: '/Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json',
     languageVariant: 1,
     isDeclarationFile: false,
     scriptKind: 6,
     statements: [ [Object], pos: 0, end: 133 ],
     endOfFileToken: TokenObject { pos: 133, end: 133, flags: 16842752, parent: undefined, kind: 1 },
     parseDiagnostics: [],
     version: '0',
     scriptSnapshot: 
      StringScriptSnapshot {
        text: '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "./out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}' } } }
-----------------
referencedProject.sourceFile.fileName /Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json
loader.rootContext /Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/index.ts
```

johnnyreilly:
Just pushed a commit to do some more logging...

johnnyreilly:
```
  projectReferencesOutDir
-----------------
filePath C:\source\ts-loader\.test\projectReferencesOutDir\app.ts
setting file.projectReference.project to  undefined
projectReference undefined

-----------------
filePath C:\source\ts-loader\.test\projectReferencesOutDir\lib\index.ts
setting file.projectReference.project to  { commandLine:
   { options:
      { composite: true,
        sourceMap: true,
        outDir: 'C:/source/ts-loader/.test/projectReferencesOutDir/lib/out',
        configFilePath:
         'C:/source/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json' },
     fileNames:
      [ 'C:/source/ts-loader/.test/projectReferencesOutDir/lib/index.ts' ],
     projectReferences: undefined,
     typeAcquisition: { enable: false, include: [], exclude: [] },
     raw: { compilerOptions: [Object], files: [Array] },
     errors: [],
     wildcardDirectories: {},
     compileOnSave: false,
     configFileSpecs:
      { filesSpecs: [Array],
        referencesSpecs: undefined,
        includeSpecs: undefined,
        excludeSpecs: [Array],
        validatedIncludeSpecs: undefined,
        validatedExcludeSpecs: [Array],
        wildcardDirectories: {} } },
  sourceFile:
   SourceFileObject {
     pos: 0,
     end: 131,
     flags: 16842752,
     transformFlags: undefined,
     parent: undefined,
     kind: 277,
     text:
      '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}',
     bindDiagnostics: [],
     bindSuggestionDiagnostics: undefined,
     languageVersion: 2,
     fileName:
      'C:/source/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json',
     languageVariant: 1,
     isDeclarationFile: false,
     scriptKind: 6,
     statements: [ [NodeObject], pos: 0, end: 131 ],
     endOfFileToken:
      TokenObject { pos: 131, end: 131, flags: 16842752, parent: undefined, kind: 1 },
     parseDiagnostics: [],
     version: '0',
     scriptSnapshot:
      StringScriptSnapshot {
        text:
         '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}' } } }
projectReference { commandLine:
   { options:
      { composite: true,
        sourceMap: true,
        outDir: 'C:/source/ts-loader/.test/projectReferencesOutDir/lib/out',
        configFilePath:
         'C:/source/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json' },
     fileNames:
      [ 'C:/source/ts-loader/.test/projectReferencesOutDir/lib/index.ts' ],
     projectReferences: undefined,
     typeAcquisition: { enable: false, include: [], exclude: [] },
     raw: { compilerOptions: [Object], files: [Array] },
     errors: [],
     wildcardDirectories: {},
     compileOnSave: false,
     configFileSpecs:
      { filesSpecs: [Array],
        referencesSpecs: undefined,
        includeSpecs: undefined,
        excludeSpecs: [Array],
        validatedIncludeSpecs: undefined,
        validatedExcludeSpecs: [Array],
        wildcardDirectories: {} } },
  sourceFile:
   SourceFileObject {
     pos: 0,
     end: 131,
     flags: 16842752,
     transformFlags: undefined,
     parent: undefined,
     kind: 277,
     text:
      '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}',
     bindDiagnostics: [],
     bindSuggestionDiagnostics: undefined,
     languageVersion: 2,
     fileName:
      'C:/source/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json',
     languageVariant: 1,
     isDeclarationFile: false,
     scriptKind: 6,
     statements: [ [NodeObject], pos: 0, end: 131 ],
     endOfFileToken:
      TokenObject { pos: 131, end: 131, flags: 16842752, parent: undefined, kind: 1 },
     parseDiagnostics: [],
     version: '0',
     scriptSnapshot:
      StringScriptSnapshot {
        text:
         '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}' } } }

-----------------
options.outDir C:/source/ts-loader/.test/projectReferencesOutDir/lib/out
projectDirectory C:/source/ts-loader/.test/projectReferencesOutDir/lib
relativePath index.ts
newExtension .js
outputJavaScriptFileName C:\source\ts-loader\.test\projectReferencesOutDir\lib\out\index.js

-----------------
filePath C:\source\ts-loader\.test\projectReferencesOutDir\lib\index.ts
file.projectReference.project { commandLine:
   { options:
      { composite: true,
        sourceMap: true,
        outDir: 'C:/source/ts-loader/.test/projectReferencesOutDir/lib/out',
        configFilePath:
         'C:/source/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json' },
     fileNames:
      [ 'C:/source/ts-loader/.test/projectReferencesOutDir/lib/index.ts' ],
     projectReferences: undefined,
     typeAcquisition: { enable: false, include: [], exclude: [] },
     raw: { compilerOptions: [Object], files: [Array] },
     errors: [],
     wildcardDirectories: {},
     compileOnSave: false,
     configFileSpecs:
      { filesSpecs: [Array],
        referencesSpecs: undefined,
        includeSpecs: undefined,
        excludeSpecs: [Array],
        validatedIncludeSpecs: undefined,
        validatedExcludeSpecs: [Array],
        wildcardDirectories: {} } },
  sourceFile:
   SourceFileObject {
     pos: 0,
     end: 131,
     flags: 16842752,
     transformFlags: undefined,
     parent: undefined,
     kind: 277,
     text:
      '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}',
     bindDiagnostics: [],
     bindSuggestionDiagnostics: undefined,
     languageVersion: 2,
     fileName:
      'C:/source/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json',
     languageVariant: 1,
     isDeclarationFile: false,
     scriptKind: 6,
     statements: [ [NodeObject], pos: 0, end: 131 ],
     endOfFileToken:
      TokenObject { pos: 131, end: 131, flags: 16842752, parent: undefined, kind: 1 },
     parseDiagnostics: [],
     version: '0',
     scriptSnapshot:
      StringScriptSnapshot {
        text:
         '{\n  "compilerOptions": {\n    "composite": true,\n    "sourceMap": true,\n    "outDir": "out"\n  },\n  "files": [\n    "./index.ts"\n  ]\n}' } } }
-----------------
referencedProject.sourceFile.fileName C:/source/ts-loader/.test/projectReferencesOutDir/lib/tsconfig.json
filePath C:\source\ts-loader\.test\projectReferencesOutDir\lib\index.ts
relativeProjectConfigPath lib\tsconfig.json
relativeFilePath lib\index.ts
jsFileName C:\source\ts-loader\.test\projectReferencesOutDir\lib\out\index.js
relativeJSFileName lib\out\index.js
jsFileName C:\source\ts-loader\.test\projectReferencesOutDir\lib\out\index.js

```

johnnyreilly:
It all looks healthy I think.  I'm wondering if this a TypeScript compiler bug that affects Windows only.

andrewbranch:
```
filePath and sourceFile /Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/app.ts undefined
filePath and sourceFile /Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/index.ts undefined
filePath and sourceFile /Users/Andrew/Developer/ts-loader/.test/projectReferencesOutDir/lib/out/index.d.ts undefined
filePath and sourceFile /Users/Andrew/Developer/ts-loader/node_modules/@types/braces/index.d.ts undefined
filePath and sourceFile /Users/Andrew/Developer/ts-loader/node_modules/@types/micromatch/index.d.ts undefined
filePath and sourceFile /Users/Andrew/Developer/ts-loader/node_modules/@types/node/index.d.ts undefined
filePath and sourceFile /Users/Andrew/Developer/ts-loader/node_modules/@types/node/inspector.d.ts undefined
filePath and sourceFile /Users/Andrew/Developer/ts-loader/node_modules/@types/semver/index.d.ts undefined
filePath and sourceFile /Users/Andrew/Developer/ts-loader/node_modules/typescript/lib/lib.d.ts undefined
filePath and sourceFile /Users/Andrew/Developer/ts-loader/node_modules/typescript/lib/lib.es5.d.ts undefined
filePath and sourceFile /Users/Andrew/Developer/ts-loader/node_modules/typescript/lib/lib.dom.d.ts undefined
filePath and sourceFile /Users/Andrew/Developer/ts-loader/node_modules/typescript/lib/lib.webworker.importscripts.d.ts undefined
filePath and sourceFile /Users/Andrew/Developer/ts-loader/node_modules/typescript/lib/lib.scripthost.d.ts undefined
```

johnnyreilly:
Here's what I'm thinking:

- We're 99.99% the way there
- I think the 0.01% is probably a `tsc` problem
- I'd like to release this functionality even given the known issue.  I think it adds value and I'm optimistic that this issue will be resolved in the future; whether by us, the TypeScript team or some other kind soul.

What do you think?  Good idea?

If so, would you be up for reporting this issue against the TypeScript project. I hope they'll take a look at it.  Also, could you update the README to put some information about this issue.  Feel free to link to this PR etc for context.

In the meantime I could come up with a mechanism to ignore the test on Windows for now.

andrewbranch:
@johnnyreilly sounds like a good idea to me. I'm in favor of getting this in, as the longer it stays open, the more work it seems to create for me by stuff changing underneath me 😄 

I would like to see if 
https://github.com/Microsoft/TypeScript/pull/27082 breaks us in the meantime and work around it if so. I think it shouldn't be too much trouble. Will try to work on it tonight.

andrewbranch:
Ok, I wasn't able to reliably test against `typescript@master` because it doesn't seem to build (😳), but based on some semi-intelligent guessing and source-code-reading, I think I've covered it with the latest commit? Feel free to ping me once https://github.com/Microsoft/TypeScript/pull/27082 lands in typescript@next to double check. I'll update the README tomorrow. We're close 🤞.

andrewbranch:
> @andrewbranch that change was published in typescript@next a few days ago

This doesn't seem to be the case insofar as what `npm i typescript@next` does. There’s no instance of `getResolvedProjectReference` in `node_modules/typescript/lib/typescript.js`. I also tried installing `typescript@master` and building from source, but master fails to build for me.

At any rate, https://github.com/TypeStrong/ts-loader/pull/817/commits/ddc5cc722ccbf068c0e6dd4ffa2be6f4163694b9 runs `getResolvedProjectReferences` instead of `getProjectReferences` if it exists, so I think that should be ok.

> Am I free to hack on your branch?

By all means!

> I'll work on ignoring the comparison test on Windows. I'll also write some execution tests so we have some in place.

🙏

ajafff:
I'm already using `program.getResolvedProjectReferences()` and it compiles just fine with `typescript@3.1.0-dev.20180915`

https://github.com/fimbullinter/wotan/blob/0518a9a78ed0f38c490cde1533b00d7681821c99/packages/wotan/src/runner.ts#L327-L330

andrewbranch:
Ok, cool, thanks @ajafff. Confirmed that `typescript@3.1.0-dev.20180915` works with the new API; `typescript@3.1.0-dev.20180920` I think was just broken ¯\_(ツ)_/¯ 

Working on the README now.

andrewbranch:
@johnnyreilly as I'm writing this README, I'm thinking that it honestly sounds kind of silly at this point without yet having support for _building_ upstream projects. I think probably very few people will want to try out the feature before there's a solid solution for having everything built for you.

A somewhat interesting effect of ts-loader not yet supporting project references in `master` is that people who have project references in their tsconfig and are using ts-loader on those projects today are not _actually_ using project references, but everything should be working perfectly for them, because each file request is just getting resolved to the source file, just as if they didn't have `references` in their tsconfig at all.

If we ship this now, enabled by default, without support for rebuilding upstream projects, we'll technically break those people, because all of a sudden, project references will start half-working instead of being ignored. Folks may not realize that ts-loader _stopped_ building their files.

I'm thinking we need to put all this work behind a loader option. What do you think?

johnnyreilly:
Oooh - good catch!  Yes that makes sense.  If you could include that information in the README that would be good.  What do you think would be a good name for the feature flag? `projectReferences`?

andrewbranch:
Awesome. After so much work, it's slightly disappointing that it’s not super usable yet, but I really think the hardest part is done. (🤞🙈) I may not be able to pick back up right away, but if nobody else jumps on it, I’ll want to see it through and figure out building upstream projects. That said, I’d be thrilled if someone else _does_ want to champion that part, but my own sense of sunk cost won’t let me leave it half-done. 😄

johnnyreilly:
> Working on ignoring the test on Windows now.

I'd forgotten just how fiddly / flaky the comparison test pack can be. Getting some false failures with my tweaked solution. Need to iterate on it a bit. Heading into work now; will try and take another look tonight / this weekend

johnnyreilly:
Cool - that worked.  I'll see if I can make a start on execution tests this weekend.

johnnyreilly:
Hey @andrewbranch ,

It's time to 🚢!  I'm planning to push out this as version 5.2.0.  I'll put a note in the README about the `outDir` Windows issue and probably put together an example repo too.

I'm going to write a blogpost to let people know about this new feature.  I'm planning to plagiarise your excellent update to the README pretty heavily.  I hope that's okay!

Thanks again for all your hard work. It's greatly appreciated!  Much ❤️

andrewbranch:
Woop! 🙌🏼 Thanks for all the support throughout the process. This has been fun! Plagiarize away! 😄

Issue #820, opened 2018-08-14
Support for TypeScript 3 Declaration map 


JonWallsten:
I'm going to be honest, I don't fully comprehend how ts-loader interacts with the typescript compiler (API?) so I'm not sure when/where in the chain this has to be corrected. But when enabling declarationMap: true in tsconfig and building with Webpack/ts-loader I don't get any map files. 
When running the typescript compiler, tsc, I get those files. 
I'm not sure if ts-loader isn't submitting the files or if I have to collect them somehow in my Webpack config. Maybe someone can enlighten me?

### Expected Behaviour
When setting declarationMap: true in tsconfig I expect the d.ts.map file to be in my dist folder
### Actual Behaviour
Only the d.ts files are ends up in the dist folder
### Steps to Reproduce the Problem
Use latest webpack and ts-loader and enable declaration and declarationMap
### Location of a Minimal Repository that Demonstrates the Issue.
Should be reproduces with any Webpack 4/Typescript@3.0.x starter repo.

johnnyreilly:
Do you have `sourcemap: true` in your `tsconfig.json`

JonWallsten:
@johnnyreilly Yeah! I get the maps when I run "tsc", I just don't get them when I run webpack/ts-loader.

JonWallsten:
Fixed by https://github.com/TypeStrong/ts-loader/pull/821

PR #821, opened 2018-08-15
Added support for TypeScript declaration map

JonWallsten:
@johnnyreilly: Please check this out and let me know if you want any changes or if I did something wrong creating the tests.

Fix for https://github.com/TypeStrong/ts-loader/issues/820

### DIFF
--- a/src/after-compile.ts
+++ b/src/after-compile.ts
@@ -1,5 +1,5 @@
 import * as path from 'path';
-
+import * as typescript from 'typescript'
 import { collectAllDependants, formatErrors } from './utils';
 import * as constants from './constants';
 import {
@@ -246,18 +246,20 @@ function provideDeclarationFilesToWebpack(
     }
 
     const outputFiles = getEmitOutput(instance, filePath);
-    const declarationFile = outputFiles
-      .filter(outputFile => outputFile.name.match(constants.dtsDtsxRegex))
-      .pop();
-    if (declarationFile !== undefined) {
-      const assetPath = path.relative(
-        compilation.compiler.context,
-        declarationFile.name
-      );
-      compilation.assets[assetPath] = {
-        source: () => declarationFile.text,
-        size: () => declarationFile.text.length
-      };
+    const declarationFiles: typescript.OutputFile[] = outputFiles
+      .filter((outputFile: typescript.OutputFile) => outputFile.name.match(constants.dtsDtsxRegex));
+
+    if (Array.isArray(declarationFiles)) {
+      declarationFiles.forEach(file => {
+        const assetPath = path.relative(
+          compilation.compiler.context,
+          file.name
+        );
+        compilation.assets[assetPath] = {
+          source: () => file.text,
+          size: () => file.text.length
+        };
+      });
     }
   }
 }
--- a/src/after-compile.ts
+++ b/src/after-compile.ts
@@ -1,5 +1,4 @@
 import * as path from 'path';
-import * as typescript from 'typescript'
 import { collectAllDependants, formatErrors } from './utils';
 import * as constants from './constants';
 import {
@@ -246,21 +245,20 @@ function provideDeclarationFilesToWebpack(
     }
 
     const outputFiles = getEmitOutput(instance, filePath);
-    const declarationFiles: typescript.OutputFile[] = outputFiles
-      .filter((outputFile: typescript.OutputFile) => outputFile.name.match(constants.dtsDtsxRegex));
+    const declarationFiles = outputFiles.filter(outputFile =>
+      outputFile.name.match(constants.dtsDtsxOrDtsDtsxMapRegex)
+    );
 
-    if (Array.isArray(declarationFiles)) {
-      declarationFiles.forEach(file => {
-        const assetPath = path.relative(
-          compilation.compiler.context,
-          file.name
-        );
-        compilation.assets[assetPath] = {
-          source: () => file.text,
-          size: () => file.text.length
-        };
-      });
-    }
+    declarationFiles.forEach(declarationFile => {
+      const assetPath = path.relative(
+        compilation.compiler.context,
+        declarationFile.name
+      );
+      compilation.assets[assetPath] = {
+        source: () => declarationFile.text,
+        size: () => declarationFile.text.length
+      };
+    });
   }
 }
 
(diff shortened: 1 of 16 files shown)


johnnyreilly:
First of all, thanks for the PR!

Question; is there overlap with the work @andrewbranch is doing in this PR?

https://github.com/TypeStrong/ts-loader/pull/817

He's been busy beavering away for the last week or so and it seems that there might be commonality here.  @andrewbranch can you comment on the work @JonWallsten has done.  

I'm wondering if it make sense for you guys to collaborate on this rather than having related PRs merged separately.

JonWallsten:
@johnnyreilly I was actually checking out his bransch earlier today. I don't think they collide at all. He should be able to merge this without conflicts. My changes are minor as you can se. I only allowed the regex to include .d.ts.map as well and made sure that the provideDeclarationFilesToWebpack function handles multiple output files (d.ts and d.ts.map). We should be able to merge this before his.

src/instances.ts
```diff
@@ -205,11 +205,11 @@ function successfulTypeScriptInstance(
   let normalizedFilePath: string;
   try {
     const filesToLoad = loaderOptions.onlyCompileBundledFiles
-      ? configParseResult.fileNames.filter(fileName =>
+      ? configParseResult.fileNames.filter((fileName: string) =>
```
johnnyreilly: is the type annotation necessary?  Presumably the type can be inferred?

src/instances.ts
```diff
@@ -205,11 +205,11 @@ function successfulTypeScriptInstance(
...
           dtsDtsxRegex.test(fileName)
         )
       : configParseResult.fileNames;
-    filesToLoad.forEach(filePath => {
+    filesToLoad.forEach((filePath: string) => {
```
johnnyreilly: is the type annotation necessary? Presumably the type can be inferred?

src/instances.ts
JonWallsten: I got an error in VS code, that's why I fixed it. But now when I check the tsconfig I can't find "noImplicitAny", so I'm not actually sure why I got that warning. VSCode has been a little shaky when it comes to errors since last release though. I don't have a problem removing them.

src/instances.ts
JonWallsten: Fixed

src/instances.ts
JonWallsten: Fixed

johnnyreilly:
> But please correct me if I'm wrong. I don't feel like I have the full picture of TS 3.0 yet.

Me neither 😄

src/after-compile.ts
```diff
@@ -246,18 +246,20 @@ function provideDeclarationFilesToWebpack(
...
-      const assetPath = path.relative(
-        compilation.compiler.context,
-        declarationFile.name
-      );
-      compilation.assets[assetPath] = {
-        source: () => declarationFile.text,
-        size: () => declarationFile.text.length
-      };
+    const declarationFiles: typescript.OutputFile[] = outputFiles
+      .filter((outputFile: typescript.OutputFile) => outputFile.name.match(constants.dtsDtsxRegex));
```
johnnyreilly: can `(outputFile: typescript.OutputFile)` just be `outputFile`?

src/after-compile.ts
```diff
@@ -246,18 +246,20 @@ function provideDeclarationFilesToWebpack(
...
+
+    if (Array.isArray(declarationFiles)) {
+      declarationFiles.forEach(file => {
```
johnnyreilly: can `file` be `declarationFile` please?

src/after-compile.ts
```diff
@@ -246,18 +246,20 @@ function provideDeclarationFilesToWebpack(
...
-        declarationFile.name
-      );
-      compilation.assets[assetPath] = {
-        source: () => declarationFile.text,
-        size: () => declarationFile.text.length
-      };
+    const declarationFiles: typescript.OutputFile[] = outputFiles
+      .filter((outputFile: typescript.OutputFile) => outputFile.name.match(constants.dtsDtsxRegex));
+
+    if (Array.isArray(declarationFiles)) {
```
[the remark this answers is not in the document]
JonWallsten: Yes, exactly! Due to the regex change, getEmitOutput will return both d.ts and d.ts.map when declarationMap is true. It felt logical to handle declaration & map files together since they are linked together inside the d.ts file, and besides the map file can't live without the declaration file. When it comes to performance loss for using a forEach instead of a pop() when it's only one should not a problem since we're not using a million and millions of files (i hope).

src/constants.ts
```diff
@@ -12,7 +12,7 @@ export const ScriptTargetES2015 = 2;
 export const ModuleKindCommonJs = 1;
 
 export const tsTsxRegex = /\.ts(x?)$/i;
-export const dtsDtsxRegex = /\.d\.ts(x?)$/i;
+export const dtsDtsxRegex = /\.d\.ts(x?)(\.map)?$/i;
```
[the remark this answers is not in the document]
JonWallsten: Absolutely! I'll fix!

src/after-compile.ts
```diff
@@ -246,18 +246,22 @@ function provideDeclarationFilesToWebpack(
...
-      compilation.assets[assetPath] = {
-        source: () => declarationFile.text,
-        size: () => declarationFile.text.length
-      };
+    const declarationFiles: typescript.OutputFile[] = outputFiles.filter(
+      (outputFile: typescript.OutputFile) =>
+        outputFile.name.match(constants.dtsDtsxOrDtsDtsxMapRegex)
+    );
+
+    if (Array.isArray(declarationFiles)) {
```
johnnyreilly: Is `Array.isArray` the right thing to do here? Presumably this will always be true as the result of the `filter` will always be an `Array`?

src/after-compile.ts
JonWallsten: Yeah, you are correct. I reinstalled VS Code, and it works now. I'm removing these.

src/after-compile.ts
JonWallsten: Sure! Fixed!

src/after-compile.ts
JonWallsten: Yeah, that is correct. But forEach will handle that. I'm only doing the array check to make sure forEach won't throw and error.

JonWallsten: Filter will always return an array. So I removed the check. It's safe to assume that we will always have an array, but it might be empty.

johnnyreilly:
Looks good.  I'll plan to merge and ship this shortly; thanks for your work!

JonWallsten:
Perfect! Pleasure all mine. I really want to feature. I'm glad it was easy to implement. Good work of the code base.

Issue #190, opened 2016-04-30
ts-loader generating definition file in wrong directory


jombie:
I am creating a new issue here as discussed in [typescript Issue] (https://github.com/Microsoft/TypeScript/issues/7958#issuecomment-215914134)
I am using typescript version 1.9.0-dev.20160412 and ts-loader version 0.8.2 as below
My tsconfig.json is as below:
```sh
{
  "compileOnSave": false,
  "buildOnSave": false,
  "compilerOptions": {
    "module": "commonjs",
    "target": "es6",
    "moduleResolution": "node",
    "preserveConstEnums": true,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "noImplicitAny": false,
    "noEmitOnError": false,
    "forceConsistentCasingInFileNames": true,
    "removeComments": false,
    "sourceMap": true,
    "jsx": "react",
    "outDir": "./lib",
    "declaration": true
  },
  "exclude": [
    "node_modules",
    "dist",
    "typings/browser",
    "typings/browser.d.ts"
  ]
}
```
My project folder structure is as below:
```sh
work/
`-- projects
    `-- projectx
        |-- dist
        |-- src
        `-- tsconfig.json
```
I am trying to compile and generate .d.ts files inside the 'lib' dir but this creates the .d.ts file inside folder with parent attached to lib i.e
```sh
work/
`-- projects
    `-- projectx
        |-- src
        |-- dist
        |   `-- work
        |       `-- projects
        |           `-- projectx
        |               `-- dist
        |                   `-- lib
        |                       `-- src
        |                           `-- index.d.ts
        `-- tsconfig.json
```

I was expecting it to be as :
```sh
work/
`-- projects
    `-- projectx
        |-- dist
        |   `-- lib
        |       `-- src
        |           `-- index.d.ts
        |-- src
        `-- tsconfig.json 
```
**dist** dir above is given in the webpack.config.js as  below:
```sh
output: {
    path: path.join(__dirname, '/dist'),
    filename: outputFile,
    library: libraryName,
    libraryTarget: 'umd',
    umdNamedDefine: true
  }
```
Sample project with the above mentioned error  [Sample Project](https://github.com/jombie/projectx).
It works as expected with the version 0.8.0 of ts-loader

leftstick:
Same issue

chrisnicola:
Actually even with 0.8.0 it outputs far more of the directory structure than I would expect. Instead of 

```
dist/
[...]

PR #822, opened 2018-08-17
Fixed issue with incorrect output path for declaration files

JonWallsten:
In regards of: https://github.com/TypeStrong/ts-loader/issues/190

I'll try to explain the fix:
Below is the code where we prepare the relative path for the d.ts files so we can add them to Webpack's compilation assets. 
```
const assetPath = path.relative(
        compilation.compiler.context,
        declarationFile.name
      );
compilation.assets[assetPath] = {
        source: () => declarationFile.text,
        size: () => declarationFile.text.length
      };
```
**Example**
Project root: /home/user/jon/projects/a/
[webpack.main]: index.ts
[webpack.output.path]: /home/user/jon/projects/a/dist/
[compilerOptions.declarationDir]: ./dist/typings/

[Scope variables]
[compilation.compiler.context]: /home/user/jon/projects/a/
[compilation.compiler.outputPath]: /home/user/jon/projects/a/dist/

[Other]
All relative assets in [compilation.assets] will have [compilation.compiler.outputPath] as base it seems.

**Results**
What will happen here is that `path.relative` will return a relative path with **compiler.context** (project root) as a base. So when TypeScript compiler outputs **/home/user/jon/projects/a/dist/index.d.ts** (declarationDir is relative to the project root), `path.relative` will strip the project root and we will end up with **dist/index.d.ts**. It looks correct, right? Well, if we would store the compiled files relative to the project root it would. The issue is that compilation.assets seems to equal webpack.output.path.
So we'll end up with this:
```
/home/user/jon/projects/a/dist/index.js
/home/user/jon/projects/a/dist/dist/index.d.ts
```
By changing `compilation.compiler.context` -> `compilation.compiler.outputPath` we will get a path relative to the output folder  instead.

So if we have  the same output from the TypeScript compiler: **/home/user/jon/projects/a/dist/index.d.ts** and we strip away **/home/user/jon/projects/a/dist/** instead, we end up with **index.d.ts**. The result would be:
```
/home/user/jon/projects/a/dist/index.js
/home/user/jon/projects/a/dist/index.d.ts
```

Note that this might be a breaking change since people probably has workarounds or have adapted to this way. Myself included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 5.0.0
+
+* [feat: Fixed issue with incorrect output path for declaration files](https://github.com/TypeStrong/ts-loader/pull/822) - thanks @JonWallsten! **BREAKING CHANGE**
+
 ## 4.5.0
 
 * [feat: Added support for TypeScript declaration map](https://github.com/TypeStrong/ts-loader/pull/821) - thanks @JonWallsten!
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "4.5.0",
+  "version": "5.0.0",
   "description": "TypeScript loader for webpack",
   "main": "index.js",
   "types": "dist/types/index.d.ts",
--- a/src/after-compile.ts
+++ b/src/after-compile.ts
@@ -251,7 +251,7 @@ function provideDeclarationFilesToWebpack(
 
     declarationFiles.forEach(declarationFile => {
       const assetPath = path.relative(
-        compilation.compiler.context,
+        compilation.compiler.outputPath,
         declarationFile.name
       );
       compilation.assets[assetPath] = {
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -149,6 +149,7 @@ export interface WebpackCompilation {
 export interface WebpackCompiler {
   isChild(): boolean;
   context: string; // a guess
+  outputPath: string;
   watchFileSystem: WebpackNodeWatchFileSystem;
   /** key is filepath and value is Date as a number */
   fileTimestamps: Map<string, number>;
--- a/test/comparison-tests/declarationOutput/expectedOutput-3.0/output.txt
+++ b/test/comparison-tests/declarationOutput/expectedOutput-3.0/output.txt
@@ -1,7 +1,7 @@
-                Asset       Size  Chunks             Chunk Names
-            bundle.js   5.13 KiB    main  [emitted]  main
-    ../types/app.d.ts  110 bytes          [emitted]  
-../types/sub/dep.d.ts   63 bytes          [emitted]  
+       Asset       Size  Chunks             Chunk Names
+   bundle.js   5.13 KiB    main  [emitted]  main
+    app.d.ts  110 bytes          [emitted]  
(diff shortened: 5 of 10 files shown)


johnnyreilly:
I really appreciate the detailed PR - that's very helpful indeed!

> Note that this might be a breaking change since people probably has workarounds or have adapted to this way. Myself included.

Could you provide some details around what workarounds people might be making use of?

JonWallsten:
I just mean that people have adapted to use a path relative to the output folder instead of the root.
My project for example:
The packages where I still use ATL have './dist/typings' 
The packages where I use TS-loader have just 'typings'
Now they will end up in the same output folder. The problem with the ts-loader however is that when I run tsc manually, for example when I compile the files for testing, the output folder will be /typings/ instead of /dist/typings/. And as you know you can't unfortunately have different tsconfig. So it's hard to work around.

johnnyreilly:
Cool. Essentially I'm trying to evaluate how "breaking" this change would be.  If it's not really that break-y then I'd consider treating this as a fix version-wise.  Alternatively we'd have to do a major version bump which I'm less keen on, but if it's appropriate then that's what it should be.  Do you have a view?

JonWallsten:
Yeah, a major seems unnecessary for this small fix. But still I might require action from the user. I'm not sure what the conventional thing to do is.
I'm not in a hurry with this fix, so take your time and let me know. Whatever works for me!

johnnyreilly:
Thanks chap. I'm not going to be online much for the next couple of weeks; I'll probably look to ship this with https://github.com/TypeStrong/ts-loader/pull/817 when the time comes.

JonWallsten:
Sound good! Just ping me when the time comes!

JonWallsten:
@gluons I would have to say that it's intended (but maybe unwanted in some cases?). It's TypeScripts default behavior to do this. If you don't add outDir or declarationDir it might be intended, and you could get annoyed by a warning, since it's in the context of TypeScript isn't wrong. What does @johnnyreilly think?

johnnyreilly:
> It's TypeScripts default behavior to do this.

Then that's what ts-loader should do 😁

As a general rule of thumb, ts-loader aims to be a drop in replacement for `tsc` behaviour wise; only deviating if there's specific cause. I don't see that in this case so I think it's just fine.

BTW I'm hoping to get this merged and shipped either next week or the week after.

johnnyreilly:
Hey @JonWallsten,

I'm just heading back from holidays.  I've been thinking; although this is a minor change, you're right that it's technically a breaking change.  As much as I don't want to bump the version number; ts-loader seeks to be a good semver citizen.  So it's time to bite the bullet.

Please could you:

1. Update the version in the `package.json` to `5.0.0`
2. Add a corresponding entry to the [`CHANGELOG.md`](https://github.com/TypeStrong/ts-loader/blob/master/CHANGELOG.md)

Then I'll close my eyes, weep a bit and push out the major version.  Cheers!

JonWallsten:
@johnnyreilly Sure. I'll fix it before nightfall!

johnnyreilly:
Thanks for your work @JonWallsten!

Released with https://github.com/TypeStrong/ts-loader/releases/tag/v5.0.0

Jogai:
> I mean, look at Chrome 68!

You mean 69?

PR #829, opened 2018-08-29
Added cache for some FS operations while compile

### LINKED ISSUE #825: Add cache for some FS operations while compile
Hi there.

TypeScript makes a lot of identical FS operations such as:

- checking that some file/directory exists (e.g. for the identification an extension of a file it can [make up to 3 `fileExists` calls](https://github.com/Microsoft/TypeScript/blob/ac0d5da6313a6ee8965bf8dfff57aa00053d4532/src/compiler/moduleNameResolver.ts#L947))
- resolving original file path

To solve an issue described in https://github.com/Microsoft/TypeScript/issues/22576 in our project we have developed a small tool, which compiles all TypeScript files one-by-one as entry point and checks errors.
Now the project has grown to 1000+ TypeScript files and the test takes more than 20-30 minutes (sometimes it even fails because of timeout after 1 hour).
After investigating the problem we found out that the reason is a lot of FS operations made by TypeScript compiler.

_(I believe that we can assume that compilation process is "atomic" operation and we can just add cache for some "heavy" syscall-operations like `fileExists`)_

After adding cache for `fileExists`, `realpath`, `directoryExists` and `getSourceFile` (`getSourceFile` is specific for our tool) we get 20-30x decreasing of working time (to 1 minute).

I hope that this approach can be used in `ts-loader` too to speedup compilation time.

I roughly added some caches to `fileExists`, `realpath` and `directoryExists` functions and here are miss/hit values (for our project) in cold-compilation:

- `fileExists`: miss=55847, hit=630450
- `directoryExists`: miss=51, hit=1914
- `realpath`: miss=39, hit=466

The compilation time is decreased from 850s to 330s.

timocov:
Fixes #825.

Open questions:

- Should it be under option?
- If so, how this option should be named? Should it be something like "experimental" flag?
- ~Do we need to add logs? Is so, what kind of logs (clearing caches or something else)?~
- ~Should we add the same caches to `makeWatchHost` too?~
- I'm not sure how to write tests for this - possible somebody has thoughts about it.


johnnyreilly:
Thanks for the PR! I'm away from home right now with only my phone and dodgy connectivity and so will struggle to review this until in front of a proper screen. 

To your questions:

> Should it be under option? If so, how this option should be named? Should it be something like "experimental" flag?

Yes please; I like the idea of this going out under a flag called something like `experimentalFileCaching`. If it works well and no problems are reported we can look to promote this to the default behaviour in future.

> Do we need to add logs? Is so, what kind of logs (clearing caches or something else)?

Don't feel strongly. I'd say probably not for now.

> Should we add the same caches to makeWatchHost too?

I'm not sure; I think not for now.

> I'm not sure how to write tests for this - possible somebody has thoughts about it.

The encouraging thing is that tests pass before you've put it behind an option. But yeah - need to think about this

johnnyreilly:
Also, please could you experiment with this on your own projects and document the performance improvements you're seeing here please? 

I'd like the benefits to be easily discoverable for people who want learn about this. Thanks!

timocov:
> Yes please; I like the idea of this going out under a flag called something like experimentalFileCaching. If it works well and no problems are reported we can look to promote this to the default behaviour in future.

Ok, I'll do it.

> Also, please could you experiment with this on your own projects and document the performance improvements you're seeing here please?

Sure. But currently I am able to check this fix against 3.5.0 version only (I have fix for this version too - see https://github.com/timocov/ts-loader/tree/fix825_v3) because we are in progress to upgrade to webpack 4 for now. But I'll try to check with v4 too.

johnnyreilly:
> Sure. But currently I am able to check this fix against 3.5.0 version only 

I'm guessing you know that ts-loader has supported only webpack 4+ since version 4?

timocov:
> I'm guessing you know that ts-loader has supported only webpack 4+ since version 4?

Yep 🙂

timocov:
Here are cold-compilation results for our project.

Env:

- PC: AMD FX-4300 Quad-Core 4x3.8GHz, 16GB RAM, HDD, Windows 10
- NodeJS version: `10.3.0`
- TypeScript version: `3.0.1`
- Approx. modules count: 3000-4700 (from webpack stats)

#### `webpack@3.10.0` runs:

| `ts-loader` version | compilation time (avg of three runs) |
|---|---|
| `2.3.3` | 554s (546s, 553s, 564s) |
| `3.5.0` | 828s (850s, 810s, 823s) |
| `3.5.0` (with cache) | 303s (296s, 312s, 302s) |

#### `webpack@4.16.0` runs:

| `ts-loader` version | compilation time (avg of two runs) |
|---|---|
| `4.5.0` | 419s (413s, 425s) |
| `4.5.0` (with cache) | 315s (307s, 324s) |

I think that we cannot compare results between webpack 3 and webpack 4 because there are some changes in our project for webpack 4 and I cannot guaratnee that the build has exactly the same input (but inside one webpack version it should be the same).

timocov:
> I'm guessing you know that ts-loader has supported only webpack 4+ since version 4?

If this PR will be merged, is there any chance to make the similar PR for 3 version (I'm ready to provide necessary changes for 3.5 version)?

johnnyreilly:
> If this PR will be merged, is there any chance to make the similar PR for 3 version (I'm ready to provide necessary changes for 3.5 version)?

You mean a PR for this branch:  https://github.com/TypeStrong/ts-loader/tree/ts-loader-3?files=1

Which I'd then publish as 3.6?

Honestly I'm not keen since I've not been supporting webpack 3 since 4 shipped and I only have a certain amount of time I can devote to open source. This would take away part of that time and given people are moving away from webpack 3 it doesn't feel like time well spent.

Will that be super painful for you?

timocov:
> given people are moving away from webpack 3

Hm, good point 👍 

> Will that be super painful for you?

I guess in this case we can just publish it to local npm until we'll upgrade to webpack 4.

timocov:
@johnnyreilly do I need to do something else in the PR? It is possible that you have any confused/ambiguously thoughts about it (I guess except tests)?

johnnyreilly:
Sorry - I hadn't spotted that you'd added further commits.  I'm back from my travels now; I'll try and take a look soon.  Please bear with me!

src/instances.ts
```diff
@@ -266,17 +272,25 @@ function successfulTypeScriptInstance(
       .getProgram()
       .getProgram();
   } else {
-    const servicesHost = makeServicesHost(scriptRegex, log, loader, instance);
+    const cachedServicesHost = makeServicesHost(
```
johnnyreilly: Could this be `servicesHost` instead please? It may or may not be cached depending upon the flag.

src/instances.ts
```diff
@@ -266,17 +272,25 @@ function successfulTypeScriptInstance(
...
       compiler.createDocumentRegistry()
     );
-  }
 
-  if (!loader._compiler.hooks) {
-    throw new Error(
-      "You may be using an old version of webpack; please check you're using at least version 4"
-    );
+    if (cachedServicesHost.clearCache !== null) {
+      loader._compiler.hooks.watchRun.tap(
```
johnnyreilly: I just wanted to check; is `watchRun` definitely `sync` not `async`?

src/servicesHost.ts
```diff
@@ -509,3 +520,45 @@ function populateDependencyGraphs(
...
+    const cache = createCache(servicesHost.realpath);
+    servicesHost.realpath = cache.cached;
+    clearCacheFuncs.push(cache.clear);
+  }
+
+  return () => clearCacheFuncs.forEach(clear => clear());
+}
+
+function createCache<TOut>(func: (arg: string) => TOut) {
+  let cache: Record<string, TOut | undefined> = {};
```
johnnyreilly: I wonder if it might be slightly more performant to use a `Map` instead?  What do you think? One of the big webpack 4 changes was it moving to using `Map`s.

src/instances.ts
timocov: Yeah, no problem.

Do I need to rename `serivcesHost` property in `CachedServicesHost` interface too? Also, it is possible, do I need to rename `CachedServicesHost` interface? Change it to `type` instead of `interface`?

src/instances.ts
timocov: It seems that `watchRun` is Tappable, which supports both `async` and `sync` methods (`tap` and `tapAsync`).

https://webpack.js.org/api/compiler-hooks/#watchrun
https://github.com/webpack/tapable/blob/04e52118691e0a6c77a42d15e99dc4bd2bfca07c/lib/AsyncSeriesHook.js#L21

src/servicesHost.ts
timocov: Oh, you're right! _I need to start to forget to use simple object as map to be compatible on all "platforms" 🙂_

src/instances.ts
johnnyreilly: Is there any advantage to using `tapAsync` instead of `tap`?  I've an idea it doesn't really matter but I thought I'd check.

timocov: I believe that it depends on what are you doing in your handler. In this case we need to use sync because we need to just clear cache without any async operation. But if you find something interesting about it - please let me know 🙂

src/servicesHost.ts
```diff
@@ -14,15 +14,21 @@ import {
   Webpack
 } from './interfaces';
 
+export interface CachedServicesHost {
```
johnnyreilly: Perhaps call this `ServiceHostWhichMayBeCacheable`?  It's super verbose but it's also very descriptive. Should make it clear to anyone hacking on the code.  I don't mind whether it's a `type` or an `interface`.

src/instances.ts
johnnyreilly: > Do I need to rename serivcesHost property in CachedServicesHost interface too? 

No I think it's fine.

src/servicesHost.ts
```diff
@@ -509,3 +520,45 @@ function populateDependencyGraphs(
     ] = true;
   });
 }
+
+function addCache(servicesHost: typescript.ModuleResolutionHost) {
+  const clearCacheFuncs: (() => void)[] = [];
```
johnnyreilly: could this become a `type` please?  Perhaps `type Action = () => void;`?

timocov: Do I need to make return type of `createCache` as `type`/`interface` too?

src/servicesHost.ts
timocov: Ok, what's about `servicesHost` property? Leave without changes?

src/servicesHost.ts
```diff
@@ -509,3 +520,45 @@ function populateDependencyGraphs(
...
+
+  if (servicesHost.realpath !== undefined) {
+    const cache = createCache(servicesHost.realpath);
+    servicesHost.realpath = cache.cached;
+    clearCacheFuncs.push(cache.clear);
+  }
+
+  return () => clearCacheFuncs.forEach(clear => clear());
+}
+
```
johnnyreilly: I wonder if we could reduce some duplication here?

```
['fileExists', 'directoryExists', 'realpath'].forEach(propertyToCache => {
      if (servicesHost[propertyToCache] !== undefined) {
        const cache = createCache(servicesHost[propertyToCache]);
        servicesHost[propertyToCache] = cache.cached;
        clearCacheFuncs.push(cache.clear);
      }
});
```

src/servicesHost.ts
johnnyreilly: No it's fine I think

src/servicesHost.ts
johnnyreilly: Yeah it's possible you might need to do this:

```
['fileExists', 'directoryExists', 'realpath'].forEach((propertyToCache: 'fileExists' | 'directoryExists' | 'realpath') => {
      if (servicesHost[propertyToCache] !== undefined) {
        const cache = createCache(servicesHost[propertyToCache]);
        servicesHost[propertyToCache] = cache.cached;
        clearCacheFuncs.push(cache.clear);
      }
});
```

johnnyreilly: I bet you could equally use `Proxy` for this functionality too: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy#Basic_example

Never actually used a `Proxy` in JS yet 😄

johnnyreilly: Ah node 6 minimum... https://node.green/#ES2015-built-ins-Proxy

timocov: It seems that it wouldn't work because TypeScript infer type `(((arg: string) => string) | (arg: string) => boolean))` to type `(arg: string) => string | boolean`, but they are not the same types.

timocov: Even if I change to this:

```ts
  [
    'fileExists' as 'fileExists',
    'directoryExists' as 'directoryExists',
    'realpath' as 'realpath',
  ].forEach(propertyToCache => {
    const origFn = servicesHost[propertyToCache];
    if (origFn !== undefined) {
      const cache = createCache(origFn);
      servicesHost[propertyToCache] = cache.cached;
      clearCacheFuncs.push(cache.clear);
    }
  });
```

and `createCache` to `function createCache<TFn extends (arg: string) => TOut, TOut = ReturnType<TFn>>(func: TFn) {` TypeScript fails with error ` Type '(arg: string) => string | boolean' is not assignable to type '((fileName: string) => boolean) | ((path: string) => string) | undefined'.`.

I can hack it with casting via `as`, but I'm not sure that it is good idea 🙂

timocov: It seems the following code compiles well, but it might be a little bit tricky:

```ts
  [
    'fileExists' as 'fileExists',
    'directoryExists' as 'directoryExists',
    'realpath' as 'realpath',
  ].forEach(propertyToCache => {
    const origFn = servicesHost[propertyToCache];
    if (origFn !== undefined) {
      const cache = createCache<ReturnType<typeof origFn>>(origFn);
      servicesHost[propertyToCache] = cache.cached as typescript.ModuleResolutionHost[typeof propertyToCache];
      clearCacheFuncs.push(cache.clear);
    }
  });
```

What do you think?

timocov: I can avoid specifying `<ReturnType<typeof origFn>>`  by changing `createCache`'s declaration as in https://github.com/TypeStrong/ts-loader/pull/829#discussion_r214755099 also.

timocov: To add additional compile-time checks of property names we can do the following:

```ts
  type CachedProperty = Extract<
    keyof typescript.ModuleResolutionHost,
    'fileExists' | 'directoryExists' | 'realpath'
  >;
  const cachedProperties: CachedProperty[] = [
    'fileExists',
    'directoryExists',
    'realpath'
  ];

  cachedProperties.forEach((propertyToCache: CachedProperty) => {
    const origFn = servicesHost[propertyToCache];
    if (origFn !== undefined) {
      const cache = createCache<ReturnType<typeof origFn>>(origFn);
      servicesHost[
        propertyToCache
      ] = cache.cached as typescript.ModuleResolutionHost[CachedProperty];
      clearCacheFuncs.push(cache.clear);
    }
  });
```

timocov:
@johnnyreilly can you please take a look?

johnnyreilly:
Thanks @timocov - I will!

johnnyreilly:
Really sorry About the delay @timocov - very very busy right now. I haven't forgotten about this - just not had time

src/servicesHost.ts
```diff
@@ -97,13 +108,15 @@ export function makeServicesHost(
...
     useCaseSensitiveFileNames: () => compiler.sys.useCaseSensitiveFileNames,
 
+    realpath: moduleResolutionHost.realpath,
+
     // The following three methods are necessary for @types resolution from TS 2.4.1 onwards see: https://github.com/Microsoft/TypeScript/issues/16772
-    fileExists: compiler.sys.fileExists,
-    readFile: compiler.sys.readFile,
+    fileExists: moduleResolutionHost.fileExists,
+    readFile: moduleResolutionHost.readFile,
     readDirectory: compiler.sys.readDirectory,
```
johnnyreilly: was there a reason that `readDirectory` wasn't included in the caching functionality?

src/servicesHost.ts
```diff
@@ -97,13 +108,15 @@ export function makeServicesHost(
     /**
      * For @types expansion, these two functions are needed.
      */
-    directoryExists: compiler.sys.directoryExists,
+    directoryExists: moduleResolutionHost.directoryExists,
 
     useCaseSensitiveFileNames: () => compiler.sys.useCaseSensitiveFileNames,
 
+    realpath: moduleResolutionHost.realpath,
```
johnnyreilly: We didn't have `realpath` before I think; I'm just curious; what is it used for?

src/servicesHost.ts
johnnyreilly: If you could check that'd be awesome. I'm currently working on tweaking the test packs so they can be made to work against a variety of `experimental...` flags - the idea being that while the cache functionality lies behind a flag we can still make sure it is tested by the existing test pack. Watch this space...

https://github.com/TypeStrong/ts-loader/pull/834

src/servicesHost.ts
johnnyreilly: Ah - we've had since we added symlink support here: https://github.com/TypeStrong/ts-loader/pull/774/files

We just didn't supply it to the `LanguageServiceHost` until now though. I'm kind of surprised we didn't get a compilation error previously 🤔

johnnyreilly:
Hey @timocov,

I've been beavering away on the test pack and I think I've now got the comparison test pack set up so it does 2 passes; the existing pass and then again but with the `experimentalFileCaching` flag set.  I'm just waiting for CI to pass.  Assuming it does I think we're nearly ready to merge. 😄

src/servicesHost.ts
timocov: > If you could check that'd be awesome

Unfortunately for now I'm away from work and cannot check it. I'll do it in next couple of days and will provide here results.

src/servicesHost.ts
timocov: Hm, interesting. Perhaps https://github.com/Microsoft/TypeScript/pull/12020#issuecomment-396287996 (and the whole PR) is related to this.

timocov:
Wow, thank you @johnnyreilly!

johnnyreilly:
My pleasure - you've been an awesome contributor! ❤️ 🌻

src/servicesHost.ts
timocov: @johnnyreilly I just checked `readDirectory` - it seems that this function is never called while compilation (at least in our case).

PR #835, opened 2018-09-10
Pass raw path to transpileModule so that customTransformers could access the raw path

Brooooooklyn:
If I want write some `path based logic` in my customTransformer,  it couldn't be done now. Because only the filename was passed to typescript.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 5.1.1
+* [fix(getTranspilationEmit): pass the raw path to transpileModule](https://github.com/TypeStrong/ts-loader/pull/835) - thanks @Brooooooklyn 
+
 ## 5.1.0
 
 * [feat: Added cache for some FS operations while compiling - `experimentalFileCaching`](https://github.com/TypeStrong/ts-loader/pull/829) - thanks @timocov!
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "5.1.0",
+  "version": "5.1.1",
   "description": "TypeScript loader for webpack",
   "main": "index.js",
   "types": "dist/types/index.d.ts",


johnnyreilly:
Could you provide an example of why this is useful / what possibilities this unblocks please? 

Also, why `transpileOnly` specific? (Unless I've misunderstood your PR it's not catering for non-transpile)

And, would the change your suggesting change the usage for other people already relying on custom transformers? Would this be a breaking change?

Brooooooklyn:
> Could you provide an example of why this is useful / what possibilities this unblocks please?

Yes, I want to write a typescript transformer plugin like https://github.com/emotion-js/emotion/tree/master/packages/babel-plugin-emotion#sourcemap, and I must inject `source-map` into output results. Before I do such things, I must access the rawPath of source files.

> Also, why transpileOnly specific?
 
in `non-transpile` situation, rawPath could get by typescript itself

> would the change you are suggesting change the usage for other people already relying on custom transformers? Would this be a breaking change?

Maybe. It would change the value of `node.getSourceFile().fileName` in  custom transformers. from `xx.ts` to `/Users/aaa/bbb/ccc/xx.ts`

johnnyreilly:
> Before I do such things, I must access the rawPath of source files.

Cool

> in non-transpile situation, rawPath could get by typescript itself

Sorry I don't follow your meaning here. Could you expand on this please?

Brooooooklyn:
let me explain it in this way.
if you write a custom transformer, and using it with `transpileOnly: true` and `transpileOnly : false` flags.
in `transpileOnly: true` situation, the value of `node.getSourceFile().fileName` is `xxx.ts`
in `transpileOnly: false` situation, the value of  `node.getSourceFile().fileName` is `/Users/aaa/bbb/ccc/xxx.ts`, this value is correct for me now, and it was provided by TypeScript (typescript.Program resolve this path and provide it to custom transformer)

johnnyreilly:
Ah I see!  So this will align the behaviour of `transpileOnly: true` and `transpileOnly: false`, is that right?

If that's the case I kind of consider this a bugfix and I think we should take it.

Brooooooklyn:
> So this will align the behaviour of transpileOnly: true and transpileOnly: false, is that right?
Yes, right

johnnyreilly:
Could you update the `package.json` to 5.1.1 and put an entry in the changelog please?

Brooooooklyn:
updated

Brooooooklyn:
@johnnyreilly

johnnyreilly:
Thanks - quite busy right now but I'll try to put a release out soon.

Brooooooklyn:
thank you😁

johnnyreilly:
Shipped with https://github.com/TypeStrong/ts-loader/releases/tag/v5.1.1

PR #841, opened 2018-09-25
Lists typescript as a peer dependency

### LINKED ISSUE #839: TypeScript should be a peerdependency
### Expected Behaviour

`ts-loader` shouldn't rely on undefined behaviors

### Actual Behaviour

`ts-loader` currently relies on the hoisting in order to have access to the `typescript` package

The fix is pretty simple: `typescript` just has to be listed as a `peerDependencies` of `ts-loader` (rather than just a `devDependencies`). This will guarantee that it will then get the exact same instance than its parent. It can be listed as both if you want to also install it by default on dev environments.

It causes issues with package trees that do not support hoisting, such as Yarn Plug'n'Play.

Because Plug'n'Play has a fallback mechanism, hitting this case is a bit more difficult to reproduce - but basically, if you have a workspace that depends on both `ts-loader` and `typescript`, it won't work (because `ts-loader` won't be allowed to access `typescript` as it's not one of its dependencies, and because the fallback won't kick in since `typescript` will be in a workspace rather than the top-level module).

arcanis:
> `ts-loader` currently relies on the hoisting in order to have access to the `typescript` package

> It causes issues with package trees that do not support hoisting, such as Yarn Plug'n'Play.

Should fix #839 😃

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 5.2.1
+
+* [feat: Lists typescript as a peer dependency](https://github.com/TypeStrong/ts-loader/pull/841) - thanks @arcanis!
+
 ## 5.2.0
 
 * [feat: Initial support for project references - `projectReferences`](https://github.com/TypeStrong/ts-loader/pull/817) - thanks @andrewbranch!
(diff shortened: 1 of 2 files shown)


johnnyreilly:
Thanks! I'll try and get this shipped soon.

johnnyreilly:
https://github.com/TypeStrong/ts-loader/releases/tag/v5.2.1

arcanis:
Thanks!

PR #853, opened 2018-10-08
fix: filter import errors and close #653

### LINKED ISSUE #653: ES2015 module issue with transpileOnly + decorators + interfaces
Hi there,

We have a very large Angular 2 codebase that we're attempting to move over to `ts-loader`; we're getting massive (4-5x) speed boosts compared to our existing awesome-typescript-loader setup, but sadly we're getting compilation issues when we turn on `transpileOnly`.

It only occurs when:

- Building ES2015 modules AND
- Using a class with a decorator on it AND
- The class uses an imported interface AND
- The imported interface comes from a module that _also_ exports something concrete (e.g. a class or a const)

(webpack complains that it can't find the interface with a stack trace suggesting that it's processing harmony imports.)

I've created a repo with a standalone repro case in it: it's just tiny 2 files both of which are under 10 lines of code:

https://github.com/RoystonS/ts-loader-issue

(Is this related to what's discussed in issue #400?)

Thanks,
 Royston.

rolandjitsu:
Fix for https://github.com/TypeStrong/ts-loader/issues/653 (following https://github.com/TypeStrong/ts-loader/issues/653#issuecomment-400911783). Note that I did not add any tests as I was unsure where to add the tests.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -20,6 +20,7 @@ import {
   formatErrors,
   getAndCacheOutputJSFileName,
   getAndCacheProjectReference,
+  noImportError,
   validateSourceMapOncePerProject
 } from './utils';
 
@@ -475,7 +476,7 @@ function getTranspilationEmit(
       loaderContext.context
     );
 
-    loaderContext._module.errors.push(...errors);
+    loaderContext._module.errors.push(...errors.filter(noImportError));
   }
 
   return { outputText, sourceMapText };
--- a/src/instances.ts
+++ b/src/instances.ts
@@ -22,7 +22,8 @@ import {
   ensureProgram,
   formatErrors,
   isUsingProjectReferences,
-  makeError
+  makeError,
+  noImportError
 } from './utils';
 import { makeWatchRun } from './watch-run';
 
@@ -111,7 +112,7 @@ function successfulTypeScriptInstance(
       loader.context
     );
 
-    loader._module.errors.push(...errors);
+    loader._module.errors.push(...errors.filter(noImportError));
 
     return {
       error: makeError(
@@ -176,7 +177,7 @@ function successfulTypeScriptInstance(
         loader.context
       );
 
-      loader._module.errors.push(...errors);
+      loader._module.errors.push(...errors.filter(noImportError));
     }
 
     instances[loaderOptions.instance] = {
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -268,6 +268,13 @@ export function isUsingProjectReferences(instance: TSInstance) {
   return false;
 }
 
+export function noImportError(error: WebpackError) {
+  const { message } = error;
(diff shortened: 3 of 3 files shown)


johnnyreilly:
Could you share a minimal repo that demonstrates this issue? I can use that as a basis for an execution test.

rolandjitsu:
Let me setup that 😄

rolandjitsu:
@johnnyreilly I just setup [tsl-import](https://github.com/rolandjitsu/tsl-import), but I also just found that you can already disable these warnings with:
```js
stats: {
  warningsFilter: /export .* was not found in/
}
```

But if you switch to `react-scripts-ts` branch, you can still see the same kind of error, but different message. I'm trying to track it down as I'm not sure if this is coming from ts-loader.

johnnyreilly:
> But if you switch to react-scripts-ts branch, you can still see the same kind of error, but different message. I'm trying to track it down as I'm not sure if this is coming from ts-loader.

It'll likely be coming from: https://github.com/Realytics/fork-ts-checker-webpack-plugin

`react-scripts-ts` uses that for checking types and ts-loader for transpilation.

rolandjitsu:
That's the thing, I'm not so sure about that as I removed the plugin from the webpack config and I still see this error.

rolandjitsu:
In any case, I don't think this PR will solve the issue anyway, so maybe we should close it?

johnnyreilly:
> That's the thing, I'm not so sure about that as I removed the plugin from the webpack config and I still see this error.

react-scripts-ts is still webpack 3. I wonder if that's the reason?

johnnyreilly:
> In any case, I don't think this PR will solve the issue anyway, so maybe we should close it?

If you're sure.  I'm sorry you went to wasted effort though!

rolandjitsu:
> > That's the thing, I'm not so sure about that as I removed the plugin from the webpack config and I still see this error.
> 
> react-scripts-ts is still webpack 3. I wonder if that's the reason?

I will try with w4 and see how that works out.

rolandjitsu:
Ok, I tracked down the error. Seems to be happening at [webpack/lib/dependencies/HarmonyExportImportedSpecifierDependency.js](https://github.com/webpack/webpack/blob/master/lib/dependencies/HarmonyExportImportedSpecifierDependency.js).

rolandjitsu:
Could I ask what is different from what ts-loader generates as output with `transpileOnly: true` vs `transpileOnly: false`?

johnnyreilly:
Sure - there's a subtle difference in terms of produced code; but this only comes into play rarely; when `emitDecoratorMetadata: true` is in play and also I've an idea that imported `enum`s also have a nuance.  There may be more detail here:

https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API#transpiling-a-single-file

PR #858, opened 2018-10-15
Fix duplicated typo.

malash:
> Instead of using happypack you can use ts-loader with ts-loader with [thread-loader](https://github.com/webpack-contrib/thread-loader)

Double `ts-loader with` found.


johnnyreilly:
Thanks!

Issue #861, opened 2018-10-24
Expose `resolveModuleNames`


arcanis:
Hi!

Would it be possible to consider exposing the `resolveModuleNames` function (set in `servicesHost.ts`) through a configuration settings?

I'm currently experimenting integrating Yarn PnP with TypeScript through ts-loader, and my initial results are very promising. I think I could get something working just by going through this hook, which would unlock pretty interesting options!

johnnyreilly:
Go for it @arcanis! Experiment on!

arcanis:
Will aim to set up a PR in the next week or so 👍

johnnyreilly:
Cool - look forward to it. If tests cause you problems (many people bump on this) then stick up the PR and we can collaborate

johnnyreilly:
> It's slow because in the current implementation, names are resolved twice, and the results are not cached.

By the way, did you know about this option recently added to ts-loader which adds caching? I'm planning to turn it on by default (currently off) as I've had no reports of issues:

https://github.com/TypeStrong/ts-loader/blob/master/README.md#experimentalfilecaching-boolean-defaultfalse

thatsmydoing:
> By the way, did you know about this option recently added to ts-loader which adds caching? I'm planning to turn it on by default (currently off) as I've had no reports of issues:

Yeah, I've tried it but it didn't make much of an improvement for us. The caching happens too late in the process it seems. Module resolution seems to be a recursive operation and that seems to be what's taking up the time.

I took a look at the current PR and the changes are also happening at too low a level for the change I was thinking of making. I'll experiment a bit more with how the resolution could be sped up.

arcanis:
#862 is merged, issue fixed 👍

PR #862, opened 2018-10-26
Implements a custom resolveModuleName option

arcanis:
Solves #861 - This PR adds a new `resolveModuleName` option that can be used to help TypeScript figure out the location of the files on the disk. It mimics the TypeScript API.

I published a first implementation as part of [ts-pnp](https://github.com/arcanis/ts-pnp), and it can be tested using the [pnp-webpack-plugin](https://github.com/arcanis/pnp-webpack-plugin#ts-loader-integration) which provides an integration.

I'm not too sure how and where I should test this - any advice?


src/interfaces.ts
```diff
@@ -295,6 +295,21 @@ export interface ReverseDependencyGraph {
...
+  moduleName: string,
+  containingFile: string,
+  compilerOptions: typescript.CompilerOptions,
+  moduleResolutionHost: typescript.ModuleResolutionHost,
+) => typescript.ResolvedModuleWithFailedLookupLocations;
+
+export type CustomResolveModuleName = (
+  moduleName: string,
+  containingFile: string,
+  options: typescript.CompilerOptions,
```
johnnyreilly: Could this be called `compilerOptions` also please?

src/servicesHost.ts
```diff
@@ -496,12 +505,17 @@ function resolveModuleName(
     // tslint:disable-next-line:no-empty
   } catch (e) {}
 
-  const tsResolution = compiler.resolveModuleName(
-    moduleName,
-    containingFile,
-    compilerOptions,
-    moduleResolutionHost
-  );
+  const tsResolver = (moduleName: string, containingFile: string, compilerOptions: typescript.CompilerOptions, moduleResolutionHost: typescript.ModuleResolutionHost) =>
```
johnnyreilly: Given `tsResolver` is a pure function could we break this out in to a separate function please?  That saves creating a new function each time this runs.

johnnyreilly:
Thanks for this! Could you address the comments and fix the linting errors that are showing up in CI please?

johnnyreilly:
Regarding testing, it would be good to implement an execution test to cover this.  You can find out more about these here: https://github.com/TypeStrong/ts-loader/blob/master/test/execution-tests/README.md

A good example to clone as a basis for your test might be this one:

https://github.com/TypeStrong/ts-loader/tree/master/test/execution-tests/3.0.1_resolveJsonModule

src/servicesHost.ts
```diff
@@ -505,17 +514,9 @@ function resolveModuleName(
...
-      moduleName,
-      containingFile,
-      compilerOptions,
-      moduleResolutionHost
-    );
-
   const tsResolution = customResolveModuleName
-    ? customResolveModuleName(moduleName, containingFile, compilerOptions, moduleResolutionHost, tsResolver)
-    : tsResolver(moduleName, containingFile, compilerOptions, moduleResolutionHost);
+    ? customResolveModuleName(moduleName, containingFile, compilerOptions, moduleResolutionHost, applyTsResolver.bind(null, compiler))
```
johnnyreilly: I hadn't expected the `.bind(null, compiler)` would be necessary... I'm guessing it is?  What happens without it? Presumably some kind of `this` issue?

test/execution-tests/3.0.1_resolveModuleName/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+	"compilerOptions": {
+		"noEmitOnError": true,
+		"resolveJsonModule": true
```
johnnyreilly: can we remove the `"resolveJsonModule": true` please? It doesn't do any harm but it might lead someone who was reading the test to believe that it was necessary for `resolveModuleName` to work.  (Not least because of the similarity in naming 😄 )

test/execution-tests/3.0.1_resolveModuleName/test/app.tests.ts
```diff
@@ -0,0 +1,12 @@
+import * as app from '../src/app';
+
+// We don't actually care about the result of the following operations;
+// what we care about is typescript resolving json as modules
```
johnnyreilly: Could you amend the comment to reflect what the intention of the test is please?

src/servicesHost.ts
arcanis: No, we simply need the `compiler` variable to be able to call `compiler.resolveModuleName` (and `compiler` is obtained from `instance`, which isn't a global)

test/execution-tests/3.0.1_resolveModuleName/test/app.tests.ts
arcanis: I left the comment because it's actually the same intent - we don't care about what is being built, just about TS being able to make the resolution

arcanis: Oh, I see a "json" is in the comment - will change it 👍

test/execution-tests/3.0.1_resolveModuleName/tsconfig.json
arcanis: Sure 👍

johnnyreilly:
Good work on the changes @arcanis!  I've a couple more comments that I've added just now.  

Well done on the execution test.  Looks good!  Question though: it looks like it's only resolving types from `foo-pkg` / `barj-pkg`.  There doesn't seem to be any actual JS being resolved.   It might be good to include some JS as well so we know that as well as the compiler resolving types correctly we can see the actual code that we want to run coming through as well.  

Does that sound sensible?  Or have I misunderstood your intent?

johnnyreilly:
> @johnnyreilly I wasn't entirely sure if the generated code had to be tested as well, 

Yes please!

> since I wasn't sure whether ts-loader was entirely relying on TypeScript for the resolution or not (versus only using it for the type resolution, leaving the regular Webpack resolver setup the links between the packages). 

It's a good question. `ts-loader` uses webpack for actual file resolution:

https://github.com/TypeStrong/ts-loader/blob/master/src/resolver.ts

Funnily enough, it's on my radar to try out using TypeScript for file resolution too, but as yet I haven't.

src/servicesHost.ts
```diff
@@ -496,12 +520,22 @@ function resolveModuleName(
...
-    moduleResolutionHost
-  );
+  const tsResolution =
+    customResolveModuleName !== undefined
+      ? customResolveModuleName(
+          moduleName,
+          containingFile,
+          compilerOptions,
+          moduleResolutionHost,
+          applyTsResolver.bind(null, compiler)
```
johnnyreilly: Seeing a `.bind` my perf spider sense is tingling!  

https://stackoverflow.com/questions/42117911/lambda-functions-vs-bind-memory-and-performance

I wonder if we could switch this to a lamba instead?  It'll be slightly more verbose but hopefully should perform better...

test/execution-tests/3.0.1_resolveModuleName/test/app.tests.ts
```diff
@@ -0,0 +1,13 @@
+import * as app from '../src/app';
+
+// We don't actually care about the result of the following operations;
```
johnnyreilly: This is not true actually; we *do* care about the result of the following operations!  Can we amend the comment to make test intent clearer please?  We care both about type resolution and that the code executes...

johnnyreilly:
jdjijjujoijiojjj

jjjjjjames1111111111111111111111111111111111122222222222222222222222333333333333333333444444444444444444444444444444444344444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444444

johnnyreilly:
Thanks chap! I've put a few comments in and I've also fixed the failing comparison test on your branch.

I think the execution test is better but it's still not quite what I had in mind; let me explain:

As I understand it `foo-pkg` etc represent node modules? That being the case you'd likely not consume them with raw TypeScript files in; you'd either:

1. Consume a node module with just JS and rely on external type definitions (whether defined by yourself or via an `@types/` scoped package published to npm by Definitely Typed.

Or 

2.  Consume a node module with JS and `.d.ts` definition files in the package too.

If my understanding is correct then ideally our tests should be covering those scenarios rather than exporting raw .ts.

It's possible I have this wrong; feel free to set me straight! :smile:

arcanis:
I think the first scenario you mention isn't relevant to `ts-loader` - it's the `resolveModuleName` implementation that handles the fact that `@types` is tried before other possible options.

I've updated `baz-pkg` to be the case you describe (a `index.js` + a `index.d.ts`), let me know if that works for you 🙂

johnnyreilly:
Cool - thanks for that!  I've just fixed up the linting errors; hopefully we should now see a clean CI 😄 

Question: in the test you only make use of the `moduleName` and `parentResolver` parameter; the others are all pass through parameters:

https://github.com/TypeStrong/ts-loader/blob/d8f3735e305b1b52756513294cabfccb5e1109db/test/execution-tests/3.0.1_resolveModuleName/webpack.config.js#L16-L23

That being the case, does it make sense for the `resolveModuleName` function to have all those parameters?  I'm assuming it does; but I wanted to double check they weren't being passed on for no reason 😄

johnnyreilly:
Ping @arcanis :wink:

arcanis:
Sorry, was flying yesterday 🙂

So it don't use those values yet (and I'm not too familiar with TS to know what they're used for, tbh), but since the TS resolveModuleName hook requires all four of them I figured it'd make sense to also accept them. 

One particular advantage is that it will make it easier for different projects/loaders to use the same custom resolveModuleName implementation (which would be a bit more difficult if they were each expecting a different set of parameters).

johnnyreilly:
That sounds reasonable! Do you have anymore work to do on this or is this ready to merge from your perspective?

arcanis:
Nope, I think it's ready on my side!

Issue #649, opened 2017-10-06
Using InputFileSystem from Webpack


LukeSheard:
In instances.ts the loader reads the file from the fs.readFileSync (https://github.com/TypeStrong/ts-loader/blob/fbd64be5bdb010e79ef0f702896c5d470f6fc619/src/instances.ts#L114). Should this not be using `this.fs` from the web pack InputFileSystem. 

I noticed this when trying to bundle code using a virtual filesystem replacing the InputFileSystem with one that uses an in memory fs, thus files cannot be found.

johnnyreilly:
Tbh I'm not sure - I think that was implemented by @jbrantly and he may be able to advise as to why. If you wanted to submit a PR that changed that we'd definitely consider it. Feel free to open one and we can see where it leads (don't initially worry about the tests - they can be fiddly)

LukeSheard:
Sorry only just had a chance to get back to this, I think it might be a bigger issue where we actually have to create a new `compilerHost` instance and override the typescript `fs` methods that are used there? I can have a go at the simple PR case where we just replace any fs import with using the webpack inputFileSystem though.

johnnyreilly:
Give it a shot - let's see how it pans out!

LukeSheard:
So actually looking at it, you've done most of the work in `servicesHost.ts`, so it should be the case of just overriding those files. I'll see if I can create a demo and a PR and then look at tests!

johnnyreilly:
Sweet!

LukeSheard:
So I made some good progress yesterday and today and put together some simple demos using my fork of ts-loader. https://github.com/LukeSheard/ts-loader-memory-fs/tree/master

As far as I can tell at this point the issue is just the diagnostics which use the `compiler.sys`, so do all the file reads. I think just replacing those with `loader.fs` should work... If you've got any better observations that would be super useful though. I've been having some issues with diagnostic issues when trying to change them over.

johnnyreilly:
> As far as I can tell at this point the issue is just the diagnostics which use the compiler.sys, so do all the file reads. I think just replacing those with loader.fs should work...

Could you elaborate what you mean by this please? An example of the issue you are facing would be helpful.

johnnyreilly:
Btw I'm away from computers right now so feel free to guide me by the hand. Good as the github UI is it's not conducive to grokking code!

LukeSheard:
[...]

PR #864, opened 2018-10-31
[WIP] Support for custom input filesystems

LukeSheard:
This is an extremely long overdue initial PR for #649 with what I hope is some fixes for work we've done internally to support virtual filesystems for webpack through the inputfilesystem property. I'm still sketching out the work to put this back into ts-loader, but essentially it's going to end up completely rewriting the `serviceHost.ts` file where we generate the `LanguageServiceHost` to use webpack's input filesystem instead of `ts.sys` for reading files which uses the underlying `fs`.

More coming, but wanted to get some opinions of work as I complete various steps. I've tried to include a little example here: https://github.com/LukeSheard/ts-loader/tree/master/examples/memory-fs


johnnyreilly:
I'm so glad you're looking into this, thank you! 🌹 

I'm snowed at work right now - but I'll try and take a look at where you've got to this weekend.  Ping me if there's been no action for a while!

examples/memory-fs/README.md
```diff
@@ -0,0 +1,13 @@
+# vanilla ts-loader
+
+```shell
+yarn install
+
+# Run in watch mode
+yarn watch
```
johnnyreilly: Doesn't look like there actually is a watch task?

johnnyreilly:
Also - it looks like the changes only affect the `config.ts` lookup is that sufficient?

opl-:
@LukeSheard Why was this PR closed?

LukeSheard:
@opl- I just didn't have time to work on it anymore. I agree though - if it's possible it totally should.

namesweretaken:
@LukeSheard What's left to do?

PR #871, opened 2018-11-26
Fix projectReferences with rootDir

### LINKED ISSUE #868: Project reference isn't working if `rootDir` is set
### Expected Behaviour
A project reference that has a `rootDir` set should work properly

### Actual Behaviour

'lambda-common' has `outDir` set to `./dist` and `rootDir` set to `./src` and 'lambda-graphql' references 'lambda-common':

```
Error: Could not find output JavaScript file for input ../lambda-common/src/index.ts (looked at ../lambda-common/dist/src/index.js).
The input file is part of a project reference located at ../lambda-common/tsconfig.json, so ts-loader is looking for the project’s pre-built output on disk. Try running `tsc --build` to build project references.
    at successLoader (/lambda-graphql/node_modules/ts-loader/dist/index.js:46:19)
    at Object.loader (/lambda-gaming/lambda-graphql/node_modules/ts-loader/dist/index.js:22:12)
```

`ts-loader` is looking for `dist/src/index.js` instead of `dist/index.js`

Create a project reference to a project that sets `rootDir` to some child folder like `./src`, compile with `tsc -b` and then attempt to use `ts-loader` on the parent project.

andrewbranch:
`rootDir` inside a referenced project’s config wasn’t being considered in figuring out where the output JS files for the project will be, a.k.a. why https://github.com/Microsoft/TypeScript/issues/26410 should still be a thing 🙂 

Fixes #868

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v5.3.1
+
+* [fix: projectReferences with rootDir](https://github.com/TypeStrong/ts-loader/pull/871) (#868) - thanks @andrewbranch!
+
 ## v5.3.0
 
 * [feat: Exposes a `resolveNodeModule` option](https://github.com/TypeStrong/ts-loader/pull/862) - thanks @arcanis!
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "5.3.0",
+  "version": "5.3.1",
   "description": "TypeScript loader for webpack",
   "main": "index.js",
   "types": "dist/types/index.d.ts",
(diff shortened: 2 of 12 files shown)


andrewbranch:
@johnnyreilly seeing one of those minor differences in source mapping metadata on the end of the file in CI... didn't we run into this at some point in #817? It looks familiar but I don't know what you did to fix it.

johnnyreilly:
You rang? 😄 

I've just deleted the `expectedOutput..` folder and regenerated it.  Just regenerating alone doesn't seem to delete files that are no longer generated.  

I've pushed this - let's see how the CI gets on

johnnyreilly:
Hey @andrewbranch,

Are we good to merge and release this? Don't know if there's anything left to do?  (I'm assuming not but I wanted to check)

andrewbranch:
Release away! Thanks for the assist!

johnnyreilly:
Released!

PR #884, opened 2018-12-18
run the tests with node11 as well

zerdos:
Right now the node 11 is an active release (https://github.com/nodejs/Release#release-schedule), so let's run the ts-loader tests on that version as well :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ node_js:
   - "6"
   - "8"
   - "10"
+  - "11"
 sudo: required
 install:
   - yarn install


johnnyreilly:
This is a very @zerdos PR 😊

johnnyreilly:
Kinda weird that the large comparison test fails on node 11. I've no good explanation for that. I don't believe it's a meaningful failure.

I think I'd like to transition the comparison test to be an execution test instead. It doesn't need to be a comparison test; and looking at the nature of the node 11 "failure" an execution test should pass consistently regardless of node version.

PR #889, opened 2018-12-27
Pass ts.Program to getCustomTransformers()

### LINKED ISSUE #860: Need ts.Program or Language Service to create CustomTransformers
### Expected Behaviour

```getCustomTransformers ( (service.LanguageService) => { before?: TransformerFactory[]; after?: TransformerFactory[]; } )```
### Actual Behaviour
```getCustomTransformers ( () => { before?: TransformerFactory[]; after?: TransformerFactory[]; } )```

none
none

----

see [awesome-typescript-loader](https://github.com/kimamula/ts-transformer-keys#webpack-with-awesome-typescript-loader) or [rollup](https://github.com/kimamula/ts-transformer-keys#rollup-with-rollup-plugin-typescript2)

andersekdahl:
A lot of transformers requires a Program instance so we pass ts.Program to getCustomTransformers() to be able to pass it on to transformers. Transformers mostly need ts.Program in order to access the type checker.

Fixes #860

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v5.3.3
+
+* [fix: Pass ts.Program to getCustomTransformers](https://github.com/TypeStrong/ts-loader/pull/889) (#860) - thanks @andersekdahl!
+
 ## v5.3.2
 
 * [feat: enable experimentalFileCaching by default](https://github.com/TypeStrong/ts-loader/pull/885) (#868) - thanks @timocov!
--- a/README.md
+++ b/README.md
@@ -252,7 +252,7 @@ This will ensure that the plugin checks for both syntactic errors (eg `const arr
 
 Also, if you are using `thread-loader` in watch mode, remember to set `poolTimeout: Infinity` so workers don't die.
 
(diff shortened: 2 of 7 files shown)


johnnyreilly:
This looks awesome! Thanks for contributing!  

It looks like the test output for the custom transformers needs regenerating. You can find instructions on how to do that here: https://github.com/TypeStrong/ts-loader/blob/master/test/comparison-tests/README.md#regenerating-test-data

andersekdahl:
The expected test output seems correct, as it should make the string literal uppercase but it doesn't. But I get the same failing test on `master` as on my commit. Do you get that as well?

Looked into it briefly and the custom transformer in the tests seems to get called, but the changes are not applied.

andersekdahl:
Never mind, I forgot to set `transformers` in an else branch and was running the tests without building first so I was fooled into believing that every commit had the same failing test. 😄 

Hopefully works now!

johnnyreilly:
Nice one! I'll look to merge and release when I'm back. Thanks!

johnnyreilly:
Thanks! I'll look to ship this with `ts-loader@5.3.3`

PR #893, opened 2019-01-05
release event not available so use push event and filter

johnnyreilly:
- publish event not available-switch to filter on push
- need to get comparison tests passing in a docker container; got execution tests passing which is enough peace of mind to greenlight a release.

It looks like filtering makes subsequent actions in GitHub show as cancelled because the filter item has failed. That seems... Not quite right? Eg:

 "Action `publish' cancelled because action `check for new tag' failed". Believe related to: https://github.com/actions/bin/issues/13
 
I'm just going to hack on this branch until I've got something I'm happy with.  Then I will definitely squash and merge rather than just merge 😄

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.git
+.github
+.vscode
+.test
+node_modules
+examples
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -1,6 +1,7 @@
 workflow "build, test and publish on release" {
   on = "push"
-  resolves = "publish"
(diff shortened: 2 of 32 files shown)


johnnyreilly:
I'm probably going to comment out the `publish` action of this workflow and merge. When the issue with the filter action is fixed then I can look to uncomment the publish step and enjoy the joys of automatic npm publishing 😁. 

Either way, I'm happy with the work done in this PR and would like to merge it to master.

PR #898, opened 2019-01-18
example of filter issue for @mcolyer

johnnyreilly:
PR for @mcolyer to illustrate filtering issue with GitHub Actions.  See https://github.com/TypeStrong/ts-loader/pull/898

Note the `check for new tag` step.  It's ending with neural as a new tag is not being published.  Ideally subsequent steps would also be marked as neutral but alas instead they get marked as failed.


johnnyreilly:
Now seems to work 😄 Merging!

mcolyer:
🎉

Issue #888, opened 2018-12-24
Reduce the number of examples


johnnyreilly:
We have 12 examples in the repo.  For my money this is far too many and I know I'm certainly not maintaining them in a way that makes them useful.

I'm planning to remove most of them and tidy up the few that remain.  Here's what I think it would be good to end up with:

- standard / vanilla ts-loader usage
- usage with fork-ts-checker-webpack-plugin
- project references example (though actually we don't have one yet)

That's it.  I wanted to share this here in case any other people have thoughts / views etc cc @andrewbranch  @kenotron

PR #899, opened 2019-01-19
there's way too many examples

johnnyreilly:
look to remove most of them in order to make getting started less daunting. Fixes https://github.com/TypeStrong/ts-loader/issues/888

@kenotron I believe that the `fork-ts-checker-webpack-plugin` example in this PR may actually be better for incremental builds than your own `fast-incremental-builds` as it uses the new watch API.  I'd like to consolidate the two examples into one really.  I'd appreciate your help if you can advise?  

Side note: @kenotron - I think that you might want to check out the `useTypescriptIncrementalApi` feature of [`fork-ts-checker-webpack-plugin`](https://github.com/Realytics/fork-ts-checker-webpack-plugin) for the improvement to your own dev workflow it can offer.

Details here: https://blog.johnnyreilly.com/2019/01/typescript-and-webpack-watch-it.html
See also: https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/198


donaldpipowitch:
About the `fast-incremental-builds` example. I saw you used `experimentalWatchApi: true` for the `ts-loader`, but not `useTypescriptIncrementalApi: true` for `ForkTsCheckerWebpackPlugin`. Do I only need one of them? Do you know if the `watch` setting of `ForkTsCheckerWebpackPlugin` should still be used with the incremental API? (_"Directories or files to watch by service. Not necessary but improves performance"_)

johnnyreilly:
> the `fast-incremental-builds` example

It predates the `useTypescriptIncrementalApi: true` being available in the `fork-ts-checker-webpack-plugin`. I'm thinking of removing the `fast-incremental-builds` example as it's no longer that.

The `experimentalWatchApi: true` has never worked as well as hoped; I'm not sure it performs significantly better and problems with it have been reported.

> Do I only need one of them? 

They're different things - one for the loader and one for the type checker. I'd only bother with the plugin option myself.

> Do you know if the `watch` setting of `ForkTsCheckerWebpackPlugin` should still be used with the incremental API?

It's not used in the case of the incremental API; if supplied it is ignored.

donaldpipowitch:
Thank you for the explanation!

PR #900, opened 2019-01-23
fixed name of fork-ts-checker-webpack-plugin

donaldpipowitch:
Really just a small typo fix 🙃Thank you for your work.


johnnyreilly:
Good catch! Thanks!

PR #936, opened 2019-05-18
Fix type reference directive resolution

### LINKED ISSUE #919: jest issue - read to end of issue; perhaps related to `resolveTypeReferenceDirectives` addition?
### Expected Behaviour
Build Successfully with no error

### Actual Behaviour
I'm working with webpack + typescript + gulp, and loading .ts and .tsx files with ts-loader.
For most of the time, everything works fine. 
But build RANDOMLY failed on ts2304 error or ts2345. 
I found out that if I don't install types/jest (ts2304: cannot find describe ...etc.) and types/es6-shim (ts2345: something caused by promise I guess), I can always create the exact same build failure. 

I added logs for npm install and ensured that the two types packages are installed successfully when build fails on this error.

So now I guess, it's probably caused by typescript not picking up those two typings? Or some ts-loader related issue?

I don't actually know how to reproduce it.

I already investigated on it for a month.... Really want to know what is happening here....

Many thanks!!!
Lois

### LINKED ISSUE #934: Yarn PnP support - resolveTypeReferenceDirective - "I'm calling for a resolution..."
It all started when @arcanis raised a PR which added Yarn PnP support to ts-loader. To quote the PR:

> This PR adds a new `resolveModuleName` option that can be used to help TypeScript figure out the location of the files on the disk. It mimics the TypeScript API.

https://github.com/TypeStrong/ts-loader/pull/862

Time passed and it was all good. Then the wonderful @arcanis submitted a very similar PR to the `fork-ts-checker-webpack-plugin`:

https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/250

This PR was very similar to the ts-loader one. However, as I was reviewing it I noticed something. https://github.com/Realytics/fork-ts-checker-webpack-plugin/pull/250#issuecomment-485061143

To quote myself: 

> Thanks! I was just looking at the work you did on ts-loader: https://github.com/TypeStrong/ts-loader/pull/862
>
> You only added `resolveModuleName` there. 
>
> This adds 2 options: `resolveModuleNameModule` and `resolveTypeReferenceDirectiveModule`
>
> Does ts-loader need  a `resolveTypeReferenceDirective` to work with pnp as well? Or is it not necessary?

And @arcanis response:

> I think that might be useful yep. When I initially made the PR for ts-loader I wasn't aware of what `resolveTypeReferenceDirective` was used for (`/// <reference types="...">`) - I just discovered it when implementing the feature for fork-ts-checker-webpack-plugin and only because `create-react-app` uses it to automatically define the `*.svg` typings.

So I decided to add the corresponding functionality to ts-loader; I gave it its own `resolveTypeReferenceDirective`:

https://github.com/TypeStrong/ts-loader/pull/921

Now up until this point, when we created a `ServicesHost` in ts-loader we didn't actually supply `resolveTypeReferenceDirective` at all - in fact it was commented out in the codebase as you can (slightly embarrassingly) see here:

https://github.com/TypeStrong/ts-loader/pull/921/files#diff-70b0dbb6219cb6164e67ffb54a8f23c0L140

And it turned out that introducing it at all has had a breaking side effect for jest users:

https://github.com/TypeStrong/ts-loader/issues/919#issuecomment-486653154

Clearly this issue could be "fixed" by just dropping the `resolveTypeReferenceDirectives` from ts-loader. But if we do that, then I'm assuming we break Yarn PnP support too (only for type reference directives that is).

Where I am is here: I can no longer remember the reason I commented out supplying `resolveTypeReferenceDirectives`. Flat out, I just don't remember. And so we finally get to the question:

Is ts-loader using this API correctly? Should we be using differently? Or is it actually not to be used? I'm after guidance. Please educate me!

cc @andrewbranch

andrewbranch:
Fixes #934 
Fixes #919 

There was really no way to know this without debugging, but getting default typeRoots for a project doesn't work if the module resolution host doesn't implement `getCurrentDirectory()`. Arguably the language service API shouldn’t treat it as optional, or should emit a warning if it’s missing, or fall back to the default from `ts.sys` (maybe I'll bring this up to Sheetal, but it certainly wouldn't land until 3.6), but I think a good policy for ts-loader would be to implement _all_ members of any kind of Host, especially if it’s as easy a just including things from `ts.sys`. It _is_ kind of misleading because there are other places (I think `CompilerHost` is an example) where missing optional members get filled in with good defaults.

### DIFF
--- a/src/servicesHost.ts
+++ b/src/servicesHost.ts
@@ -69,7 +69,9 @@ export function makeServicesHost(
     fileExists,
     readFile: readFileWithFallback,
     realpath: compiler.sys.realpath,
-    directoryExists: compiler.sys.directoryExists
+    directoryExists: compiler.sys.directoryExists,
+    getCurrentDirectory: compiler.sys.getCurrentDirectory,
+    getDirectories: compiler.sys.getDirectories
   };
 
   const clearCache = enableFileCaching ? addCache(moduleResolutionHost) : null;
@@ -187,8 +189,11 @@ function makeResolvers(
   ): (typescript.ResolvedTypeReferenceDirective | undefined)[] =>
     typeDirectiveNames.map(
       directive =>
-        resolveTypeReferenceDirective(directive, containingFile)
-          .resolvedTypeReferenceDirective
+        resolveTypeReferenceDirective(
+          directive,
+          containingFile,
+          _redirectedReference
+        ).resolvedTypeReferenceDirective
     );
 
   const resolveModuleName = makeResolveModuleName(
@@ -500,7 +505,8 @@ export function makeWatchHost(
 
 type ResolveTypeReferenceDirective = (
   directive: string,
-  containingFile: string
+  containingFile: string,
+  redirectedReference?: typescript.ResolvedProjectReference
 ) => typescript.ResolvedTypeReferenceDirectiveWithFailedLookupLocations;
 
 function makeResolveTypeReferenceDirective(
@@ -512,16 +518,17 @@ function makeResolveTypeReferenceDirective(
     | undefined
 ): ResolveTypeReferenceDirective {
   if (customResolveTypeReferenceDirective === undefined) {
-    return (directive: string, containingFile: string) =>
+    return (directive, containingFile, redirectedReference) =>
       compiler.resolveTypeReferenceDirective(
         directive,
         containingFile,
         compilerOptions,
-        moduleResolutionHost
+        moduleResolutionHost,
+        redirectedReference
       );
   }
 
-  return (directive: string, containingFile: string) =>
+  return (directive, containingFile) =>
     customResolveTypeReferenceDirective(
       directive,
       containingFile,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v6.0.1
+
+* [Fix issue with `resolveTypeReferenceDirective` causing errors like `Cannot find name 'it'` with Jest](https://github.com/TypeStrong/ts-loader/pull/936) (#934) (#919) - thanks @andrewbranch!
+
 ## v6.0.0
 
 * [Drop support for node < 8.6 related to micromatch upgrade to 4](https://github.com/TypeStrong/ts-loader/pull/930); see: https://github.com/TypeStrong/ts-loader/issues/929
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "6.0.0",
+  "version": "6.0.1",
   "description": "TypeScript loader for webpack",
   "main": "index.js",
   "types": "dist/types/index.d.ts",


src/servicesHost.ts
```diff
@@ -187,8 +189,11 @@ function makeResolvers(
   ): (typescript.ResolvedTypeReferenceDirective | undefined)[] =>
     typeDirectiveNames.map(
       directive =>
-        resolveTypeReferenceDirective(directive, containingFile)
-          .resolvedTypeReferenceDirective
+        resolveTypeReferenceDirective(
+          directive,
+          containingFile,
+          _redirectedReference
```
andrewbranch: This wasn't related to the particular repro provided, but it seemed like a place that might be broken for even edgier edge cases

src/servicesHost.ts
```diff
@@ -512,16 +518,17 @@ function makeResolveTypeReferenceDirective(
     | undefined
 ): ResolveTypeReferenceDirective {
   if (customResolveTypeReferenceDirective === undefined) {
-    return (directive: string, containingFile: string) =>
+    return (directive, containingFile, redirectedReference) =>
```
andrewbranch: Contextual parameter types FTW 😄

johnnyreilly:
Heya @andrewbranch! 

thanks for this! Will take a look now

src/servicesHost.ts
```diff
@@ -69,7 +69,9 @@ export function makeServicesHost(
     fileExists,
     readFile: readFileWithFallback,
     realpath: compiler.sys.realpath,
-    directoryExists: compiler.sys.directoryExists
+    directoryExists: compiler.sys.directoryExists,
+    getCurrentDirectory: compiler.sys.getCurrentDirectory,
+    getDirectories: compiler.sys.getDirectories
   };
```
johnnyreilly: So if I understand this correctly, the fix is the implementation of these: 

```
    directoryExists: compiler.sys.directoryExists,
    getCurrentDirectory: compiler.sys.getCurrentDirectory,
    getDirectories: compiler.sys.getDirectories
```
?

That's interesting.  I wish it was more "fail fast" here if it actually needed these...

andrewbranch: Yep, specifically `getCurrentDirectory`. I agree. I'll suggest something like this to the team.

johnnyreilly:
> Arguably the language service API shouldn’t treat it as optional, or should emit a warning if it’s missing, or fall back to the default from ts.sys 

This is very interesting.  Completely agree

> but I think a good policy for ts-loader would be to implement all members of any kind of Host, especially if it’s as easy a just including things from ts.sys. It is kind of misleading because there are other places (I think CompilerHost is an example) where missing optional members get filled in with good defaults.

Yeah - agree!

andrewbranch:
Done 👍

andrewbranch:
Hold on the release, I've got another fix coming

andrewbranch:
I'm going to mercifully let CI off the hook of finishing testing my updates to the CHANGELOG—save some electricity, you know

PR #937, opened 2019-05-18
Fix missing error output in root project with project references

### LINKED ISSUE #932: No Errors Displayed with Project References
I have a project reference. The reference seems to work correctly. The output seems to correctly include the referenced project. However, strangely, no errors are displayed in the command line interface for the main project itself, as opposed to the referenced project. If I temporarily remove the `projectReferences` option in `webpack.config.js`, the errors are displayed. I'm not sure why this is.

I have included the relevant configuration files below.

Any help would be much appreciated. 👍

#### `package.json`
```JSON
{
	"ts-loader": "6.0.0",
	"typescript": "3.4.5",
	"webpack": "4.31.0",
	"webpack-cli": "3.3.2"
}
```

#### `tsconfig.json`
```JSON
{
	"compilerOptions":
	{
		"target": "es2017",
		"module": "ES6",
		"moduleResolution": "node",
		"baseUrl": "./",
		"outDir": "./build/",
		"lib":
		[
			"es2017"
		],
		"removeComments": true,
		"downlevelIteration": true,
		"alwaysStrict": true,
		"allowJs": true,
		"allowSyntheticDefaultImports": true,
		"paths":
		{
			"common/*": ["../Common/*"]
		},
		"watch": true
	},
	"references":
	[
		{ "path": "../Common" }
	],
	"include":
	[
		"src/**/*"
	]
}
```

#### `webpack.config.js`
```JavaScript
'use strict';

// External Modules
const Path = require('path');
const NodeExternals = require('webpack-node-externals');

// Constants
const IGNORE = /(?:node_modules)$/;

module.exports =
{
	mode: 'development',
	entry: './src/index.ts',
	target: 'node',
	resolve:
	{
		extensions:
		[
			'.js',
			'.ts'
		],
		alias:
		{
			src: __dirname + '/src',
			common: Path.join(__dirname, '../Common')
		}
	},
	output:
	{
		filename: 'index.js',
		path: Path.resolve(__dirname, './')
	},
	watch: true,
	module:
	{
		rules:
		[
			{
				loader: 'ts-loader',
				test: /\.tsx?$/,
				exclude: IGNORE,
				options:
				{
					projectReferences: true
				}
			}
		]
	},
	externals:
	[
		NodeExternals()
	]
};
```

andrewbranch:
Fixes #932

### DIFF
--- a/src/after-compile.ts
+++ b/src/after-compile.ts
@@ -208,7 +208,11 @@ function provideErrorsToWebpack(
     if (program && sourceFile) {
       errors.push(
         ...program!.getSyntacticDiagnostics(sourceFile),
-        ...program!.getSemanticDiagnostics(sourceFile)
+        ...program!
+          .getSemanticDiagnostics(sourceFile)
+          // Output file has not been built from source file - this message is redundant with
+          // program.getOptionsDiagnostics() separately added in instances.ts
+          .filter(({ code }) => code !== 6305)
       );
     }
     if (errors.length > 0) {
--- a/src/after-compile.ts
+++ b/src/after-compile.ts
@@ -1,4 +1,5 @@
 import * as path from 'path';
+import * as ts from 'typescript';
 import * as webpack from 'webpack';
 
 import * as constants from './constants';
@@ -12,6 +13,7 @@ import {
 } from './interfaces';
 import {
   collectAllDependants,
+  ensureProgram,
   formatErrors,
   isUsingProjectReferences
 } from './utils';
@@ -174,8 +176,6 @@ function provideErrorsToWebpack(
 ) {
   const {
     compiler,
-    program,
-    languageService,
     files,
     loaderOptions,
     compilerOptions,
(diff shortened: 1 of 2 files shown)


src/after-compile.ts
```diff
@@ -187,13 +187,14 @@ function provideErrorsToWebpack(
       ? constants.dtsTsTsxJsJsxRegex
       : constants.dtsTsTsxRegex;
 
+  // I’m pretty sure this will never be undefined here
+  const program = ensureProgram(instance);
```
johnnyreilly: I trust you dude 😁

src/after-compile.ts
```diff
@@ -203,16 +204,13 @@ function provideErrorsToWebpack(
       continue;
     }
 
-    const errors =
-      program === undefined
-        ? [
-            ...languageService!.getSyntacticDiagnostics(filePath),
-            ...languageService!.getSemanticDiagnostics(filePath)
-          ]
```
johnnyreilly: Just out of curiosity, can you recall why we used to alternate between `languageService` and `program` here?

andrewbranch: To clarify, these lines weren't the issue, I just updated them for consistency with the changes above, which were the issue.

`languageService.get*Diagnostics()` essentially just calls its internal `program.get*Diagnostics()`. Based on loader options (and I honestly can't remember what options produce what scenarios), we either have a program directly (`instance.program`) or a language service which has a program under the hood (`instance.languageService.getProgram()`). Before, we were branching here in this function based on where the underlying program is. But, we already have a util—`ensureProgram`—that does that branching.

TL;DR: these lines are basically equivalent because of `ensureProgram`; the lines above _should_ have been looking at both `program` and `languageService` but were only looking at `program`.

johnnyreilly:
I'm about to head to bed - the UK has come last in the Eurovision song contest. Clearly I am shellshocked 🙄🙈🙉

Feel free to merge when you're ready. 😀 If you wanted to push out the release you're welcome to; it's simple, just:

- create a new release in https://github.com/TypeStrong/ts-loader/releases
- this will push out a release via our shiny new GitHub action: https://github.com/TypeStrong/ts-loader/actions

PR #942, opened 2019-05-29
Fix Issue 939

### LINKED ISSUE #939: Potential changes required for typescript 3.5.0
### Expected Behaviour
- Specify `incremental` in tsconfig.json file
- Project is able to compile successfully with against typescript@3.5.0

Note that this issue is not present with typescript@3.4.x

### Actual Behaviour
An error similar to the following one is received:
```
[tsl] ERROR
      TS5074: Option '--incremental' can only be specified using tsconfig, emitting to single file or when option `--tsBuildInfoFile` is specified
```

Per conversation with typescript team (microsoft/TypeScript/issues/31447), it appears that this is a change in typescript but a change in ts-loader will be required.

Current workaround is that users will also need to specify `tsBuildInfo` in their package.json (whereas they previously did not).

https://github.com/konpikwastaken/broken-example-ts-loader-tsc

konpikwastaken:
Hopefully this fixes #939 (sorry, dyslexic branch name).

Note that to repro you'll need to make sure that the root level typescript version is set to 3.5.0-rc, otherwise the new set of tests happily pass.

Please do let me know if there's anything else to clean up.


johnnyreilly:
Awesome work! I'm away camping right now and so can't take a look at this at the moment. 

If I haven't reviewed it by the end of next week then please ping me to remind me 😄

Thanks for your efforts!

konpikwastaken:
The issues failing tests seem to be related to missing jasmine types, shouldn't be anything related to my change. Any suggestions to re-queue or fix?

gfx:
Today TypeScript 3.5.1 has been released, I upgraded tsc, run into `TS5074: Option '--incremental' can only be specified using tsconfig, emitting to single file or when option `--tsBuildInfoFile` is specified`, and then found this PR.

Thanks to @konpikwastaken and @johnnyreilly for the quick works!

johnnyreilly:
> The issues failing tests seem to be related to missing jasmine types, shouldn't be anything related to my change. Any suggestions to re-queue or fix?

I've requeued - let's see where it goes!

konpikwastaken:
@johnnyreilly Have the changes you asked staged, will push soon.

I investigated the test failures - looks like the `configFilePath` field has existed for a while on `typescript.CompilerOptions`, just not on the public interface. In early version of typescript (at least 2.8.1), it looks like the `ParsedCommandLine` property `typeAcquisition.enable` defaults to `false` but is `undefined` in later versions which I believe through a [quick](https://github.com/microsoft/TypeScript/blob/b8def16e92f609327971f07232757fa6c7d29a56/src/testRunner/unittests/config/convertTypeAcquisitionFromJson.ts) [scan](https://github.com/microsoft/TypeScript/blob/c0587191fc536ca62b68748b0e47072e6f881968/src/jsTyping/jsTyping.ts) of the typescript repo, it results in auto-discovery of types.  If we set it on these early versions, it also impacts some other type resolution (e.g. you'll have to specify `types` or `typesRoot` explicitly). If you're comfortable checking currently `tsc` version, then we can conditionally set it only on the +3.5.0 branches that rely on it.

tl;dr; Setting this internal field has side effects in earlier typescript versions, breaking existing tsconfigs. Are you comfortable doing a version check and setting field on tsc>3.5.0? `semver` is already a dependency. I'll push these with my changes so you can take a look.

johnnyreilly:
> tl;dr; Setting this internal field has side effects in earlier typescript versions, breaking existing tsconfigs. Are you comfortable doing a version check and setting field on tsc>3.5.0? semver is already a dependency. I'll push these with my changes so you can take a look.

Sounds like a solid strategy. If this gains us passing tests then I think going with this should be fine; we can always adjust later if there's issues.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## v6.0.2
+* [Set configFilePath when reading config file](https://github.com/TypeStrong/ts-loader/pull/942) (#939) @konpikwastaken
```
johnnyreilly: Minor nit - could this be ` - thanks  @konpikwastaken` please? It's ts-loader policy to thank all contributors publicly 🥰

konpikwastaken: Done.

johnnyreilly:
I wouldn't normally attempt to ship a release whilst camping but it looks like people may be being impacted by not having this.... It's actually not a massive change in the end and we have passing tests so I think I'm going to try and push 6.0.2 out.

However, if it causes problems I'm not going to be able to do anything about them... A Google Pixel will only get you so far 😁

Fingers crossed 🤞

johnnyreilly:
https://github.com/TypeStrong/ts-loader/releases/tag/v6.0.2 should appear on npm in the next 30 mins. Thanks for your work! ♥️🌻

jasongrout:
Thanks to both of you!

lppedd:
Specifying only `incremental: true` still causes
```
Option '--incremental' can only be specified using tsconfig, emitting to single file or when option `--tsBuildInfoFile` is specified.
```
I had to set also `tsBuildInfoFile`
```
"incremental": true,
"tsBuildInfoFile": "build/www/app/out-tsc",
```
Is this wanted?

konpikwastaken:
AFAICT this update to ts-loader (released as 6.0.2) should make setting `incremental: true` sufficient for the error to go away. If there's additional scenarios we need to handle, can you please share a repro or open a new issue referencing this PR or any other related content?

PR #951, opened 2019-06-08
Don’t try to resolve project references if project references are disabled in loader options

### LINKED ISSUE #950: Files in project references are resolved to the project’s outDir even when projectReferences is disabled in options
If you have a project reference with a tsconfig that has an `outDir`, and you have `projectReferences: false` in your ts-loader options, and you have `transpileOnly: false`, ts-loader will try to get the emit for the TS file’s declaration file in its outDir.

andrewbranch:
Fixes #950 

Execution test is identical to 3.0.1_projectReferences except for the Webpack loader config.


johnnyreilly:
Great work!

andrewbranch:
It seems like there are a couple other things close to going in, and apparently I'm the only person that found this bug and cared to report it, so I'm ok with holding off on a release if you are—up to you.

johnnyreilly:
Cool with me. Thanks!

PR #953, opened 2019-06-11
Fix tiny typo

sajadtorkamani:


johnnyreilly:
There's nothing wrong with this PR at all. However. This one line was definitely by @jbrantly (the original creator of ts-loader). And though it sounds weird, I kinda don't want to merge this and lose that - pretty much everything has changed in ts-loader but that hasn't....

I am aware this makes zero sense 😁

jbrantly:
Doh! That typo lasted awhile. Don't not fix this on my account 😄

PR #954, opened 2019-06-13
Upgrade typescript version

fa93hws:
Latest version typescript has more strict type check.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "6.0.2",
+  "version": "6.0.3",
   "description": "TypeScript loader for webpack",
   "main": "index.js",
   "types": "dist/types/index.d.ts",
@@ -88,7 +88,7 @@
     "rimraf": "^2.6.2",
     "tslint": "^5.11.0",
     "tslint-config-prettier": "^1.15.0",
-    "typescript": "^3.1.1",
+    "typescript": "^3.5.1",
     "webpack": "^4.5.0",
     "webpack-cli": "^3.1.1"
   },
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "rimraf": "^2.6.2",
     "tslint": "^5.11.0",
     "tslint-config-prettier": "^1.15.0",
-    "typescript": "^3.5.1",
+    "typescript": "^3.5.2",
     "webpack": "^4.5.0",
     "webpack-cli": "^3.1.1"
   },
--- a/src/servicesHost.ts
+++ b/src/servicesHost.ts
@@ -65,7 +65,8 @@ export function makeServicesHost(
     compiler.sys.fileExists(filePathToCheck) ||
     readFile(filePathToCheck) !== undefined;
 
-  const moduleResolutionHost: ModuleResolutionHost = {
+  let clearCache: null | (() => void) = null;
+  let moduleResolutionHost: ModuleResolutionHost = {
     fileExists,
     readFile: readFileWithFallback,
     realpath: compiler.sys.realpath,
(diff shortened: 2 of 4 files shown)


src/servicesHost.ts
```diff
@@ -670,9 +670,19 @@ function addCache(servicesHost: typescript.ModuleResolutionHost) {
       const cache = createCache<ReturnType<typeof originalFunction>>(
         originalFunction
       );
-      servicesHost[
-        functionToCache
-      ] = cache.getCached as typescript.ModuleResolutionHost[CacheableFunction];
+      switch (functionToCache) {
```
johnnyreilly: It's cool that you can make the compiler play ball by doing this.  However, I kind of don't like it as we're essentially writing more code to get round a problem that the compiler is presenting us with....

@andrewbranch can you think of a nicer approach than the one suggested here? (Funnily enough it looks like this compiled fine prior to TS 3.5... Something changed there I'm guessing)

johnnyreilly: Although maybe the approach suggested here is better as it's certainly clearer... It's just the boilerplate I object to I guess.  It's one of those hard to understand compiler messages:

```
Type '((fileName: string) => boolean) | ((path: string) => string) | ((directoryName: string) => boolean) | undefined' is not assignable to type '((fileName: string) => boolean) & ((directoryName: string) => boolean) & ((path: string) => string)'.
  Type 'undefined' is not assignable to type '((fileName: string) => boolean) & ((directoryName: string) => boolean) & ((path: string) => string)'.
    Type 'undefined' is not assignable to type '(fileName: string) => boolean'.ts(2322)
```

andrewbranch:
Yeah, I don't think the loop is really worth it. Personally I’d probably go with something like

```ts
function addCache(servicesHost: typescript.ModuleResolutionHost): {
  moduleResolutionHost: typescript.ModuleResolutionHost,
  clearCache: () => void,
} {
  const clearCacheFunctions: Action[] = [];
  return {
    moduleResolutionHost: {
      ...servicesHost,
      fileExists: createCache(servicesHost.fileExists),
      directoryExists: servicesHost.directoryExists && createCache(servicesHost.directoryExists),
      realpath: servicesHost.realpath && createCache(servicesHost.realpath),
    },
    clearCache: () => clearCacheFunctions.forEach(clear => clear()),
  };

  function createCache<TOut>(originalFunction: (arg: string) => TOut) {
    const cache = new Map<string, TOut>();
    clearCacheFunctions.push(() => cache.clear());
    return function getCached(arg: string) {
      let res = cache.get(arg);
      if (res !== undefined) {
        return res;
      }

      res = originalFunction(arg);
      cache.set(arg, res);
      return res;
    };
  }
}
```

and then at the usage site

```ts
  let clearCache: null | (() => void) = null;
  let moduleResolutionHost: ModuleResolutionHost = {
    fileExists,
    readFile: readFileWithFallback,
    realpath: compiler.sys.realpath,
    directoryExists: compiler.sys.directoryExists,
    getCurrentDirectory: compiler.sys.getCurrentDirectory,
    getDirectories: compiler.sys.getDirectories
  };

  if (enableFileCaching) {
    const cached = addCache(moduleResolutionHost);
    clearCache = cached.clearCache;
    moduleResolutionHost = cached.moduleResolutionHost;
  }
```

I guess that’s slightly more wordy than what’s proposed here, but doesn’t contain any unsafe type assertions.

johnnyreilly:
Thanks for the suggestion @andrewbranch!

@fa93hws do you fancy tweaking your PR with the suggested implementation?

andrewbranch:
I feel like I should add that there are of course lots of different solutions that largely come down to personal preference of coding style. Avoiding unnecessary type assertions is a worthwhile goal, but I’m sure mine is not the most elegant possible solution.

fa93hws:
No problem!

package.json
```diff
@@ -88,7 +88,7 @@
     "rimraf": "^2.6.2",
     "tslint": "^5.11.0",
     "tslint-config-prettier": "^1.15.0",
-    "typescript": "^3.1.1",
+    "typescript": "^3.5.1",
     "webpack": "^4.5.0",
```
johnnyreilly: Shall we go for TypeScript 3.5.2 since it's out?

fa93hws: No problem!

johnnyreilly:
Thanks!

PR #955, opened 2019-06-19
Fix: Increase the instance version when new root files are found

davazp:
This is a minimal fix for #943.

When a new root file is processed, ensure that the instance version is increased. 

It is an alternative to the bigger PR #945.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## v6.0.4
+* [Fix issue when handling files not included in tsconfig.json](https://github.com/TypeStrong/ts-loader/issues/943) (#934) - thanks @davazp!
+
 ## v6.0.3
 * [Upgrade typescript version to 3.5.2](https://github.com/TypeStrong/ts-loader/pull/954) (#954) - thanks @fa93hws
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "6.0.3",
+  "version": "6.0.4",
   "description": "TypeScript loader for webpack",
   "main": "index.js",
   "types": "dist/types/index.d.ts",
--- a/src/index.ts
+++ b/src/index.ts
@@ -68,7 +68,7 @@ function successLoader(
         )
       : rawFilePath;
 
-  const fileVersion = updateFileInCache(filePath, contents, instance);
+  const fileVersion = updateFileInCache(options, filePath, contents, instance);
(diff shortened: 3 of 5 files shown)


davazp:
This fixes the problem by correctly allowing files not covered by tsconfig.json.

There is no conflict with any other option and ts-loader behaves as it used to. 

@johnnyreilly, @andrewbranch

johnnyreilly:
Heya, thanks for this!

It looks like this tweak may be breaking the `nodeModulesMeaningfulErrorWhenImportingTs` test.

davazp:
Ok, this is interesting. It seems very similar to the bug I was trying to fix! Only that in my situation, instead of an explicit `node_modules` directly,  I was using yarn workspaces.

I guess I can normalize the path to remove symlinks and check for _node_modules_  as well.

johnnyreilly:
It doesn't look like your change has resolved the issue? Was it intended to?

davazp:
> It doesn't look like your change has resolved the issue? Was it intended to?
That is strange!  It works for me locally. I'll try to find out more.

> I think my original questions still stand; is this something that we should be concerned about?
I think we should. My changes allow you to compile files that are under node_modules even if `allowTsInNodeModules` is falsed

davazp:
Silly mistake.  Wrong comparison... !  It worked locally because it updated the snapshot  😄 

I updated the PR. 

Should we also check for `/node_modules/` . (for whatever separators the OS has?) . Not likely but a path like `abcnode_modulesxyz` would match this and be considered node_modules/.  I leave it out for another PR maybe.

johnnyreilly:
Yeah - I think that's an edge case. I'm happy with how things are now I think. Do you want to update the `package.json` and `CHANGELOG.md` please?

johnnyreilly:
I'm afraid we already have `6.0.3` published... 6.0.4?

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 # Changelog
 
-## v6.0.3
```
johnnyreilly: Can we keep the 6.0.3 entry in the `CHANGELOG.md` please?

davazp: oh, I messed it in the conflict resolution.

johnnyreilly:
Release going out now - thanks for your help!

https://github.com/TypeStrong/ts-loader/releases/tag/v6.0.4

davazp:
Awesome.  Thank you @johnnyreilly and @andrewbranch for your help,  responsiveness and the nice test suite 🙂

PR #994, opened 2019-09-03
Add output path to vanilla example webpack config

samal84:
The webpack-dev-server uses the root path not the dist folder, so a specific output folder config is needed for the yarn watch command to work.


johnnyreilly:
Thanks!

PR #996, opened 2019-09-03
Expose the typescript diagnostic object to error formatter functions

srgpqt:
This makes it easy to use TypeScript's builtin error formatters. Of particular interest is TypeScript's `formatDiagnosticsWithColorAndContext` function that produces much better error output.

A custom formatter could be defined as:

```js
ts.formatDiagnosticsWithColorAndContext([error.diagnostic], {
    getCurrentDirectory: () => __dirname,
    getCanonicalFileName: (fileName) => path.normalize(fileName),
    getNewLine: () => os.EOL,
})
```


johnnyreilly:
This looks interesting! Could you write an execution test for this please?

You can likely clone this one:

https://github.com/TypeStrong/ts-loader/tree/master/test/comparison-tests/errorFormatter

There's some instructions here: https://github.com/TypeStrong/ts-loader/blob/master/test/comparison-tests/README.md

We only need expected output for the latest version of TypeScript 3.6

srgpqt:
@johnnyreilly here you go :)

johnnyreilly:
Nice work! I can completely see that people might value this.

Would you like to update the `README.md` with details of this extra option please?

Also could you update the `package.json` version to reflect we're adding a new feature? So probably this will be `6.1.0`. Likewise could you update the `CHANGELOG.md` as well?

It's probably worth blogging about this new feature BTW, do you write a blog?

srgpqt:
I've given this some more thought, and now I believe everyone would prefer to have the improved error formatter (`formatDiagnosticsWithColorAndContext`) used by default if available. What do you think?

johnnyreilly:
I'm not sure.... The thing that makes me slightly cautious about that is that this API isn't always available and it could change. If it changes it might break ts-loader consumers on different versions of TypeScript. So the current approach seems safer.

cassiozen:
@johnnyreilly I'm also interested in this feature - beginner question: Why was it marked as stale? What can I do to help?

srgpqt:
Feel free to pick up where I left off!

srgpqt:
Feel free to pick up where I left off!

cassiozen:
I think it was just a matter of updating `package.json` and adding to the changelog. At this point you would also need to fast-forward to the updates on master.

I can do it if you're not interested, but I also don't want to steal your credits by making a new PR.

srgpqt:
I feel like package.json should be updated when preparing a new release, but I suppose I could do it anyway. It just feels wrong to do that when it’s uncertain when the next release to npm would be...

cassiozen:
@johnnyreilly - any directions here?

johnnyreilly:
Would you like to update:

- `package.json`
- `CHANGELOG.md`
- add some usage instructions to the `README.md` please?

Could the docs include an example of how to use `formatDiagnosticsWithColorAndContext` please?

PR #1003, opened 2019-09-13
Fix SolutionBuilder watches

### LINKED ISSUE #998: projectReferences: true causes node to not exit (webpack host)
### Expected Behaviour
Node to exit once webpack finished

### Actual Behaviour
Node just hangs and doesn't exit.
Have tried node 10 and 12. 
Both hang

1. clone this repository
2. yarn install
3. yarn build.
4. cd test/execution-tests/3.6.0_projectReferencesToBeBuilt
5. yarn install
6. ../../../node_modules/.bin/webpack-cli --config webpack.config.js

image

This repo

sheetalkamat:
Fixes #998


src/instances.ts
```diff
@@ -285,10 +285,19 @@ function successfulTypeScriptInstance(
...
       instance
     );
-    instance.solutionBuilder = compiler.createSolutionBuilderWithWatch(
+    instance.solutionBuilder = compiler.createSolutionBuilder(
       instance.solutionBuilderHost,
       [configFilePath],
-      { verbose: true, watch: true }
+      { verbose: true }
+    );
+    loader._compiler.hooks.watchRun.tap(
```
sheetalkamat: I am not sure if this is correct way to do cleanup before watch compilation begins.. Also I think we need to create solutionbuilder host and solutionBuilder both again but I don't know how to check that..

sheetalkamat: Edit i checked this isnt right way to do this. I think we need to create with watch host when running with --watch. Can someone let me know how to detect that here?

andrewbranch: There should be a `watchMode` property on the webpack `Compiler` object: https://github.com/webpack/webpack/pull/8253/files

sheetalkamat: I think even with that we need to implement `watchFile` as `loader.addDepencedency` but that needs more work I think given there are too many places clearing dependencies...

johnnyreilly:
Thanks for this @sheetalkamat!

Copied across from linked issue:

> I remembered that this was todo.. While in normal build scenario non watch makes sense but we would want to handle --watch portion of it since webpack wont know about the referenced project's files (to watch).

From looking at the changes here it mostly seems that be dropping usage of the watch APIs. Whilst this fixes the linked issue does this mean that watch mode doesn't work for project references with this PR? i.e. If I make a change to a dependent project with this change in place, does that mean that changes *will* not be picked up / builds retriggered?

I'm not totally clear on the intent of this PR and so wanted to clarify.

Thanks again so much for your time and work!

Bessonov:
Run in the same issue with hanging node. Tested this PR (a2a37c). Just building works - webpack exists. But if webpack is in watch mode, then recompilation works only for the current project, but not for referenced one. Does it make sense to synchronize webpack watch mode with typescript watch mode?

sublimator:
> Does it make sense to synchronize webpack watch mode with typescript watch mode?
On the surface it would seem so, no?

Are there other contexts other than webpack need to consider as well ? (Personally only ever used ts-loader with webpack)

sheetalkamat:
I am looking into supporting watch mode separately. This mean while should unblock people that just run webpack

sublimator:
@sheetalkamat 

Makes sense, sounds good to me :)

src/instances.ts
```diff
@@ -285,11 +285,12 @@ function successfulTypeScriptInstance(
...
     );
-    instance.solutionBuilder = compiler.createSolutionBuilderWithWatch(
+    instance.solutionBuilder = compiler.createSolutionBuilder(
       instance.solutionBuilderHost,
-      [configFilePath],
-      { verbose: true, watch: true }
+      [instance.configFilePath!],
+      { verbose: true }
     );
+    // TODO:: handle --watch?
```
sublimator: Ja, if it's tricky, maybe just should roll this out in the short term

sublimator:
@Bessonov 
Hanging is a bit painful on CI, but yeah --watch would be --pretty-please-terrific!

johnnyreilly:
I'm happy for this to be merged and released as v6.1.1 - is there anything more to do?

If not, would you like to update the `package.json` / `CHANGELOG.md`?

sublimator:
image

Ran the branch and it fails on the first invokation. 
Works on the subsequent?

sublimator:
```
➜  3.6.0_projectReferencesToBeBuilt git:(fixWatch) ✗ rm lib/tsconfig.tsbuildinfo 
➜  3.6.0_projectReferencesToBeBuilt git:(fixWatch) ✗ rm lib/index.js.map 
➜  3.6.0_projectReferencesToBeBuilt git:(fixWatch) ✗ rm lib/index.d.ts 
➜  3.6.0_projectReferencesToBeBuilt git:(fixWatch) ✗ node ../../../node_modules/.bin/webpack-cli --config webpack.config.js
Hash: 1723e4ce27cdd156732c
Version: webpack 4.20.2
Time: 2933ms
Built at: 09/14/2019 12:36:30 AM
bundle.js  4.27 KiB    main  [emitted]  main
Entrypoint main = bundle.js
[./src/app.ts] 518 bytes {main} [built] [failed] [1 error]

ERROR in ./src/app.ts
Module build failed (from /Users/nicholasdudfield/projects/ts-loader/index.js):
Error: TypeScript emitted no output for /Users/nicholasdudfield/projects/ts-loader/test/execution-tests/3.6.0_projectReferencesToBeBuilt/src/app.ts.
    at makeSourceMapAndFinish (/Users/nicholasdudfield/projects/ts-loader/dist/index.js:78:15)
    at successLoader (/Users/nicholasdudfield/projects/ts-loader/dist/index.js:68:9)
    at Object.loader (/Users/nicholasdudfield/projects/ts-loader/dist/index.js:22:12)
➜  3.6.0_projectReferencesToBeBuilt git:(fixWatch) ✗ node ../../../node_modules/.bin/webpack-cli --config webpack.config.js
Hash: 44ad6121c445f269b58c
Version: webpack 4.20.2
Time: 2207ms
Built at: 09/14/2019 12:36:35 AM
bundle.js  4.39 KiB    main  [emitted]  main
Entrypoint main = bundle.js
[./lib/index.ts] 97 bytes {main} [built]
[./src/app.ts] 221 bytes {main} [built]
➜  3.6.0_projectReferencesToBeBuilt git:(fixWatch) ✗ 
```

Hrmm, something is up

sublimator:
image

And if you delete just one of the js files it gets "stuck", unable to recover with the error above

johnnyreilly:
Thanks for testing and reporting with detail @sublimator. @sheetalkamat - it looks like there's still issues.

sheetalkamat:
Looking into it

sheetalkamat:
Note:: This should fix initial thing but still working on it.. (reporting error and I think watch)

sheetalkamat:
Are there tests reporting error ?

sheetalkamat:
Also note that I know there are tests for project references failing on older build because I was trying to work clean slate.. Once this work is stabilized, will add the deleted code back.

sublimator:
@sheetalkamat 
Let me know when ready for some testing :)
Thanks!

johnnyreilly:
> Are there tests reporting error that I can see sample of to add test to it

I think what you're looking for is a comparison test. There's a specific test that you could clone if you wanted to create a new comparison test which specifically tests for an expected error case:

https://github.com/TypeStrong/ts-loader/tree/master/test/comparison-tests/errors

Details of how you can work with comparison tests can be found here: https://github.com/TypeStrong/ts-loader/blob/master/test/comparison-tests/README.md

Does that give you what you need?

sheetalkamat:
@johnnyreilly Thank you for the pointer.. That's exactly what I needed..

johnnyreilly:
We are green! Are you ready for this to be merged and released @sheetalkamat?

sheetalkamat:
Will look into --watch today and if that takes longer, we can merge this? Will let you know..

johnnyreilly:
Cool - thanks!

test/comparison-tests/projectReferencesWatch/expectedOutput-3.6/patch1/bundle.js
```diff
@@ -0,0 +1,113 @@
...
+
+/***/ "./lib/index.ts":
+/*!**********************!*\
+  !*** ./lib/index.ts ***!
+  \**********************/
+/*! no static exports found */
+/***/ (function(module, exports, __webpack_require__) {
+
+"use strict";
+eval("\r\nexports.__esModule = true;\r\nexports.lib = {\r\n    one: 1,\r\n    two: 2,\r\n    three: 3\r\n};\r\n\n\n//# sourceURL=webpack:///./lib/index.ts?");
```
sheetalkamat: Output Here doesn't reflect what happens if I run it locally ?

test/comparison-tests/projectReferencesWatch/expectedOutput-3.6/patch1/output.txt
```diff
@@ -0,0 +1,10 @@
+    Asset      Size  Chunks             Chunk Names
+bundle.js  4.33 KiB    main  [emitted]  main
+Entrypoint main = bundle.js
+[./app.ts] 169 bytes {main} [built] [1 error]
+[./lib/index.ts] 104 bytes {main}
+
+ERROR in app.ts
```
sheetalkamat: Is this bug in the test runner ? When I run locally same scenario and edit files, the error goes away and bundle has correct contents..

image

sheetalkamat: @johnnyreilly for help

johnnyreilly: It's possible that it is a bug, yes. The nature of the comparison tests is that they're a little flaky / imperfect. I've always meant to looking into replacing them with something slightly more solid but just never had the time.

Sorry if that's impacting you.

I inherited the comparison test infrastructure when I picked up ts-loader. The flakiness aspect is one of the reasons I quickly went on to write the execution tests which are more dependable / reliable. The downside of them is you can't test error scenarios very well.

If you've ideas / suggestions then please feel free to implement. That is always welcome

johnnyreilly: Other note: perhaps this test might benefit from being a "watch" comparison test: https://github.com/TypeStrong/ts-loader/blob/master/test/comparison-tests/README.md#watch-specific-tests

sheetalkamat: @johnnyreilly that's what this test is? did I miss something. I did create patches to be applied and that's where it went wrong

johnnyreilly: Sorry - I didn't notice that. Yeah you may be bumping on a problem with the comparison test infrastructure.

johnnyreilly: `yarn run comparison-tests -- --single-test projectReferencesWatch`

sublimator:
@sheetalkamat 
I tried your latest commit (91fe036) and it still seems to have problems if a compile file is missing: 
image

You can see I deleted one of the compiled js files then ran the webpack command. If I instead delete the tsbuildinfo file, then it can recover. 

Of course `tsc -b` recovers from this situation, so it would be an expectation that webpack/ts-loader can as well ?

test/comparison-tests/projectReferencesWatch/expectedOutput-3.6/patch1/output.txt
johnnyreilly: I deleted the `expectedOutput-3.6` folder in case there was any stale files inside and regenerated the output with this command:

`yarn run comparison-tests -- --save-output --single-test projectReferencesWatch`

test/comparison-tests/projectReferences/expectedOutput-3.6/bundle.js
```diff
@@ -106,7 +106,7 @@ eval("\nexports.__esModule = true;\nvar lib_1 = __webpack_require__(/*! ./lib */
 /***/ (function(module, exports, __webpack_require__) {
 
 "use strict";
-eval("\nexports.__esModule = true;\nexports.lib = {\n    one: 1,\n    two: 2,\n    three: 3\n    // I am adding this comment here by hand to ensure\n    // Webpack is using the JS output for project references\n};\n\n\n//# sourceURL=webpack:///./lib/index.ts?");
+eval("\r\nexports.__esModule = true;\r\nexports.lib = {\r\n    one: 1,\r\n    two: 2,\r\n    three: 3\r\n};\r\n\n\n//# sourceURL=webpack:///./lib/index.ts?");
```
sublimator: windows line line feeds ... "\r\n" 
will that cause an issue on CI?

johnnyreilly: It shouldn't, no we normalize line endings in the test runner when we do the comparison; see: https://github.com/TypeStrong/ts-loader/blob/bfa48c75adc424a33bbc1f0ec3eeddb8ea1c2722/test/comparison-tests/create-and-execute-test.js#L435

test/comparison-tests/projectReferencesWatch/expectedOutput-3.6/patch1/output.txt
johnnyreilly: You say running locally it works.  Let me try:

```
# pwd is C:\source\ts-loader
cd .\test\comparison-tests\projectReferencesWatch\
# pwd is C:\source\ts-loader\test\comparison-tests\projectReferencesWatch
node ..\..\..\node_modules\webpack\bin\webpack -w
```

johnnyreilly: I've been thinking about this.  I think the problem may lie in `patch0` somehow.  Follow my train of thought:

The comparison test does not introduce errors along the way.  It works like this: 
1. set up legitimate mini typescript project
2. in `patch0`, updated the project reference `lib/index.ts` to include a number `four`.
3. update `app.ts` to start consuming the number `four`.

The error we're presented with in `patch1` is this: 

```
ERROR in app.ts
./app.ts
[90m[tsl] [39m[1m[31mERROR[39m[22m[1m[31m in [39m[22m[1m[36mapp.ts(3,46)[39m[22m
[1m[31m      TS2339: Property 'four' does not exist on type '{ one: number; two: number; three: number; }'.[39m[22m
```

Which suggests `patch0` didn't successfully copy over `lib/index.ts`

sheetalkamat:
> @sheetalkamat
> Of course tsc -b recovers from this situation, so it would be an expectation that webpack/ts-loader can as well ?

`tsc -b` does not handle that scenario (Refer https://github.com/microsoft/TypeScript/issues/30602)
```
c:\temp\test>node c:\TypeScript\built\local\tsc.js -b  --clean

c:\temp\test>node c:\TypeScript\built\local\tsc.js -b

c:\temp\test>dir main.js
 Volume in drive C is OSDisk
 Volume Serial Number is F41A-DBC9

 Directory of c:\temp\test

09/18/2019  08:19 AM                43 main.js
               1 File(s)             43 bytes
               0 Dir(s)  809,078,542,336 bytes free

c:\temp\test>del main.js

c:\temp\test>node c:\TypeScript\built\local\tsc.js -b

c:\temp\test>dir main.js
 Volume in drive C is OSDisk
 Volume Serial Number is F41A-DBC9

 Directory of c:\temp\test

File Not Found

c:\temp\test>
```

sublimator:
I tested it. Tsc -b definitely worked for me.
I was on MacOS though.
Currently out at the moment.
Will test again when I get home.

sublimator:
Note that the deleted file in my test was in reference project, not the
primary one, if that means anything?

sublimator:
Still haven’t tested btw

test/comparison-tests/projectReferencesWatch/expectedOutput-3.6/patch1/output.txt
sheetalkamat: This seems issue with `copySync` keeping old timestamps.. The timestamp of ts file is older than generated js file and hence the reference is not built and error shown
```
c:\github\ts-loader\test\comparison-tests\projectReferencesWatch>dir C:\github\ts-loader\.test\projectReferencesWatch\lib\index.ts C:\github\ts-loader\.test\projectReferencesWatch\lib\index.js
 Volume in drive C is OSDisk
 Volume Serial Number is F41A-DBC9

 Directory of C:\github\ts-loader\.test\projectReferencesWatch\lib

09/17/2019  12:31 PM                84 index.ts

 Directory of C:\github\ts-loader\.test\projectReferencesWatch\lib

09/18/2019  12:02 PM               137 index.js
               2 File(s)            221 bytes
               0 Dir(s)  806,273,327,104 bytes free

c:\github\ts-loader\test\comparison-tests\projectReferencesWatch>type C:\github\ts-loader\.test\projectReferencesWatch\lib\index.ts C:\github\ts-loader\.test\projectReferencesWatch\lib\index.js

C:\github\ts-loader\.test\projectReferencesWatch\lib\index.ts


export const lib = {
  one: 1,
  two: 2,
  three: 3,
  four: 4 // Add new number
};

C:\github\ts-loader\.test\projectReferencesWatch\lib\index.js


"use strict";
exports.__esModule = true;
exports.lib = {
    one: 1,
    two: 2,
    three: 3
};
//# sourceMappingURL=index.js.map
c:\github\ts-loader\test\comparison-tests\projectReferencesWatch>                                                     
```

johnnyreilly: I tweaked the test runner to slow it down (applies patches every 10 seconds) like this:

```
function copyPatchOrEndTest(testStagingPath, watcher, testState, done) {
    // check for new files to copy in
    const patchPath = path.join(testStagingPath, 'patch' + testState.iteration);
    if (fs.existsSync(patchPath)) {
        testState.iteration++;

        console.info("patchPath: " + patchPath)

        // can get inconsistent results if copying right away
        setTimeout(function () {
            fs.copySync(patchPath, testStagingPath, { overwrite: true });
        }, 10000);
    }
    else {
        // ...
}
```

This allows me to watch the comparison test as it works.  What it's doing I can't understand right now.  It's like this:

cant-explain

johnnyreilly: Actually - it turns out me slowing it down that much messes up the test.  If I run it at normal speed it seems to be doing the right thing.  But it just doesn't seem to be picking up the changes.....

johnnyreilly: I think maybe the comparison test runner is fine.  I think this might be a bug. It looks as though the `lib/index.ts` file is being patched, the `lib/index.d.ts` file is not being regenerated.  Take a look:

actually-maybe-it-is-a-bug

johnnyreilly: Yeah - I was able to repeat this locally.  Whilst webpack *is* emitting `lib/index.d.ts`, `lib/index.d.ts` has *not* be recompiled.  It never gets updated from the initial compilation:

the-d-ts-files-are-not-regenerated

johnnyreilly: Sorry I missed your comment about `copySync` timestamps before.

When I reproduced locally I copied the exact file over - would that be the same behaviour?

sheetalkamat: Fixed it :)

johnnyreilly:
@sublimator have you had chance to test with @sheetalkamat's latest changes? It sounds like we may be there, I'd love to get you to try it out and confirm.

sheetalkamat:
@johnnyreilly I think this should be it for this PR,., This still doesn't emit `.tsbuildinfo` for current program when using `experimentalWatchAPI`( note its not possible to get the .tsbuildinfo from language service as its not meant for that). But that's a separate issue than supporting project references.

sheetalkamat:
@sublimator as mentioned at https://github.com/TypeStrong/ts-loader/pull/1003#issuecomment-532739996 deleting js file from referenced project's output is not supported scenario even with `tsc --b`

sublimator:
@sheetalkamat 
Sorry, just got to my computer

I am referring to a leaf project as per:
image

Although, editing that test project to make the entry point "main" tsconfig.json have compilerOptions `composite: true` seems to work too?? 
image

Is this a windows vs MacOS thing ??

sublimator:
@sheetalkamat 
https://github.com/microsoft/TypeScript/blob/master/src/compiler/diagnosticMessages.json#L4055-L4059

sheetalkamat:
> @sheetalkamat
> https://github.com/microsoft/TypeScript/blob/master/src/compiler/diagnosticMessages.json#L4055-L4059


That is the message that indicates to build but does not really build.. Check out the  microsoft/TypeScript#30602 as per my earlier comment. that's not something that would be fixed here.

sublimator:
@sheetalkamat 
image
I don't know why I'm getting different results than you, but it's definitely building for me ??

johnnyreilly:
Let's get this merged and released. There's been a ton of progress here - thanks so much @sheetalkamat! If there's further things to do they can be addressed subsequently.

sublimator:
Yeah, don't mean to come off as ungrateful :) 
Just reporting my findings. I can't explain it ?? MacOS vs Windows differences ??
Having a play now with the 'sweet path'

sublimator:
--watch mode seems to be working, and picking up changes!
normal mode seems to be working and picking up changes!

PR #1008, opened 2019-09-18
fix: no errors reported if flagged with @ts-check

### LINKED ISSUE #1004: no errors TypeScript errors reported for .js when enabling checks using // @ts-check
### Expected Behaviour
TypeScript errors shall be reported for .js files, if `"checkJS": false` in tsconfig.json but enabled for a few files selectively using `// @ts-check` .

### Actual Behaviour
no TypeScript errors are reported for such .js files

Build the repo linked below, the error in [test.js](https://github.com/reinholdk/ts-loader-sample/blob/416b140e4b7f76c54c59a3174706e0820d0000f0/src/test.js#L4) is not reported.

https://github.com/reinholdk/ts-loader-sample

reinholdk:
fixes #1004

### DIFF
--- a/src/after-compile.ts
+++ b/src/after-compile.ts
@@ -183,7 +183,7 @@ function provideErrorsToWebpack(
   } = instance;
 
   const filePathRegex =
-    compilerOptions.checkJs === true
+    compilerOptions.allowJs === true
       ? constants.dtsTsTsxJsJsxRegex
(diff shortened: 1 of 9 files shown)


johnnyreilly:
Awesome!  Thanks!  This will go out with the next release (v6.1.1)

Issue #1014, opened 2019-09-20
compilation using project references fails if a tsconfig.json includes an external .d.ts.


gvinaccia:
### Expected Behaviour

successful compilation.

### Actual Behaviour

Error: Debug Failure. False expression.
    at Object.getOutputDeclarationFileName

### Steps to Reproduce the Problem

just clone the repository and run:
```
yarn install 
yarn build
```

### Location of a Minimal Repository that Demonstrates the Issue.
https://github.com/gvinaccia/ts-loader-issue

PR #1015, opened 2019-09-20
fixes https://github.com/TypeStrong/ts-loader/issues/1014

gvinaccia:
it seems that ts-loader is  trying to emit declaration files for a declaration file, this causes typescript's getOutputDeclarationFileName throw an error

### DIFF
--- a/src/instances.ts
+++ b/src/instances.ts
@@ -567,7 +567,7 @@ export function isReferencedFile(instance: TSInstance, filePath: string) {
 }
 
 export function getEmitOutput(instance: TSInstance, filePath: string) {
-  if (fileExtensionIs(filePath, typescript.Extension.Dts)) {
+  if (fileExtensionIs(filePath, instance.compiler.Extension.Dts)) {
     return [];
   }
   const program = ensureProgram(instance);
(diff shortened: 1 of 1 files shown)


src/instances.ts
```diff
@@ -507,7 +507,11 @@ function getOutputFileNames(
     }
     if (
       (configFile.options.declaration || configFile.options.composite) &&
-      (instance.compiler as any).hasTSFileExtension(inputFileName)
```
sheetalkamat: You want to add this part to `getEmitOutput` to return [] early even before getting program rather than here.

gvinaccia: done, thank you

src/instances.ts
```diff
@@ -567,6 +567,9 @@ export function isReferencedFile(instance: TSInstance, filePath: string) {
 }
 
 export function getEmitOutput(instance: TSInstance, filePath: string) {
+  if (fileExtensionIs(filePath, typescript.Extension.Dts)) {
```
sheetalkamat: Sorry missed this earlier but for extension use `instance.compiler.Extension.Dts` `typescript` is used only for types for tsloader code.. The compiler = typescript of the compilation

gvinaccia: done, thanks for your time

johnnyreilly:
This looks great!

PR #1016, opened 2019-09-20
Consume typescript apis from typescript nightly

sheetalkamat:
Added apis from microsoft/TypeScript#33503

Also added --match-test to run tests matched by regexp passed in

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v6.1.2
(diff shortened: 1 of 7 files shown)


johnnyreilly:
Thanks! Nice work on `--match-test` - good addition 🙂

johnnyreilly:
Let's plan to get this out as 6.1.2. it would be good to include https://github.com/TypeStrong/ts-loader/pull/1015 with that.

PR #1018, opened 2019-09-27
converted main.workflow to Actions V2 yml files

johnnyreilly:
This PR migrates the GitHub actions to use the non-deprecated YML syntax.

See: 
https://help.github.com/en/articles/migrating-github-actions-from-hcl-syntax-to-yaml-syntax


johnnyreilly:
Having a few problems getting this working so this it only publishes on a release.

https://github.com/actions/starter-workflows/blob/master/ci/npm-publish.yml

Found something that might help here: https://github.com/actions/starter-workflows/issues/112

PR #1028, opened 2019-10-22
WIP for supporting project references further

sheetalkamat:
Note there are console.log statements in the loader to debug this.
Also fyi this is not fully working


Teoxoy:
Will this PR fix the issue brought up in https://github.com/TypeStrong/ts-loader/pull/1003#issuecomment-531295558?

> Whilst this fixes the webpack not exiting with project references issue does this mean that watch mode doesn't work for project references with this PR? i.e. If I make a change to a dependent project with this change in place, does that mean that changes will not be picked up / builds retriggered?

I stumbled across this issue too.

FrozenKiwi:
+1.

Is there anything we can do to help this along?  I've started poking around the insides of ts-loader, but it's a lot to take in quickly.  If you give some pointers, I'd be more than happy to try and contribute

Issue #962, opened 2019-07-10
LoaderOptions.compilerOptions typed incorrectly


slavafomin:
Hello!

Thank you for this great loader.

However, it looks like the `LoaderOptions.compilerOptions` property is typed incorrectly. The `target` property is typed as `ScriptTarget`, which is:

```typescript
enum ScriptTarget {
    ES3 = 0,
    ES5 = 1,
    ES2015 = 2,
    ES2016 = 3,
    ES2017 = 4,
    ES2018 = 5,
    ES2019 = 6,
    ES2020 = 7,
    ESNext = 8,
    JSON = 100,
    Latest = 8
}
```

So, it requires the number value. However, in runtime it's actually expecting a string value, like `'ES2015'`. This leads to an error: `TS5024: Compiler option 'target' requires a value of type string.`

Please, see my issue in TypeScript project:
https://github.com/microsoft/TypeScript/issues/32317

johnnyreilly:
Do you want to submit a PR to fix this?

ark120202:
It's actually not just `target`, it also affects all other enums and `lib` option (during tsconfig parsing raw names like `es2015` are changed to file names, like `lib.es2015.d.ts`), so `compilerOptions` section of tsconfig isn't the same thing as `ts.CompilerOptions`. TypeScript doesn't have a type that reflects tsconfig structure (see https://github.com/microsoft/TypeScript/pull/32501), so I see 3 options:
- Manually maintain a config type
- Generate it from `ts.optionDeclarations` internal value
- Actually use `LoaderOptions.compilerOptions` as `ts.CompilerOptions`: merge it with options from config file after parsing (which would be a breaking change)

slavafomin:
I've released [a package](https://github.com/moebius-mlm/ts-compiler-options) with proper compiler options types. Please, review [the PR](https://github.com/TypeStrong/ts-loader/pull/1033).

PR #1033, opened 2019-11-13
Fixed compiler options typings

slavafomin:
TypeScript declarations doesn't provide proper types for `CompilerOptions` that can be used to describe content of `tsconfig.json` file, instead, they provide internal representation of CompilerOptions, where enum values must be specified using numbers instead of strings.

I've created a [custom package](https://github.com/moebius-mlm/ts-compiler-options) that exports `CompilerOptions` in the format [described in TypeScript documentation](https://www.typescriptlang.org/docs/handbook/compiler-options.html).

This PR replaces `CompilerOptions` types from `typescript` to [`@moebius/ts-compiler-options`](https://github.com/moebius-mlm/ts-compiler-options). This should effectively fix typing issues discussed in #962.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/.idea/
 /*.js
 !index.js
 /src/*.js
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
   },
   "homepage": "https://github.com/TypeStrong/ts-loader",
   "dependencies": {
-    "@moebius/ts-compiler-options": "^1.1.0",
+    "@moebius/ts-compiler-options": "^1.2.0",
     "chalk": "^2.3.0",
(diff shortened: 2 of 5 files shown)


johnnyreilly:
Thanks for this but we don't want to create another dependency unless we really need to. There's no guarantee that this dependency will be updated on an ongoing basis and so I'd rather avoid it. Thanks for the suggestion though.

slavafomin:
Feel free to merge it into your codebase then. I've spent my very limited and expensive time to fix this issue.

johnnyreilly:
Thanks - i appreciate your effort. However if I understand this correctly this works around an issue in the TypeScript types. I'd rather not do that in case this is fixed upstream by then at some point.

Thanks anyway.

PR #1040, opened 2019-12-11
Expose loader context to custom transformer

xbtsw:
It is necessary to expose webpack loader context to custom transformer factory function. For example, when the transformer examine the AST and need to call `loaderContext.addDependency()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## v6.2.2
+* [Expose loader context to custom transformer](https://github.com/TypeStrong/ts-loader/pull/1040) - thanks @xbtsw!
+
 ## v6.2.1
 * [Output types alongside JS files, enable declaration maps](https://github.com/TypeStrong/ts-loader/pull/1026) - thanks @meyer!
 
--- a/README.md
+++ b/README.md
@@ -329,7 +329,7 @@ Also, if you are using `thread-loader` in watch mode, remember to set `poolTimeo
 
 These options should be functions which will be used to resolve the import statements and the `<reference types="...">` directives instead of the default TypeScript implementation. It's not intended that these will typically be used by a user of `ts-loader` - they exist to facilitate functionality such as [Yarn Plug’n’Play](https://yarnpkg.com/en/docs/pnp).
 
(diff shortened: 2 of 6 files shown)


johnnyreilly:
Could you give us a clearer example of the problem that this solves please?  It sound very interesting but I don't quite follow what you're trying to achieve....

xbtsw:
@johnnyreilly, let's say I have a transformer, which whenever I see a magic comment in typescript AST `/* inline: myconstants:myId */`. Then I will look up a JSON file, `/src/myconstants.json` and find the corresponding field in the JSON (`myId`) and replace the comment with the value of the field.

So with JSON of
```
// /src/myconstants.json
{
  "myId": "testValue"
}
```
with original source
```
// /src/mycode.ts
console.log(/* inline: myconstants:myId */);
```
After compile it will be
```
console.log("testValue");
```

Now this could be done solely by using a transformer. However, in order for webpack to correctly understand the dependency (and make the watch scenarios work). I also need to mark `/src/myconstants.json` as a dependency of `/src/mycode.ts`. For that, I will need to use the webpack loader API (`loaderContext.addDependency()`), which is what this PR made possible.

johnnyreilly:
Thanks for the explanation! Could you add a comparison test that illustrates this please? You could use this as a basis: https://github.com/TypeStrong/ts-loader/tree/master/test/comparison-tests/customTransformer

You only need create it for TypeScript v3.6 (current test pack version)

xbtsw:
I already modified the current transformer test to assert the context being passed. `ts-loader` doesn't do the full workflow here. It just need to pass the context to enable the custom transformer.

johnnyreilly:
Okay cool. Could you update the version number in the `package.json` and the `CHANGELOG.md` as well please? Remember to thank yourself!

xbtsw:
@johnnyreilly Updated, thanks 👍

xbtsw:
Actually, I just realized this doesn't work. Once the compiler instance is created, the loaderContext will never be passed to the transformer again by compiler when invoking the compiler.

Zn4rK:
@xbtsw Can I ask - did you ever get this to work with another method?
 I'm in need of something like this...

xbtsw:
Yes I did. Will submit a PR later.

PR #1032, opened 2019-11-04
feat(docs): add table of contents to README

jsjoeio:
This PR adds a table of contents to the README (auto-generated using [`github-markdown-toc`](https://github.com/ekalinin/github-markdown-toc)).

It also makes the heading of the README look a little better. 

Resolves one task from #1024 

### Gif

Here's the package `markdown-toc` in action:
ezgif com-optimize

**UPDATE**
I automated it by adding a `pre-push` hook that runs this script:
```json
"generate-toc": "markdown-toc -i ./README.md && git add README.md && git commit -m 'chore: update docs'"
```

It's super fast! And I did `pre-push` because I figured `pre-commit` might be too much.

PR #1044, opened 2019-12-21
fix links to heading IDs

sheonhan:
In the Table of Contents, links to heading IDs had superfluous underscores, so the links weren't jumping to corresponding sections in the README.


johnnyreilly:
Thanks for this! This arrived with https://github.com/TypeStrong/ts-loader/pull/1032

The tricky thing is that the Table of Contents isn't hand cranked but generated using `github-markdown-toc`. If we fix it manually then presumably the issue will present again next time the TOC is regenerated.

Is there another way we could fix this "at source" as it were? cc @jsjoeio

jsjoeio:
Thanks for catching this @sheonhan and @johnnyreilly for tagging me. 

Apologies - when I added it in, I didn't test every link. The majority were working so I thought it was good. I'll take a look to see what I can do to fix the underscores in the unique headings.

sheonhan:
Thanks for such a quick response, everyone! I'll close this PR for now.

PR #1045, opened 2019-12-22
Fix Underscores in Headings

jsjoeio:
### Description
The goal of this is to propose a new format for part of the `README`, which will:
- improve readability
- fix issues with the autogenerated TOC

### Screencast

Here's what it will look like:


jsjoeio:
@johnnyreilly would love to get your thoughts on this before proceeding further.

Essentially, I propose going from this:
"
#### transpileOnly _(boolean) (default=false)_
"
to this:
"
#### transpileOnly
| Type | Default Value |
|------|--------------|
| `boolean` | `false`|

"

I chose this solution because:
a) it doesn't seem like we could fix this with the CLI based on the docs for [`markdown-toc`](https://github.com/jonschlinkert/markdown-toc#cli)
b) I think this could improve readability potentially
c) it would fix the same issue #1044 brought up

johnnyreilly:
Yeah I think this is great! Thanks for proposing it!  Do you want to submit a full PR?

Thanks again!

jsjoeio:
Woo! Definitely. Going to update then will mark ready for review when it's complete!

README.md
```diff
@@ -435,20 +471,32 @@ Does not compute.... code: 2307,severity: error,content: Cannot find module 'com
...
+| `object` | `{}`|
 
 Allows overriding TypeScript options. Should be specified in the same format
 as you would do for the `compilerOptions` property in tsconfig.json.
 
-#### instance _(string)_
+#### instance
+| Type | Default Value |
+|------|--------------|
+| `string` | `TODO`|
```
jsjoeio: Does this option have a default value? I didn't see one in the heading and wanted to ask. 

If not, I can remove that column

jsjoeio:
@johnnyreilly this is ready to go!

johnnyreilly:
Nice one! I'll merge as soon as CI finishes doing it's thing. Thanks!

README.md
jsjoeio: tagging @johnnyreilly just in case you didn't see this

Issue #1056, opened 2020-02-19
Supporting '.js' imports


guybedford:
TypeScript are currently recommending that the best way to handle Node.js ES module support output is to use explicit .js extensions in the imports, even between TypeScript files (see https://github.com/microsoft/TypeScript/issues/16577#issuecomment-309169829):

main.ts
```js
import dep from './dep.js';
```

dep.ts
```js
export default 'dependency';
```

This is the only pattern that guarantees that extensions are included in a tsc compilation, making the output work in Node.js modules.

I just tried this out with ts-loader in ncc, and am getting that this is not supported.

I think it would be ideal if this project could follow that same pattern.

johnnyreilly:
If you'd like to send a PR that adds support we're open to it! Don't worry about tests initially - start with getting it working and I'll help out 🤗

guybedford:
Closing as a non-bug since it appears this is working correctly after all!

HKalbasi:
It isn't working. I need to use `"module": "ES2020",` in my tsconfig, for enabling top level await support of webpack, and I need to remove extensions for this plugin to work. Annoyingly, I have nodejs files in the same project and for that to work, I need to add extension to make nodejs ESM happy. So I need to remember which file is node and which file is browser. Since extensions are ES standard, I think I should solve the problem with webpack and this plugin.

PR #1058, opened 2020-02-20
feat: support .js extensions from .ts files

guybedford:
Thanks for the quick response on https://github.com/TypeStrong/ts-loader/issues/1056 - I went ahead and put together the PR.

The gist of the logic is this - when importing a file through the .ts resolver, TypeScript will strip `.js` extensions such that they can still resolve to `.ts` files or `.js` files depending on what exists on the file system.


guybedford:
Hmm, actually I did some more digging and I believe this test is passing just fine without this change.... so I think we are good, it's just ncc support that is needed explicitly.

PR #1083, opened 2020-04-07
fix: watch-run

### LINKED ISSUE #1082: Watch-run doesn't update correctly
This issue is born from https://github.com/cevek/ttypescript/issues/86. This comment in particular describes the problem a bit better: https://github.com/cevek/ttypescript/issues/86#issuecomment-609432990

It's not caused by some internal order that I wrote there.

### Expected Behaviour
I expect that files that I've changed is updated correctly.

I've not managed to reliably reproduce this with the testing strategy that `ts-loader` uses.

Clone https://github.com/Zn4rK/ttypescript-repro

Run `npm run start:webpack`

---

So, this is a strange one. I've spent some time trying to understand what's happening. 

Looking at the source for watch-run:\
https://github.com/TypeStrong/ts-loader/blob/a43736e4cfb9c38dc60ade7b9b38b8d6e9c915a4/src/watch-run.ts#L20-L21

It basically says: If date is more than the lastTime AND filePath is a tsTsxJsJsx file, skip. 

For modified files, date will always be more than the last time.

Looking at [one of the tests](https://github.com/TypeStrong/ts-loader/tree/87340f49e2ebfd5158c26dbcfef1b2beb177f7c3/test/comparison-tests/importsWatch/expectedOutput-3.8) that fails if I change the if-statement, I would argue that the current result is wrong. 

If I understand the test suite correctly, we should see something along the lines of:

```
TS2322: Type '"foobar"' is not assignable to type 'boolean'.
```

In `patch1/output.txt` but right now that doesn't happen. There is no expected outputs for either patch0 or patch1. 

There are a few more tests that has inconsistent results with what I would assume should be happening.

My suggestion is simply to change the first for-loop to:
```js
    for (const [filePath, date] of times) {
      const lastTime = lastTimes.get(filePath) || startTime;

      if (date <= lastTime) {
        continue;
      }

      lastTimes.set(filePath, date);
      updateFile(instance, filePath);
    }
```

Which seems to fix my original issue, and get the expected test results for (some of) the existing tests as well. I don't really understand how the comparison tests works with the `_WatchAPI` syntax.

I've removed the match for tsTsxJsJsx files completely, because I don't think that the check was intentional.

Zn4rK:
Closes #1082 

I need a bit of help verifying that these tests are correct. I tried to understand what's happening with _WatchApi, but couldn't wrap my head around it. 

It looks scary to change `/projectReferencesWatch_Composite_WatchApi/expectedOutput-3.8/patch4/output.txt` to not emit an error, but `projectReferencesWatch` still has the corresponding error.

### DIFF
--- a/test/comparison-tests/projectReferencesWatch_Composite_WatchApi/expectedOutput-3.8/patch4/output.txt
+++ b/test/comparison-tests/projectReferencesWatch_Composite_WatchApi/expectedOutput-3.8/patch4/output.txt
@@ -1,12 +1,6 @@
-                  Asset      Size  Chunks             Chunk Names
-              bundle.js  4.38 KiB    main  [emitted]  main
-            ../app.d.ts  11 bytes          [emitted]  
-../tsconfig.tsbuildinfo  51.8 KiB          [emitted]  
+      Asset      Size  Chunks             Chunk Names
+  bundle.js  4.31 KiB    main  [emitted]  main
+../app.d.ts  11 bytes          [emitted]  
 Entrypoint main = bundle.js
-[./app.ts] 186 bytes {main} [built] [1 error]
-[./lib/index.ts] 132 bytes {main}
-
-ERROR in app.ts
-./app.ts
-[90m[tsl] [39m[1m[31mERROR[39m[22m[1m[31m in [39m[22m[1m[36mapp.ts(3,56)[39m[22m
-[1m[31m      TS2551: Property 'ffive' does not exist on type '{ one: number; two: number; three: number; four: number; five: number; }'. Did you mean 'five'?[39m[22m
+[./app.ts] 131 bytes {main} [built]
+[./lib/index.ts] 123 bytes {main}
--- a/test/comparison-tests/projectReferencesWatch_WatchApi/expectedOutput-3.8/patch4/output.txt
+++ b/test/comparison-tests/projectReferencesWatch_WatchApi/expectedOutput-3.8/patch4/output.txt
@@ -1,10 +1,5 @@
     Asset      Size  Chunks             Chunk Names
-bundle.js  4.38 KiB    main  [emitted]  main
+bundle.js  4.31 KiB    main  [emitted]  main
 Entrypoint main = bundle.js
-[./app.ts] 186 bytes {main} [built] [1 error]
-[./lib/index.ts] 132 bytes {main}
-
-ERROR in app.ts
-./app.ts
-[90m[tsl] [39m[1m[31mERROR[39m[22m[1m[31m in [39m[22m[1m[36mapp.ts(3,56)[39m[22m
-[1m[31m      TS2551: Property 'ffive' does not exist on type '{ one: number; two: number; three: number; four: number; five: number; }'. Did you mean 'five'?[39m[22m
+[./app.ts] 131 bytes {main} [built]
+[./lib/index.ts] 123 bytes {main}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## v7.0.1
+* [fix: watch-run](https://github.com/TypeStrong/ts-loader/pull/1083) - thanks @zn4rk
+
 ## v7.0.0
 * [Project reference support enhancements](https://github.com/TypeStrong/ts-loader/pull/1076) - thanks @sheetalkamat!
 * Following the end of life of Node 8, `ts-loader` no longer supports Node 8 **BREAKING CHANGE**
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "7.0.0",
+  "version": "7.0.1",
   "description": "TypeScript loader for webpack",
   "main": "index.js",
   "types": "dist",
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "ts"
   ],
   "engines": {
-    "node": ">=8.6"
+    "node": ">=10.0.0"
   },
   "author": "John Reilly <johnny_reilly@hotmail.com> (https://blog.johnnyreilly.com)",
   "contributors": [
--- a/src/watch-run.ts
+++ b/src/watch-run.ts
@@ -14,15 +14,15 @@ export function makeWatchRun(instance: TSInstance) {
   const startTime = 0;
 
   return (compiler: webpack.Compiler, callback: () => void) => {
-    const times = compiler.fileTimestamps;
     if (instance.loaderOptions.transpileOnly) {
       instance.reportTranspileErrors = true;
     } else {
+      const times = compiler.fileTimestamps;
+
       for (const [filePath, date] of times) {
-        if (
-          date > (lastTimes.get(filePath) || startTime) &&
-          filePath.match(constants.tsTsxJsJsxRegex) !== null
-        ) {
+        const lastTime = lastTimes.get(filePath) || startTime;
+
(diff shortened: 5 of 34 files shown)


johnnyreilly:
Thanks for contributing! I must admit to wondering what the consequence of removing the regex would be... I have a feeling I put that in for a reason but, shame on me, I didn't put a comment to explain 😭

Zn4rK:
@johnnyreilly Yeah, I tried a bunch of stuff to try and figure out why it was there to begin with.
What essentially is let through the conditionals are files like these:

```
// many more of these, on every single watchRun
filePath node_modules/typescript/lib/README.md date: 499162500010
filePath node_modules/typescript/lib/diagnosticMessages.generated.json, date: 499162500010
filePath node_modules/typescript/lib/typesMap.json date: 499162500010
```

Since the first check doesn't include equality, they are updated on every "watch-run". 

The file in cache eventually gets updated by the successLoader, but that's after it's needed.

johnnyreilly:
Cool - thanks for the explanation. We've got some big stuff happening with project references right now; see this PR: https://github.com/TypeStrong/ts-loader/pull/1076

I'd like to get that landed and then take a look at this afterwards. Please bear with us!

Zn4rK:
Alright. Sounds good! I'll just use my patch for now. 

And FYI I noticed that my excerpt above only included files from `typescript`, but the list contains a whole bunch of files from many different packages.

There's another conditional further down that I think can be removed completely as well.
If the file has changed since the last time it was included (i.e. since the initial build), my "fix" would fix it.

The typescript program needs to have a new version of the file that's been changed before emitting basically.

johnnyreilly:
v7.0.0 is now merged - I think you may have some merge conflicts I'm afraid!

Zn4rK:
I've rebased and updated, so I guess we're only waiting on Appveyor and Travis! :)

johnnyreilly:
Heya! 

I'll admit to some nervousness around the change to this test output here:

https://github.com/Zn4rK/ts-loader/blob/82c192229be94edec5286e10698fa18230d33eb7/test/comparison-tests/projectReferencesWatch_WatchApi/expectedOutput-3.8/patch4/output.txt

Is this what you meant when you said:

> It looks scary to change `/projectReferencesWatch_Composite_WatchApi/expectedOutput-3.8/patch4/output.txt` to not emit an error, but `projectReferencesWatch` still has the corresponding error.

I am aware of some quirkiness in our comparison tests which @sheetalkamat bumped on here: https://github.com/TypeStrong/ts-loader/pull/1076#issuecomment-612991360

That change was rolled back later on. So this may be fine... Thoughts?

Zn4rK:
> Heya!
> 
> I'll admit to some nervousness around the change to this test output here:
> 
> https://github.com/Zn4rK/ts-loader/blob/82c192229be94edec5286e10698fa18230d33eb7/test/comparison-tests/projectReferencesWatch_WatchApi/expectedOutput-3.8/patch4/output.txt

> Is this what you meant when you said [...]

I understand your nervousness and I'm a bit nervous too! But I've tested this pretty extensively and I'm fairly confident that it should work.

That's basically what I meant. I just tested this manually, and ideally this file should be reverted to emit an error - since it still does that (as it should). 

Screenshot 2020-04-19 at 18 11 31

> I am aware of some quirkiness in our comparison tests which @sheetalkamat bumped on here: [#1076 (comment)](https://github.com/TypeStrong/ts-loader/pull/1076#issuecomment-612991360)

We could maybe make the test flaky instead and take another look at improving the comparison tests in the future?

johnnyreilly:
Let's roll with it as is for now. Worst comes to worst we can revert this later, but I'll go 🤞 it's down to the quirkiness of the comparison test pack.

Could you update the `package.json` to `7.0.1` please? And add an entry to `CHANGELOG.md` - don't forget to thank yourself!

Zn4rK:
> Let's roll with it as is for now. Worst comes to worst we can revert this later, but I'll go 🤞 it's down to the quirkiness of the comparison test pack.

Fair enough! :) I guess we just have to remember that it actually should emit an error. But the correct expected output still exists in `projectReferencesWatch`.

> Could you update the `package.json` to `7.0.1` please? And add an entry to `CHANGELOG.md` - don't forget to thank yourself!

Done! 👍Thank you for taking the time with this!

johnnyreilly:
Released! https://github.com/TypeStrong/ts-loader/releases/tag/v7.0.1

test/comparison-tests/projectReferencesWatch_Composite_WatchApi/expectedOutput-3.8/patch4/output.txt
```diff
@@ -1,12 +1,6 @@
-                  Asset      Size  Chunks             Chunk Names
-              bundle.js  4.38 KiB    main  [emitted]  main
-            ../app.d.ts  11 bytes          [emitted]  
-../tsconfig.tsbuildinfo  51.8 KiB          [emitted]  
+      Asset      Size  Chunks             Chunk Names
+  bundle.js  4.31 KiB    main  [emitted]  main
+../app.d.ts  11 bytes          [emitted]  
 Entrypoint main = bundle.js
-[./app.ts] 186 bytes {main} [built] [1 error]
```
sheetalkamat, line 6: How is this correct change.. Is this issue with the test (where it baselines wrong result) or the change?

test/comparison-tests/projectReferencesWatch_WatchApi/expectedOutput-3.8/patch4/output.txt
```diff
@@ -1,10 +1,5 @@
     Asset      Size  Chunks             Chunk Names
-bundle.js  4.38 KiB    main  [emitted]  main
+bundle.js  4.31 KiB    main  [emitted]  main
 Entrypoint main = bundle.js
-[./app.ts] 186 bytes {main} [built] [1 error]
```
sheetalkamat, line 4: Same here?

test/comparison-tests/projectReferencesWatch_Composite_WatchApi/expectedOutput-3.8/patch4/output.txt
Zn4rK: I believe that's an issue with this test. It should (and still does) produce the expected error. We could revert this make it flaky instead?

test/comparison-tests/projectReferencesWatch_WatchApi/expectedOutput-3.8/patch4/output.txt
Zn4rK: Same as above.

test/comparison-tests/projectReferencesWatch_Composite_WatchApi/expectedOutput-3.8/patch4/output.txt
Zn4rK: I've manually verified `projectReferencesWatch_Composite_WatchApi` and  `projectReferencesWatch_WatchApi`, both emit the error when running "manually" but not when using the comparison tests.

johnnyreilly: > Is this issue with the test (where it baselines wrong result) or the change?

I think the problem may be the comparison test pack which seems to be demonstrating unhelpful (incorrect) behaviour. The comparison test pack is not as reliable as I'd like (and I'm partly responsible).

This doesn't give me a warm fluffy feeling but I don't believe it represents a bug in ts-loader.

PR #1101, opened 2020-05-06
Ensure that JSON files are included in build module resolution (#905)

### LINKED ISSUE #905: TS6307: JSON files not in file list when using composite projects
### Expected Behaviour

Using `resolveJsonModule` and `composite` projects works as `tsc` does. 

Note that `tsc` has had some issues with this that were solved in  https://github.com/Microsoft/TypeScript/issues/25636 and that `tsc` works correctly now. See reproduction below. 

### Actual Behaviour

```
ERROR in ./tsloader-ts6307/tsconfig.composite.json
[tsl] ERROR
      TS6307: File 'tsloader-ts6307/src/test.json' is not in project file list. Projects must list all files or use an 'include' pattern.
```

```sh
$  git clone git@github.com:thovden/tsloader-ts6307.git
$ cd tsloader-ts6307/
$ npm i

# tsc works well for the composite config
$ npm run tsc

# webpack + ts-loader works without composite set
$ npm run works
[./src/test.json] 20 bytes {main} [built]
    + 12 hidden modules
ℹ ｢wdm｣: Compiled successfully.

# webpack + ts-loader fails with composite
$ npm run fails
ERROR in ts6307/tsconfig.fail.json
[tsl] ERROR
      TS6307: File 'ts6307/src/test.json' is not in project file list. Projects must list all files or use an 'include' pattern.
ℹ ｢wdm｣: Failed to compile.
```

https://github.com/thovden/tsloader-ts6307

berickson1:
Ensure that json files are resolved if resolveJsonModule flag is set in tsconfig

Fixes #905


andrewbranch:
Thanks so much for doing this @berickson1! Do you think you’d be able to [add a comparison test](https://github.com/TypeStrong/ts-loader/blob/master/test/comparison-tests/README.md)?

berickson1:
@andrewbranch - Added a comparison-test for the positive case (& confirmed it didn't work before this change). Should I also add one for when resolveJsonModule is false?

test/comparison-tests/resolveJsonModule/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "compilerOptions": {
+    "resolveJsonModule": true,
+    "composite": true
+  },
+  "include": ["ap.ts", "file.json"]
```
andrewbranch, line 6:
```suggestion
  "include": ["app.ts", "file.json"]
```

johnnyreilly:
Thanks!

johnnyreilly:
Shipping with https://github.com/TypeStrong/ts-loader/releases/tag/v7.0.3 ⛴️

Issue #1102, opened 2020-05-07
No longer possible to use different loader options for different files


appzuka:
### Expected Behaviour

It should be possible to specify different options for each file type in webpack.config:

module: {
    rules: [
      { test: /\.md$/, use: [{loader: "ts-loader", options: {logLevel: 'WARN', transpileOnly: true}}, "otherLoader"]},
      { test: /\.ts$/, use: [{loader: "ts-loader", options: {logLevel: 'INFO', transpileOnly: false}}, "otherLoader"]}
    ]
}

### Actual Behaviour

Since 7.0.0 this no longer works.  successLoader in index.ts no longer accepts the options as a parameter so the loaderOptions will always be the options specified the first time ts_loader is invoked.

### Steps to Reproduce the Problem

Create a webpack config which needs different options for different invocations of ts-loader

This could be fixed by adding a line to getTypeScriptInstance in instances.js:

function getTypeScriptInstance(loaderOptions, loader) {
    if (instances.hasOwnProperty(loaderOptions.instance)) {
        const instance = instances[loaderOptions.instance];
        instance.loaderOptions = loaderOptions; // << Add this line

johnnyreilly:
Thanks for the detailed issue - that's very helpful! Would you like to submit a PR?

appzuka:
I have submitted a PR.  The updated code passes the execution tests and also works in my application, which depends on this behaviour.

I note that the comparison tests don't work since 7.0.1.

appzuka:
I have generated a new pull request including an execution test.

While working on this I found that my previous patch was not sufficient if the changed options were functions (such as getCustomTransformers) as these are only attached when a new instance is created.

I also found that an option already exists to manage this behaviour, the 'instance' option.  However, it is easy to miss that this is necessary so my new pull request automatically creates a new instance whenever the options change.

PR #1103, opened 2020-05-08
Fix instances to allow different loader options

appzuka:
It should be possible to specify different options for each file type in webpack.config:

module: {
rules: [
{ test: /.md$/, use: [{loader: "ts-loader", options: {logLevel: 'WARN', transpileOnly: true}}, "otherLoader"]},
{ test: /.ts$/, use: [{loader: "ts-loader", options: {logLevel: 'INFO', transpileOnly: false}}, "otherLoader"]}
]
}

Actual Behaviour
Since 7.0.0 this no longer works. successLoader in index.ts no longer accepts the options as a parameter so the loaderOptions will always be the options specified the first time ts_loader is invoked.

This commit updates instances.ts to use the current loaderoptions for each instance.


johnnyreilly:
Thanks for the PR related to https://github.com/TypeStrong/ts-loader/issues/1102#issuecomment-625831876

I'm not sure what you mean by

> I note that the comparison tests don't work since 7.0.1

The tests appear to work fine?

Could you add a test that covers this case please? That way we can prevent future regression. Execution tests are preferred if possible as they're faster and more reliable. However they don't fit in all cases; this may be one ☺️

PR #1104, opened 2020-05-10
Ensure a separate webpack instance is created for different loader options

appzuka:
This patch automatically creates a new webpack instance if ts-loader is invoked with different loader options.

This behaviour could be created manually by including the 'instance' option on each loader, but it is easy to miss that this is necessary.  My application failed with update 7.0.0 because I used 2 different sets of options and did not appreciate that the second set would be ignored unless I specified a different instance string.

This update tags the instance with a hash of the options specified and will create a new instance if the hashes do not match.  This behaviour will function even if the options specified are functions, such as getCustomTransformers.

A new execution test, loaderOptions, has been added to test this functionality.


johnnyreilly:
Looks like you've a lint issue?

https://github.com/TypeStrong/ts-loader/issues/1102#issuecomment-626352317

johnnyreilly:
This is awesome work! Thank you! 🥰

Could you update the version number in the `package.json` and add an entry to the `CHANGELOG.md` please?

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## v7.0.4
+* [Ensure a separate webpack instance is created for different loader options](https://github.com/TypeStrong/ts-loader/pull/1104) - @appzuka
```
johnnyreilly, line 4: Can you thank yourself please? It's traditional now 🤗

johnnyreilly:
Shipping with https://github.com/TypeStrong/ts-loader/releases/tag/v7.0.4

Thanks for your help 🤗🌻

appzuka:
I am pleased to be able to help.  If there are other issues you don't have time to look into, in this or other repos, please let me know and I'll investigate when I have some time.

johnnyreilly:
Thanks @appzuka! All help is always gratefully received 🥰

There's always issues and I'm massively time poor. If you should spot something that catches your eye, do feel free to have a go!

Thanks again 😊

PR #1115, opened 2020-05-21
Apply other loaders when updating files in watch mode

### LINKED ISSUE #1111: Other loaders are not applied to unchanged files in watch mode
### Expected Behaviour
In watch mode, unchanged .ts files should be passed to loaders other than `ts-loader`, as in v6.2.2.

### Actual Behaviour
Unchanged .ts files are not passed to other loaders.

Please see the below repository. The version of `ts-loader` is `7.0.4`.

`./app.ts` and `./error.ts` are transpiled using `./remove-error-loader.js` and `ts-loader` (in this order). `./error.ts` has an "error", but it is removed by `./remove-error-loader.js`.
When I run `npx webpack --watch`, it seems to be working well at first:
```
% npx webpack --watch

webpack is watching the files…

Hash: f695893648cef87b9d90
Version: webpack 4.43.0
Time: 981ms
Built at: 2020/05/17 22:18:19
bundle.js  8.93 KiB    main  [emitted]  main
Entrypoint main = bundle.js
[./app.ts] 90 bytes {main} [built]
[./error.ts] 0 bytes {main} [built]
```
However, when I update `./app.ts` (for example, change `'Hello'` to `'Hello, world!'`), an error occurs:
```
Hash: a3f00bec5e5882c9d219
Version: webpack 4.43.0
Time: 39ms
Built at: 2020/05/17 22:18:25
bundle.js  8.95 KiB    main  [emitted]  main
Entrypoint main = bundle.js
[./app.ts] 98 bytes {main} [built]
    + 1 hidden module

ERROR in C:\Users\iorate\ts-loader-error\error.ts
./error.ts
[tsl] ERROR in C:\Users\iorate\ts-loader-error\error.ts(1,1)
      TS2552: Cannot find name 'error'. Did you mean 'Error'?
```
This means that `./error.ts` (unchanged) is compiled by a TypeScript compiler, but not processed by `./remove-error-loader.js` before that.

With `ts-loader@6.2.2`, an update of `./app.ts` does not cause an error:
```
Hash: 412257a27b607df16d5e
Version: webpack 4.43.0
Time: 29ms
Built at: 2020/05/17 22:16:30
bundle.js  8.95 KiB    main  [emitted]  main
Entrypoint main = bundle.js
[./app.ts] 98 bytes {main} [built]
    + 1 hidden module
```

Is this regression, or expected behavior in v7?

https://github.com/iorate/ts-loader-error

iorate:
Fix #1111 

Currently watch-run does not apply loaders other than ts-loader to TypeScript files when updating them in watch mode.
This behavior may lead to complication errors when files are "preprocessed" by other loaders (e.g. [ifdef-loader](https://github.com/nippur72/ifdef-loader)).


iorate:
I need help... :cry:

One test failed in Travis CI + Node.js v10 + TypeScript v3.8.2, while all tests passed in Travis CI + Node.js v12 + TypeScript v3.8.2 and AppVeyor + Node.js v10 + TypeScript 3.8.2.

The first comparison test with `transpileOnly` failed:
```
  aliasResolution
    ✓ should have the correct output (4391ms)
    1) should work with transpileOnly
  1 passing (7s)
  1 failing
  1) aliasResolution
       should work with transpileOnly:
      Uncaught AssertionError [ERR_ASSERTION]: patch0/output.txt is different between actual and expected
      + expected - actual
           Asset     Size  Chunks             Chunk Names
       bundle.js A-NUMBER-OF KiB main  [emitted]  main
       Entrypoint main = bundle.js
      -[./app.ts] A-NUMBER-OF bytes {main} [built]
      +[./app.ts] A-NUMBER-OF bytes {main}
       [./common/components/myComponent.ts] A-NUMBER-OF bytes {main} [built]
```

I don't understand why `[built]` was output only in this environment.

johnnyreilly:
Comparison tests are somewhat flaky - requeued

johnnyreilly:
Thanks for the PR! We really appreciate your time. I can see this adds a dependency to https://github.com/webpack/loader-runner#readme

I don't know much about `loader-runner` and we tend to be cautious about adding dependencies because they don't all have great futures 😉

Can you provide some details about what loader-runner does, and some details around whether you believe it has a bright future? I ask as webpack 5 is on the horizon and I am curious as to whether loader-runner remains relevant. I have looked at the loader-runner repo but there aren't tons of docs. 

You may want to reach out to @sokra and see if he can advise.

iorate:
Thank you @johnnyreilly ! All tests have passed.

> what loader-runner does
> whether you believe it has a bright future

It applies webpack loaders to a resource and returns a result. webpack uses it internally to apply loaders ([lib/NormalModule.js@4.43.0](https://github.com/webpack/webpack/blob/c9d4ff7b054fc581c96ce0e53432d44f9dd8ca72/lib/NormalModule.js#L16)). So it is the best way to run loaders in the same manner as webpack.

I think it has a good future. webpack v5.0.0-beta.16 also depends on loader-runner ([lib/NormalModule.js@5.0.0-beta.16](https://github.com/webpack/webpack/blob/b7f08b42b85f89e52facac426251d26ec64f2e6d/lib/NormalModule.js#L9)). Changelog of webpack 5 explicitly says loader-runner was upgraded ([link](https://github.com/webpack/changelog-v5#other-minor-changes)). We can expect it to be used by upcoming webpack 5.

src/watch-run.ts
```diff
@@ -37,26 +47,58 @@ export function makeWatchRun(instance: TSInstance) {
...
+  return new Promise<void>((resolve, reject) => {
+    if (instance.rootFileNames.has(path.normalize(filePath))) {
+      loaderRunner.runLoaders(
+        {
+          resource: filePath,
+          loaders: loader.loaders
+            .slice(loaderIndex + 1)
+            .map(l => ({ loader: l.path, options: l.options })),
+          context: {},
+          readResource: fs.readFile.bind(fs)
```
johnnyreilly, line 83: It looks like this will create a new instance of `readFile` each time `updateFile` runs. Would it be more performant to create this once and reuse it? In fact is the `bind` necessary at all?

src/watch-run.ts
```diff
@@ -27,7 +37,7 @@ export function makeWatchRun(instance: TSInstance) {
         }
 
         lastTimes.set(filePath, date);
-        updateFile(instance, filePath);
+        promises.push(updateFile(instance, loader, loaderIndex, filePath));
```
johnnyreilly, line 40: What is the significance of making this array of `Promise`s vs just calling `updateFile` there and then?

sokra:
Instead of using loader-runner directly, use `this.loadModule` from loader context. Otherwise you are missing out webpack and other plugin modifications to the loader context.

johnnyreilly:
Thanks @sokra - really appreciate you advising 🌻🥰

iorate:
@sokra 
Thank you for your advice.

However, `this.loadModule` does not seem to fit my purpose.
What I attempt is to manually get a TypeScript file preprocessed by loaders other than ts-loader.
When a rule is, for example,
```javascript
{
  test: /\.ts$/,
  use: ['ts-loader', 'preprocess-ts-loader']
}
```
I want a result of 'preprocess-ts-loader'.
`this.loadModule('!!preprocess-ts-loader!unpreprocessed-ts-file.ts', (err, source) => { ... })` seems to handle the result as a JavaScript module (actually it is a TypeScript file), to fail in most cases.

> Otherwise you are missing out webpack and other plugin modifications to the loader context.

I can understand it, but I expect simply running `runLoaders` will work for ts-loader and the few TypeScript-preprocessing loaders in most cases.

src/watch-run.ts
iorate: It may be a bit efficient to cache the result of `fs.readFile.bind(fs)` as you say.
I think `bind` is necessary, because there is no guarantee that `fs.readFile` does not use `this` internally.

... In the first place, `readResource` is not necessary, because it is optional and defaulted to `readFile`. As [@types/loader-runner](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/loader-runner/index.d.ts#L25) does not mark it as optional, I pass the default value explicitly to avoid a type error.
This may be better:
```
runLoaders(
  {
    resource: ...,
    loaders: ...
  } as loaderRunner.RunLoaderOption,
```

src/watch-run.ts
iorate: `runLoaders` called in `updateFile` is an async function. We must wait for all of them to finish before calling `callback`.
To implement waiting for multiple async functions, I think `Promise.all` is a good way.

sokra:
Put a stringify loader in front to make it valid js. JSON.parse the result. This has been done before in some css related loader.

iorate:
@sokra I really appreciate your kindness. The "stringify loader" approach is working well for my problem.

iorate:
@johnnyreilly I replaced `runLoaders` with `loadModule`, thanks to @sokra.

(Please ignore the second commit, in which I tried raw-loader as a stringify loader)

src/watch-run.ts
```diff
@@ -37,26 +45,56 @@ export function makeWatchRun(instance: TSInstance) {
...
-  );
+function updateFile(
+  instance: TSInstance,
+  filePath: string,
+  loader: webpack.loader.LoaderContext,
+  loaderIndex: number
+) {
+  return new Promise<void>((resolve, reject) => {
+    if (
+      loaderIndex + 1 < loader.loaders.length &&
```
johnnyreilly, line 74: Could you put a comment here to clarify the purpose of the 2 code branches here please?

iorate: OK, I will.

iorate:
Added a comment to `updateFile`.

src/watch-run.ts
```diff
@@ -37,26 +45,60 @@ export function makeWatchRun(instance: TSInstance) {
...
     // Update all the watched files from solution builder
     if (instance.solutionBuilderHost) {
       for (const filePath of instance.solutionBuilderHost.watchedFiles.keys()) {
-        updateFile(instance, filePath);
+        promises.push(updateFile(instance, filePath, loader, loaderIndex));
       }
     }
 
-    callback();
+    Promise.all(promises)
```
johnnyreilly, line 60: I can't remember if we've discussed this already, are there any issues with us moving to a `Promise` (ie async) approach instead of the current approach? Given we're dealing with callbacks I suspect not, but I wanted to check

iorate: I think there are no new issues introduced by the 'async' approach.

iorate:
May I "thank myself" in the changelog?
```
* [Apply other loaders when updating files in watch mode](https://github.com/TypeStrong/ts-loader/pull/1115) - thanks @iorate 
```

johnnyreilly:
That's the tradition!

iorate:
Updated `package.json` and `CHANGELOG.md`. Thank you.

johnnyreilly:
Will ship with https://github.com/TypeStrong/ts-loader/releases/tag/v7.0.5 - thanks for your help! :sunflower:

PR #1135, opened 2020-07-01
Fix webpack deprecations (fix #1126)

### LINKED ISSUE #1126: Deprecation Warning on Webpack 5
### Expected Behaviour
Clean output without deprecation warning.

### Actual Behaviour

After webpack update (from v4 to v5) I'm getting deprecation warnings in my console.

```
(node:7361) [DEP_WEBPACK_DEPRECATION_ARRAY_TO_SET] DeprecationWarning: Compilation.modules was changed from Array to Set (using Array method 'reduce' is deprecated)

(node:41349) [DEP_WEBPACK_MODULE_ERRORS] DeprecationWarning: Module.errors was removed (use getErrors instead)
```

Using tracing:

```sh
node --trace-deprecation node_modules/webpack/bin/webpack.js
```

Output:

```
(node:41650) [DEP_WEBPACK_DEPRECATION_ARRAY_TO_SET] DeprecationWarning: Compilation.modules was changed from Array to Set (using Array method 'reduce' is deprecated)
    at Set.set.<computed> [as reduce] (/Users/.../myproject/node_modules/webpack/lib/util/deprecation.js:78:4)
    at determineModules (/Users/.../myproject/node_modules/ts-loader/dist/after-compile.js:58:32)
```

```
(node:41650) [DEP_WEBPACK_MODULE_ERRORS] DeprecationWarning: Module.errors was removed (use getErrors instead)
    at /Users/.../myproject/node_modules/ts-loader/dist/after-compile.js:140:45
    at Array.forEach (<anonymous>)
```

### Environment

```
Node.js v12.16.2
webpack@5.0.0-beta.16
ts-loader@7.0.5
```

g-plane:
This PR should remove some of webpack deprecation warnings.

### DIFF
--- a/src/after-compile.ts
+++ b/src/after-compile.ts
@@ -189,7 +189,7 @@ function determineFilesToCheckForErrors(
 }
 
 function provideErrorsToWebpack(
-  loader: webpack.loader.LoaderContext,
+  loader: webpack.loader.LoaderContext | null,
   filesToCheckForErrors: TSFiles,
   filesWithErrors: TSFiles,
   compilation: webpack.compilation.Compilation,
@@ -246,9 +246,12 @@ function provideErrorsToWebpack(
           { module },
           compilation.compiler.context
         );
-        formattedErrors.forEach(({ message }) => {
-          loader.emitError(message);
-        });
+
+        if (loader) {
+          formattedErrors.forEach(({ message }) => {
+            loader.emitError(message);
+          });
+        }
 
         compilation.errors.push(...formattedErrors);
       });
@@ -269,7 +272,7 @@ function provideErrorsToWebpack(
 }
 
 function provideSolutionErrorsToWebpack(
-  loader: webpack.loader.LoaderContext,
+  loader: webpack.loader.LoaderContext | null,
   compilation: webpack.compilation.Compilation,
   modules: Map<FilePathKey, WebpackModule[]>,
   instance: TSInstance
@@ -304,9 +307,12 @@ function provideSolutionErrorsToWebpack(
           { module },
           compilation.compiler.context
         );
-        formattedErrors.forEach(({ message }) => {
-          loader.emitError(message);
-        });
+
+        if (loader) {
+          formattedErrors.forEach(({ message }) => {
+            loader.emitError(message);
+          });
+        }
 
         compilation.errors.push(...formattedErrors);
       });
--- a/src/after-compile.ts
+++ b/src/after-compile.ts
@@ -20,7 +20,6 @@ import {
 } from './utils';
 
 export function makeAfterCompile(
-  loader: webpack.loader.LoaderContext,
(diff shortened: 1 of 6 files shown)


johnnyreilly:
@sheetalkamat has a PR which drops support for older versions of TypeScript here: https://github.com/TypeStrong/ts-loader/pull/1136

When that's merged we can take a look at your PR - it only seems to be failing on older versions of TypeScript.

g-plane:
It seems that these are weird failures, and I'm not sure whether upgrading webpack will fix these or not.

johnnyreilly:
Very strange....

src/after-compile.ts
```diff
@@ -63,6 +61,7 @@ export function makeAfterCompile(
 
     const filesWithErrors: TSFiles = new Map();
     provideErrorsToWebpack(
+      loader,
```
sheetalkamat, line 64: why do you need to pass this as instance.loader is there?

g-plane:
I've confirmed that upgrading webpack will solve the failures in CI, as those failures also can be reproduced on my local computer. Can I open another PR to upgrade webpack? If not, we may need to add some hacky code to check the loader context.

johnnyreilly:
Yeah sure - I'd like to decouple the upgrade of webpack from this change if that's okay. So upgrade webpack first and loop back to this next.

g-plane:
PR for upgrading webpack is created: #1140 .

g-plane:
I've marked this PR as ready for review, however I don't know how to address the problems at AppVeyor.

johnnyreilly:
AppVeyor has had a bad day - let's see if it fixes up now.

g-plane:
Maybe we can migrate to GitHub Actions in the future.

johnnyreilly:
We're already using GitHub actions, but it doesn't cover the use case that AppVeyor provides; namely running CI on Windows.

Maybe Azure pipelines has something to offer - I haven't taken a look in a while. I did start looking here though:

https://github.com/TypeStrong/ts-loader/pull/856

g-plane:
I will. By the way, GitHub Actions supports Linux, Windows and macOS.

johnnyreilly:
Really? Did not know that! That said, I never managed to get comparison tests passing with GitHub actions. But that was the early days

g-plane:
Maybe we can try it after this PR.

g-plane:
I've updated `package.json` and changelog. Is this OK?

johnnyreilly:
Looks good! Let's wait for Travis to get happy then we can merge

johnnyreilly:
Right now I want AppVeyor to die

g-plane:
All passed now.

johnnyreilly:
Shipping https://github.com/TypeStrong/ts-loader/releases/tag/v8.0.1 - thanks for your work!

PR #1141, opened 2020-07-10
Change README and examples from obsolete checkSyntacticErrors to syntactic diagnostic option

ansf:
`checkSyntacticErrors` option is not available anymore. We are supposed to use diagnosticOptions now (see https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/issues/477)

This PR updates README and thread-loader example accordingly.


johnnyreilly:
Thanks!

PR #1144, opened 2020-07-15
update GitHub Actions

g-plane:
Switched to GitHub Actions.


g-plane:
All tests which ran on GitHub Actions passed. AppVeyor and Travis CI won't succeed because I've deleted their configuration files.

johnnyreilly:
Nice work! 

Could you undelete the Travis and AppVeyor files please? I'd like to run them side by side for a while before decommissioning (though I'll probably not require them to pass before merging; just interested in looking at the difference)

g-plane:
I will.

While this question is not related to this PR, are there any plans for migrating the comparison tests to snapshot tests with Jest?

johnnyreilly:
> are there any plans for migrating the comparison tests to snapshot tests with Jest?

Great question! I've long pondered doing something like that. It looks like webpack uses jest:

https://github.com/webpack/webpack/blob/master/package.json#L122

I guess the problems to solve here are:

- the output contains file paths and escaped file paths and similar. That's what all this stuff is about: https://github.com/TypeStrong/ts-loader/blob/master/test/comparison-tests/create-and-execute-test.js#L379 - we'd need a way to perform similar regex voodoo on the jest snapshots
- the comparison test pack has the concept of "patches". Which means applying a series of changes to the source files so we can trigger the compilation again and harvest output once again. Finding a way (and I'm sure there is) to plug Jest snapshots into that as well.

If you wanted to take a look, we'd greatly appreciate it! 🌻❤️

g-plane:
Yes, I know. And, can this PR be merged?

johnnyreilly:
Nice work! Thank you 😄 

Assuming all goes well we'll look to drop AppVeyor soon

PR #1170, opened 2020-08-14
Fix getOptionsHash when two options has different props but same values.

meowtec:
Fix the issue when I use both `lang="tsx"` and `lang="ts"` in vue single file components.
For `lang="ts"`: 
```
{
  appendTsSuffixTo: [/\.vue$/],
  transpileOnly: true,
}
```

For `lang="tsx"`: 
```
{
  appendTsxSuffixTo: [/\.vue$/],
  transpileOnly: true,
}
```

They get the same hash because `getOptionsHash` use only values for hashing.
So for tsx vue file, it reuses the loader options which is only for ts vue, which add `.ts` suffix not `.tsx`.
That can lead to wrong compilation result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## v8.0.3
+* [Fix the wrong instance caching when using `appendTsSuffixTo` and `appendTsxSuffixTo` together](https://github.com/TypeStrong/ts-loader/pull/1170) - thanks @meowtec
+
 ## v8.0.2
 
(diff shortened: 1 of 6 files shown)


meowtec:
There is still something wrong, let me find it out.

johnnyreilly (approved):

johnnyreilly:
Just published https://github.com/TypeStrong/ts-loader/releases/tag/v8.0.3 - thanks for your work! 🤗🌻

Issue #1158, opened 2020-07-22
`transpileOnly` does not work with `projectReferences`


chancancode:
### Steps to Reproduce the Problem

Set up a project that uses `transpileOnly: true` and `projectReferences: true`. Something like this:

```js
// root tsconfig.json
{
  "include": ["index.ts"],
  "compilerOptions": {
    "outDir": "dist",
    "references": [
      { "path": "./packages/sub-project" },
      // ...
    ]
  }
}
```

```js
// packages/sub-project/tsconfig.json
{
  "compilerOptions": {
    "outDir": "dist",
    "composite": true,
    "declaration": true
  }
}
```

### Expected Behaviour

When a type error is introduced in either the root project or the sub-project, the build should work and emit the JS files despite the type error.

### Actual Behaviour

When a type error is introduced in either the root project, the build would work and the JS files are emitted. However, when a type error is introduced in the sub-project(s), the build fails and no JS files are emitted.

### Location of a Minimal Repository that Demonstrates the Issue.

https://github.com/emberjs/ember-debug/tree/d8ae4c9019f1fd673532c56575dc35898335d958 can be used to test this (by enabling `transpileOnly: true` which is not included in the repo yet)

appzuka:
I believe this behaviour is expected.  When you define a project reference, that sub-project is built in the same way as if you typed <code>tsc --build</code> in the sub-project's directory.  If there are errors they will be reported.  Files which were not part of the sub-project will be built by webpack calling ts-loader and if you include the transpileOnly option then the file will not be checked for type errors.

There is a need for documentation for project references as this is a relatively new feature.  This behaviour could be part of the documentation.

desmap:
just to get the terminology right, I also just tested this, check https://github.com/TypeStrong/ts-loader/pull/1173#issuecomment-681040638 and when...

#### CASE A: I have `transpileOnly: true` in `ts-loader` options + `fork-ts-checker` enabled

- No type errors if the type error is in a **non-reference file** (eg. a file that imports a reference but doesn't get imported itself)
- ✔Type errors if the type error is in a **reference file** (eg. a file that gets imported)

#### CASE B: I just have fork-ts-checker enabled:

- ✔Type errors if the type error is in a **non-reference file** (eg. a file that imports a reference but doesn't get imported itself)
- ✔Type errors if the type error is in a **reference file** (eg. a file that gets imported)

[...]

Issue #1157, opened 2020-07-19
Multiples re-compile when using webpack-dev-server with project references


noelebrun:
### Expected Behaviour

Build referenced projects, and then run a single Webpack compilation.

### Actual Behaviour

Multiples Webpack compilations are ran, my guess is that it watch referenced project and re-compile for each emitted file by the referenced project instead of waiting for the referenced project to be compiled first.

### Steps to Reproduce the Problem

- Clone this sample repo using project references: [monorepo-typescript-references](https://github.com/noelebrun/monorepo-typescript-references)
- Run `yarn` at the root
- Move current directory to `/packages/app`
- Run `yarn webpack-dev-server`
- Output is:
```
ℹ ｢wds｣: Project is running at http://localhost:8080/
ℹ ｢wds｣: webpack output is served from /
ℹ ｢wds｣: Content not from webpack is served from /Users/noelebrun/Documents/projects/monorepo-webpack/packages/app
ℹ ｢wdm｣: Compiled successfully.
ℹ ｢wdm｣: Compiling...
ℹ ｢wdm｣: Compiled successfully.
ℹ ｢wdm｣: Compiling...
ℹ ｢wdm｣: Compiled successfully.
ℹ ｢wdm｣: Compiling...
ℹ ｢wdm｣: Compiled successfully.
```
- Edit `/packages/utils/src/index.ts`
- Additional output is:
```
ℹ ｢wdm｣: Compiling...
ℹ ｢wdm｣: Compiled successfully.
ℹ ｢wdm｣: Compiling...
ℹ ｢wdm｣: Compiled successfully.
```

### Location of a Minimal Repository that Demonstrates the Issue.

Linked above: [monorepo-typescript-references](https://github.com/noelebrun/monorepo-typescript-references)

### Additional notes

As you can see from the steps to reproduce, when running `webpack-dev-server` for the first time, 4 compilations are ran. My first guess is that it's due to the `packages/utils` project to output 4 files when built (`index.js`, `index.d.ts`, `index.d.ts.map`, `index.js.map`): the watcher would trigger 4 separate compilations for each of those files.

When editing the file in the referenced project, two compilations are triggered (maybe because the `.d.ts` and the `.d.ts.map` are not re-build when not changing types so only 2 files are updated in the build folder?).

Note that, when running `webpack-dev-server` a second time when the referenced project is already built, only one compilation occurs.

johnnyreilly:
[...]

PR #1173, opened 2020-08-26
fix transpileOnly usage with fork-ts-checker

desmap:
docs are wrong: `transpileOnly: true` paired with fork-ts-checker together turns off type-checking entirely; just fork-ts-checker is sufficient

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,7 +292,7 @@ If you want to speed up compilation significantly you can set this flag.
 However, many of the benefits you get from static type checking between
 different dependencies in your application will be lost.
 
-It's advisable to use `transpileOnly` alongside the [fork-ts-checker-webpack-plugin](https://github.com/Realytics/fork-ts-checker-webpack-plugin) to get full type checking again. To see what this looks like in practice then either take a look at [our simple example](examples/fork-ts-checker-webpack-plugin). For a more complex setup take a look at our [more involved example](examples/react-babel-karma-gulp).
+Be aware that if you set `transpileOnly: true` together with the [fork-ts-checker-webpack-plugin](https://github.com/Realytics/fork-ts-checker-webpack-plugin) it will  disable  type checking with both ts-loader and with fork-ts-checker-webpack-plugin. To get latter doing the type checking but not former just omit `transpileOnly`.
 
 If you enable this option, webpack 4 will give you "export not found" warnings any time you re-export a type:
 


johnnyreilly (changes requested):
I don't understand your edit here I'm afraid; setting `transpileOnly: true` is a `ts-loader` option only and should have no effect upon the behaviour of the `fork-ts-checker-webpack-plugin`.

desmap:
hey @johnnyreilly, 

>  should have no effect upon the behaviour of the fork-ts-checker-webpack-plugin.

this is also what I don't get and in my prior apps I assumed exactly what you write.

But when I was benchmarking I found that type checking was turned off if you have both `transpileOnly: true` (tO) in the `ts-loader` `options` and `fork-ts-checker` (ftsc) as plugin enabled. I tested this multiple times with deliberate type errors. Also look at my [benchmarks ](https://github.com/TypeStrong/ts-loader/issues/1157#issuecomment-680674236)I at the two different versions: The tO + ftsc was as fast as the pure tO version and faster than the pure ftsc version. Test it yourself.

desmap:
just tested again and added a type error to one of the 1,000 functions:
```
(param: string) => {
  const extraError: number = param
  console.log(extraError)
}
```

#### `transpileOnly: true` in `ts-loader` options + `fork-ts-checker` enabled:
```
Entrypoint main = runtime~main.js main.js
[./client/index.tsx] 89.5 KiB {main} [built]
    + 47 hidden modules
ℹ ｢wdm｣: Compiled successfully.
No issues found.
```

#### just `fork-ts-checker` enabled:
```
Entrypoint main = runtime~main.js main.js
[./client/index.tsx] 89.5 KiB {main} [built] [1 error]
    + 47 hidden modules

ERROR in [...]/client/index.tsx
./client/index.tsx
[tsl] ERROR in [...]/client/index.tsx(4003,11)
      TS2322: Type 'string' is not assignable to type 'number'.
ℹ ｢wdm｣: Failed to compile.
```
CC @appzuka

johnnyreilly:
> Edit: i have `projectReference: true`

This seems significant. Do you want to repeat your test in a non-project references context. I suspect that's what's causing your issue.

desmap:
I did already in Edit2 and in this probably related thread: https://github.com/TypeStrong/ts-loader/issues/1158#issuecomment-681047669

desmap:
Ok then this pull is obsolete and it's something different, closing

PR #1175, opened 2020-08-30
Update docs to show transpileOnly does not affect project references

appzuka:
Make it clear in the docs that transpileOnly does not affect the way project references are built.


johnnyreilly:
Thanks @appzuka!  I've tweaked the formatting slightly.  Nice contribution ❤️

PR #1176, opened 2020-08-31
Shares solution builder between instances

sheetalkamat:
I am not sure if this right change or not as i dont understand much about config and need for different instances but this came up as part of investigation where multiple instances of solution builder were created because different instances were used. The project in question takes about 2-3 minutes to just check if the solution is upto date adding to perf and memory implications.

I am not sure if we need this PR or people just fix their config file to use same instance instead. (which is what investigation resulted in)

### DIFF
--- a/src/after-compile.ts
+++ b/src/after-compile.ts
@@ -6,6 +6,7 @@ import * as constants from './constants';
 import { getEmitFromWatchHost, getEmitOutput } from './instances';
 import {
   FilePathKey,
+  LoaderOptions,
   TSFiles,
   TSInstance,
   WebpackModule,
@@ -17,6 +18,7 @@ import {
   formatErrors,
   isReferencedFile,
   populateReverseDependencyGraph,
+  tsLoaderSource,
 } from './utils';
 
 export function makeAfterCompile(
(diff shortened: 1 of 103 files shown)


johnnyreilly:
Thanks for both the PRs @sheetalkamat! :sunflower:  

I've looked at both and each seems like a legitimate approach...  To your question:

> i dont understand much about config and need for different instances

I've never found a need for using different instances myself either; it predates my involvement with `ts-loader` and I've never removed it.  Every now and then someone has contributed tweaks to it (in fact the last release https://github.com/TypeStrong/ts-loader/pull/1170 was a tweak to `instance` functionality)  so it's a feature that's used; just not by me :smile: 

Given that it's a feature that's not likely to go away I'm tempted to go with one of the PRs you're suggesting.  What are the pros 
and cons of each of the approaches you've suggested?  They don't seems wildly dissimilar...

sheetalkamat:
This one just shares solution builder and uses program/language service as if its own.. the other one just does not create new instance so program/LS are same across instances in that other PR.

johnnyreilly:
> This one just shares solution builder and uses program/language service as if its own.. 

Okay cool. I don't feel strongly about the different approaches; each sounds good! Without having tested it, I'd guess that https://github.com/TypeStrong/ts-loader/pull/1177 may perform slightly better as it's not creating new instances. I haven't tested it and it may be a marginal difference, but maybe https://github.com/TypeStrong/ts-loader/pull/1177 is the way to go.

Do you fancy incrementing the version in the `package.json` and adding an entry to the `CHANGELOG.md` on https://github.com/TypeStrong/ts-loader/pull/1177 ?

johnnyreilly:
Closing in favour of #1177 which shipped with https://github.com/TypeStrong/ts-loader/releases/tag/8.0.4

PR #1177, opened 2020-08-31
Uses existing instance if config file is same as already built solution

sheetalkamat:
This is another take on #1176 again not sure if this is right approach.


johnnyreilly:
Copied over from #1176

> > This one just shares solution builder and uses program/language service as if its own..
> 
> Okay cool. I don't feel strongly about the different approaches; each sounds good! Without having tested it, I'd guess that #1177 may perform slightly better as it's not creating new instances. I haven't tested it and it may be a marginal difference, but maybe #1177 is the way to go.

Do you fancy incrementing the version in the package.json and adding an entry to the CHANGELOG.md on this PR ?

PR #1178, opened 2020-09-02
Add ignoreEmptyOutput config flag for js-only projects

SahAssar:
I'd like to add a config flag to allow empty output to allow using ts-loader with the noEmit option. This is because I use ts-loader in a js-only project where it is used to add typechecking via the checkJs and allowJs flags in combination with type info in jsdoc. Even though ts-loader in this setup effectively does nothing I like having the typechecking within the webpack pipeline.

If the proposed ignoreEmptyOutput is true then the callback is simply passed the input instead of calculating the output.

Please let me know if there are other/better ways to achive this!

### DIFF
--- a/README.md
+++ b/README.md
@@ -626,6 +626,14 @@ loading only those files that are actually bundled by webpack, as well as any `.
 by the `tsconfig.json` settings. `.d.ts` files are still included because they may be needed for
 compilation without being explicitly imported, and therefore not picked up by webpack.
 
+#### ignoreEmptyOutput
+| Type | Default Value |
+|------|--------------|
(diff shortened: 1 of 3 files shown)


johnnyreilly:
hey @SahAssar! 

thanks for the pull request!  Actually I think there is probably a better way than what you suggest; using `ts-loader`'s sister project: https://github.com/TypeStrong/fork-ts-checker-webpack-plugin

It does type checking for a webpack project and just that.  I think that's probably exactly what you're after?

cc @piotr-oles

SahAssar:
Thanks, that sounds better!

PR #1184, opened 2020-09-09
Add documentation and example code for projectReferences

appzuka:
I have added a new document, REFERENCES.md and linked to it from the main README.md.  I have also added an example to the examples/ folder.

Please feel free to make any changes to the wording/formatting/structure as appropriate.


johnnyreilly:
This is fantastic work! I need to give it a proper read but thanks so much already ❤️🌻

REFERENCES.md
```diff
@@ -0,0 +1,329 @@
+
+# Using TypeScript Project References with ts-loader and Webpack
```
johnnyreilly, line 2: Lowercase w for webpack please; surprising though it may be, that's the official casing

REFERENCES.md
```diff
@@ -0,0 +1,329 @@
...
+[https://github.com/appzuka/project-references-example](https://github.com/appzuka/project-references-example)
+
+There are 2 stages to using project references in your project:
+
+1. Configure and build the project references
+1. Setup your codebase to consume the compiled projects
+
+### Configure and build the project references
+
+This stage just involves following the directions from the TypeScript documentation:
```
johnnyreilly, line 23: As I understand it from the "Using ts-loader to build project references" section, following the approach prior to that is an alternative approach that could be used and also probably *not* the one people will likely go for. `ts-loader` building references is just going to be simpler (my presumption)

As such, is it worth calling out here that these docs are going to outline 2 approaches more explicitly?

johnnyreilly (commented):
Left some comments - this is great stuff BTW!

REFERENCES.md
appzuka: I'll make this change throughout the document.

REFERENCES.md
appzuka: I wanted to show how to configure using tsc --build because you need to get that part right or nothing is going to work.  It is also helpful to check that it is working with tsc independently if you are having problems.  Once it is working I believe most people would use ts-loader.

I'll amend the docs to make it clear I am not proposing 2 alternatives.

johnnyreilly:
This is awesome - thanks @appzuka!

Issue #1194, opened 2020-10-11
Webpack 5: deprecation warning [DEP_WEBPACK_MODULE_ERRORS] 


sanex3339:
### Expected Behaviour
No deprecation warning

### Actual Behaviour
```
(node:69430) [DEP_WEBPACK_MODULE_ERRORS] DeprecationWarning: Module.errors was removed (use getErrors instead)
```

### Steps to Reproduce the Problem
PR with the error:
https://github.com/javascript-obfuscator/javascript-obfuscator/pull/776

Full trace:
```
(node:69499) [DEP_WEBPACK_MODULE_ERRORS] DeprecationWarning: Module.errors was removed (use getErrors instead)
    at Object.reportTranspileErrors (/Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/ts-loader/dist/instances.js:263:24)
    at successLoader (/Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/ts-loader/dist/index.js:28:17)
    at Object.loader (/Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/ts-loader/dist/index.js:24:5)
    at LOADER_EXECUTION (/Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/loader-runner/lib/LoaderRunner.js:132:14)
    at runSyncOrAsync (/Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/loader-runner/lib/LoaderRunner.js:133:4)
    at iterateNormalLoaders (/Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/loader-runner/lib/LoaderRunner.js:251:2)
    at Array.<anonymous> (/Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/loader-runner/lib/LoaderRunner.js:224:4)
    at runCallbacks (/Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/webpack/node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:24:49)
    at /Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/webpack/node_modules/enhanced-resolve/lib/CachedInputFileSystem.js:194:4
    at /Users/sanex/WebstormProjects/javascript-obfuscator/node_modules/webpack/node_modules/graceful-fs/graceful-fs.js:123:16
```

### Location of a Minimal Repository that Demonstrates the Issue.
[...]

PR #1195, opened 2020-10-12
Fixed deprecation warnings on webpack@5.

sanex3339:
Fixed https://github.com/TypeStrong/ts-loader/issues/1194

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -653,7 +653,7 @@ function getTranspilationEmit(
      * Since webpack 5, the `errors` property is deprecated,
      * so we can check if some methods for reporting errors exist.
      */
-    if (!!module.addError) {
+    if (module.addError) {
       errors.forEach(error => module.addError(error));
     } else {
       module.errors.push(...errors);
--- a/src/index.ts
+++ b/src/index.ts
@@ -632,6 +632,7 @@ function getTranspilationEmit(
     reportDiagnostics: true,
     fileName,
   });
+  const module = loaderContext._module;
 
   addDependenciesFromSolutionBuilder(instance, fileName, file =>
     loaderContext.addDependency(path.resolve(file))
@@ -644,11 +645,19 @@ function getTranspilationEmit(
       instance.loaderOptions,
       instance.colors,
       instance.compiler,
-      { module: loaderContext._module },
+      { module },
       loaderContext.context
     );
 
-    loaderContext._module.errors.push(...errors);
+    /**
+     * Since webpack 5, the `errors` property is deprecated,
+     * so we can check if some methods for reporting errors exist.
+     */
+    if (!!module.addError) {
+      errors.forEach(error => module.addError(error));
+    } else {
+      module.errors.push(...errors);
+    }
   }
 
   return { outputText, sourceMapText };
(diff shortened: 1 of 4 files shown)


sanex3339:
Don't sure how to test it, especially because previous checks for webpack@5 were without tests.

src/index.ts
```diff
@@ -644,11 +645,19 @@ function getTranspilationEmit(
...
+      { module },
       loaderContext.context
     );
 
-    loaderContext._module.errors.push(...errors);
+    /**
+     * Since webpack 5, the `errors` property is deprecated,
+     * so we can check if some methods for reporting errors exist.
+     */
+    if (!!module.addError) {
```
johnnyreilly, line 656: Do we need the `!!` here?

```suggestion
    if (module.addError) {
```

src/instances.ts
```diff
@@ -165,7 +166,15 @@ function successfulTypeScriptInstance(
       loader.context
     );
 
-    loader._module.errors.push(...errors);
+    /**
+     * Since webpack 5, the `errors` property is deprecated,
+     * so we can check if some methods for reporting errors exist.
+     */
+    if (!!module.addError) {
```
johnnyreilly, line 173: Do we need the `!!` here?

```suggestion
    if (module.addError) {
```

src/instances.ts
```diff
@@ -431,7 +441,15 @@ export function reportTranspileErrors(
       { file: instance.configFilePath || 'tsconfig.json' },
       loader.context
     );
-    loader._module.errors.push(...solutionErrors, ...errors);
+    /**
+     * Since webpack 5, the `errors` property is deprecated,
+     * so we can check if some methods for reporting errors exist.
+     */
+    if (!!module.addError) {
```
johnnyreilly, line 448: Do we need the `!!` here?

```suggestion
    if (module.addError) {
```

johnnyreilly:
Looks good!  I've left some comments / suggestions above.  What do you think?

src/instances.ts
sanex3339: Fixed in https://github.com/TypeStrong/ts-loader/pull/1195/commits/f0a81318d255c18f11027e043d3bd56b13460bad

src/instances.ts
sanex3339: Fixed in https://github.com/TypeStrong/ts-loader/pull/1195/commits/f0a81318d255c18f11027e043d3bd56b13460bad

src/index.ts
sanex3339: Fixed in https://github.com/TypeStrong/ts-loader/pull/1195/commits/f0a81318d255c18f11027e043d3bd56b13460bad

sanex3339:
> Looks good!  I've left some comments / suggestions above. (Essentially the same point three times :smile: )  What do you think?

Fixed

johnnyreilly:
Could you update the version in the `package.json` to 8.0.5 and add an entry to the `package.json` please? If the tests pass we're likely safe to ship.

sanex3339:
> Could you update the version in the `package.json` to 8.0.5 and add an entry to the `package.json` please? If the tests pass we're likely safe to ship.

https://github.com/TypeStrong/ts-loader/pull/1195/commits/57d3fe946692a9904ac19e317904eab8c9d24521

Issue #1196, opened 2020-10-14
Still got DeprecationWarning on Webpack 5


Vincent0700:
I'm already update deps to latest but still get the deprecation warinning. Does the problem need to be fixed? Or just add the flag "--no-deprecation" to disable it.

environment:
```
"ts-loader": "^8.0.5",
"ts-node": "^9.0.0",
"webpack": "^5.1.0"
```

commands:
```bash
$ node --trace-deprecation node_modules/webpack/bin/webpack.js --config=build/webpack.prod.config.ts
(node:12757) [DEP_WEBPACK_COMPILATION_ASSETS] DeprecationWarning: Compilation.assets will be frozen in future, all modifications are deprecated.
BREAKING CHANGE: No more changes should happen to Compilation.assets after sealing the Compilation.
        Do changes to assets earlier, e. g. in Compilation.hooks.processAssets.
        Make sure to select an appropriate stage from Compilation.PROCESS_ASSETS_STAGE_*.
    at outputFileToAsset (/Users/Vincent/Playground/config-table/node_modules/ts-loader/dist/after-compile.js:216:35)
    at outputFilesToAsset (/Users/Vincent/Playground/config-table/node_modules/ts-loader/dist/after-compile.js:224:13)
    at addDeclarationFilesAsAsset (/Users/Vincent/Playground/config-table/node_modules/ts-loader/dist/after-compile.js:210:5)
    at provideDeclarationFilesToWebpack (/Users/Vincent/Playground/config-table/node_modules/ts-loader/dist/after-compile.js:206:9)
    at /Users/Vincent/Playground/config-table/node_modules/ts-loader/dist/after-compile.js:30:9
    at Hook.eval [as callAsync] (eval at create (/Users/Vincent/Playground/config-table/node_modules/webpack/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:5:1)
    at Hook.CALL_ASYNC_DELEGATE [as _callAsync] (/Users/Vincent/Playground/config-table/node_modules/webpack/node_modules/tapable/lib/Hook.js:18:14)
    at /Users/Vincent/Playground/config-table/node_modules/webpack/lib/Compiler.js:986:33
    at /Users/Vincent/Playground/config-table/node_modules/webpack/lib/Compilation.js:2025:10
[...]

PR #1200, opened 2020-10-19
Replace afterCompile to stop webpack 5 warning

appzuka:
This fixes the deprecation warning in webpack 5 as discussed in issue #[1196](https://github.com/TypeStrong/ts-loader/issues/1196). Writing assets such as declarations is moved to the afterProcessAssets hook of the compilation from the afterCompile hook of the compiler.

As the afterProcessAssets hook does not exist in webpack 4 to make ts-loader work for both webpack 4 and 5 it is necessary to execute different code depending on whether afterProcessAssets exists.  This means that the new code is not tested at all by either the execution or comparison tests, which feels wrong.  I'm not sure how to fix this until tests using webpack 5 are introduced, which feels like a major task.

I have tested this build with a small project using webpack 4 and webpack 5.  If I introduce a deliberate error into the webpack 5 branch of the new code all the tests still pass although building the project with webpack 5 fails.

### DIFF
--- a/src/instances.ts
+++ b/src/instances.ts
@@ -341,10 +341,26 @@ export function initializeInstance(
     instance.transformers = getCustomTransformers(program);
     // Setup watch run for solution building
     if (instance.solutionBuilderHost) {
-      loader._compiler.hooks.afterCompile.tapAsync(
-        'ts-loader',
-        makeAfterCompile(instance, instance.configFilePath)
-      );
+      if (loader._compilation.hooks.afterProcessAssets) {
+        // afterProcessAssets does not exist in webpack4
+        loader._compilation.hooks.afterProcessAssets.tap(
+          'ts-loader',
+          (_: any) => {
+            makeAfterCompile(instance, instance.configFilePath)(
+              loader._compilation,
+              () => {
+                return null;
+              }
+            );
+          }
+        );
+      } else {
+        // adding assets in afterCompile is depracated in webpack 5
+        loader._compiler.hooks.afterCompile.tapAsync(
+          'ts-loader',
+          makeAfterCompile(instance, instance.configFilePath)
+        );
+      }
       loader._compiler.hooks.watchRun.tapAsync(
         'ts-loader',
         makeWatchRun(instance, loader)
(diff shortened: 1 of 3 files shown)


src/instances.ts
```diff
@@ -341,10 +341,26 @@ export function initializeInstance(
...
+            makeAfterCompile(instance, instance.configFilePath)(
+              loader._compilation,
+              () => {
+                return null;
+              }
+            );
+          }
+        );
+      } else {
+        // adding assets in afterCompile is depracated in webpack 5
```
johnnyreilly, line 358:
```suggestion
        // adding assets in afterCompile is deprecated in webpack 5
```

src/instances.ts
```diff
@@ -391,11 +407,27 @@ export function initializeInstance(
...
+          makeAfterCompile(instance, instance.configFilePath)(
+            loader._compilation,
+            () => {
+              return null;
+            }
+          );
+        }
+      );
+    } else {
+      // adding assets in afterCompile is depracated in webpack 5
```
johnnyreilly, line 424:
```suggestion
      // adding assets in afterCompile is depracated in webpack 5
```

src/instances.ts
```diff
@@ -391,11 +407,27 @@ export function initializeInstance(
         instance.languageService!.getProgram()
       );
     }
+    if (loader._compilation.hooks.afterProcessAssets) {
+      // afterProcessAssets does not exist in webpack4
+      loader._compilation.hooks.afterProcessAssets.tap(
+        'ts-loader',
+        (_: any) => {
```
johnnyreilly, line 414: Do you happen to know what the type is here? Not that we're using it but I'm curious

johnnyreilly:
This looks great @appzuka ! Thanks for your work 🌻❤️

Would you like to update the `package.json` and the `CHANGELOG.md` and we can ship a release?

src/instances.ts
appzuka: It is a list of the assets:

https://github.com/webpack/webpack/blob/9342540a1f11b84cd693b186ec4e72d37f5a0aeb/lib/Compilation.js#L299

If I include the parameter there is a warning that it is unused, so I decided to include it as <code>_:any</code>

PR #1202, opened 2020-10-23
Speeds up project reference build and doesnt store the result in memory

sheetalkamat:
Given the project references output files are no more in memory but have to be written to disk in order for webpack to handle symlinks, we dont need in memory read and cache of input/output files from project references. They are on disk and handled it that way.

This speeds up upto-date check to behave just like what `tsc -b` would do so that we are not reading files in memory.
This was done as part of investigation for a team who was running into OOM and slower builds on huge project.


johnnyreilly:
This is amazing work @sheetalkamat!  Thank you!

I've added this to the `CHANGELOG.md` and updated the version in the `package.json`.  (I also change the casing of a letter by 1 character to satisfy my OCD tendencies :smile: )

I'll plan to publish this later today assuming the build plays nice.  Thank you so much!

berickson1:
This has actually resulted in a huge perf regression in our builds. Now all incremental builds are as slow as clean builds(~3 minutes). Is there any debugging logs I can gather - I don't really want to blindly submit a non-actionable bug since this doesn't repo in a small test repo.

Issue #1203, opened 2020-10-31
Memory leak when running with multiple webpack (compiler) instances


valerio:
### Expected Behaviour
When creating and disposing multiple webpack instances using `ts-loader`, compiler instances should not be retained.
 
### Actual Behaviour
When running a build with `ts-loader`, webpack compiler instances are never removed by the GC.

### Steps to Reproduce the Problem

1. create a minimal `ts-loader` configuration [like the one from the readme](https://github.com/TypeStrong/ts-loader#configuration)
2. create a webpack compiler and call `run()` on it without storing any references
3. repeat step 2 enough times and the node process will run out of memory

### Location of a Minimal Repository that Demonstrates the Issue.

https://github.com/valerio/ts-loader-leak

### Additional details

I have an application that runs webpack in-memory to transpile some TypeScript code.
The webpack compilers are disposed after some time, but I noticed that references to them were still being held by `ts-loader`.

I've spent some time debugging this and I saw that these references are held by the `webpackInstances` array:
<img width="824" alt="Screenshot 2020-10-31 at 13 42 33" src="https://user-images.githubusercontent.com/39825307/97785618-62bce600-1ba6-11eb-843a-a86b9108f778.png">

The array is declared here: https://github.com/TypeStrong/ts-loader/blob/master/src/index.ts#L30
This is a global variable and references are only ever added to it, never removed, as far as I can see.

The array is used to determine a key for an instance cache, which is using a `WeakMap`: https://github.com/TypeStrong/ts-loader/blob/master/src/index.ts#L189-L196

I think this array could be replaced with another `WeakMap` itself (or use `WeakRef`) so it doesn't retain compiler instances.

johnnyreilly:
Thanks for digging into that! If you'd like to experiment with trying out the `WeakMap` / `WeakRef` approach we could potentially be interested.

valerio:
> Thanks for digging into that! If you'd like to experiment with trying out the `WeakMap` / `WeakRef` approach we could potentially be interested.

Yep, I'll give it a try and submit a PR if that fixes it. 👍

PR #1205, opened 2020-11-03
Fix memory leak when using multiple webpack instances

valerio:
Closes https://github.com/TypeStrong/ts-loader/issues/1203

This PR should fix a memory leak issue when using multiple webpack instances that get discarded over time.

I initially converted the `webpackInstances` array to a WeakMap (plus a global index) but that wasn't enough to fix the leak, since the cached `TSInstance` held in `instances.ts` was still retaining the webpack compiler and never being freed.

In the end, I removed the webpack instance array and changed the TS instance cache by using a combined WeakMap + Map so that we maintain the same behaviour but keep the lifetime of `TSInstance`s tied to the webpack compiler that spawned them.

There is one small change in how `loaderOptions.instance` names are generated: 
- old names: `0_default_e3b0c44298fc1c14`
- new names: `default_e3b0c44298fc1c14`

They now don't contain the unique index for the webpack compiler that generated them, this shouldn't change the way instances are cached, since they are now tied to a specific webpack instance by reference (the key of the WeakMap) instead of using the index.

I've tested that the leak is fixed using this repo: https://github.com/valerio/ts-loader-leak
Memory now gets periodically freed between runs:
```
➜  ts-loader-leak git:(main) node index.js 1
webpack will run, process should run out of memory
current heap usage { heapMBs: 101 }
current heap usage { heapMBs: 151 }
// omitted for brevity
current heap usage { heapMBs: 882 }
current heap usage { heapMBs: 937 }
current heap usage { heapMBs: 991 }
current heap usage { heapMBs: 98 } <-- memory was freed 
current heap usage { heapMBs: 154 }
current heap usage { heapMBs: 208 }
current heap usage { heapMBs: 257 }
current heap usage { heapMBs: 312 }
current heap usage { heapMBs: 366 }
current heap usage { heapMBs: 155 } <-- memory was freed 
current heap usage { heapMBs: 202 }
current heap usage { heapMBs: 257 }
current heap usage { heapMBs: 312 }
current heap usage { heapMBs: 359 }
```

I thought about adding a test case in here but I don't think it would fit with the current setup, it also might be non-deterministic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## v8.0.8
+* [Fixed memory leak when using multiple webpack instances](https://github.com/TypeStrong/ts-loader/pull/1205) - thanks @valerio
+
 ## v8.0.7
 * [Speeds up project reference build and doesnt store the result in memory](https://github.com/TypeStrong/ts-loader/pull/1202) - thanks @sheetalkamat
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "8.0.7",
+  "version": "8.0.8",
   "description": "TypeScript loader for webpack",
   "main": "index.js",
   "types": "dist",
--- a/src/index.ts
+++ b/src/index.ts
@@ -27,11 +27,6 @@ import {
   isReferencedFile,
 } from './utils';
 
-// WeakMaps are not enumerable, so we need to keep track of how many webpack instances
-// were used, this index should be incremented whenever a new instance is used.
-let lastWebpackIndex = 0;
-// Using a WeakMap prevents compiler instances from being kept in memory forever.
-const webpackIndexMap: WeakMap<webpack.Compiler, number> = new WeakMap();
 const loaderOptionsCache: LoaderOptionsCache = {};
 
 /**
@@ -178,14 +173,6 @@ function getOptionsHash(loaderOptions: LoaderOptions) {
  * or creates them, adds them to the cache and returns
  */
 function getLoaderOptions(loaderContext: webpack.loader.LoaderContext) {
-  // differentiate the TypeScript instance based on the webpack instance
-  let webpackIndex = webpackIndexMap.get(loaderContext._compiler);
-  if (!webpackIndex) {
-    webpackIndex = lastWebpackIndex;
-    webpackIndexMap.set(loaderContext._compiler, lastWebpackIndex);
-    lastWebpackIndex++;
-  }
-
   const loaderOptions =
     loaderUtils.getOptions<LoaderOptions>(loaderContext) ||
     ({} as LoaderOptions);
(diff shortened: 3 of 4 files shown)


johnnyreilly:
This is fantastic work!  Do you want to update the `package.json` to a new version and add an entry to the `CHANGELOG.md`?  Let's ship this!

valerio:
> This is fantastic work! Do you want to update the `package.json` to a new version and add an entry to the `CHANGELOG.md`? Let's ship this!

Done and done! 🚀 
And thanks for the speedy review! 😄

liorcode:
Seems this change breaks my build:
```
        Module build failed (from ../node_modules/sled-test-runner/node_modules/thread-loader/dist/cjs.js):
        Thread Loader (Worker 2)
        Invalid value used as weak map key
            at PoolWorker.fromErrorObj (node_modules/sled-test-runner/node_modules/thread-loader/dist/WorkerPool.js:262:12)
            at node_modules/sled-test-runner/node_modules/thread-loader/dist/WorkerPool.js:204:29
            at WeakMap.set (<anonymous>)
            at Object.getTypeScriptInstance (node_modules/ts-loader/dist/instances.js:37:23)
            at Object.loader (node_modules/ts-loader/dist/index.js:16:41)  
```
Had to pin the ts-loader version to 8.0.7 to work around it

johnnyreilly:
Oh no! @valerio would you be able to take a look please?

@liorcode is it ThreadLoader being in the mix that causes the issue?

liorcode:
I am actually using a 3rd party service which itself uses thread loader, so unfortunately I can't control it, but thanks!

valerio:
> ```
>         Module build failed (from ../node_modules/sled-test-runner/node_modules/thread-loader/dist/cjs.js):
>         Thread Loader (Worker 2)
>         Invalid value used as weak map key
>             at PoolWorker.fromErrorObj (node_modules/sled-test-runner/node_modules/thread-loader/dist/WorkerPool.js:262:12)
>             at node_modules/sled-test-runner/node_modules/thread-loader/dist/WorkerPool.js:204:29
>             at WeakMap.set (<anonymous>)
>             at Object.getTypeScriptInstance (node_modules/ts-loader/dist/instances.js:37:23)
>             at Object.loader (node_modules/ts-loader/dist/index.js:16:41)  
> ```

That's a strange issue indeed, some values like `Symbol` or `string` can't be used as a weak map key, maybe ThreadLoader is hooking into webpack and replacing the compiler instance with something else?

I'll see if I can reproduce, @liorcode a reproduction repo would definitely help if you have an easy way to set that up!

Krasnopir:
Have same problems after update.
```

ERROR in ./apps/admin-client/src/index.tsx
Module build failed (from ./node_modules/thread-loader/dist/cjs.js):
Thread Loader (Worker 0)

Invalid value used as weak map key
at PoolWorker.fromErrorObj (/usr/src/app/node_modules/thread-loader/dist/WorkerPool.js:344:12)
at WeakMap.set (<anonymous>)
at Object.getTypeScriptInstance (/usr/src/app/node_modules/ts-loader/dist/instances.js:37:23)
at Object.loader (/usr/src/app/node_modules/ts-loader/dist/index.js:16:41)
```

---------
my webpack config 
```

        {
          test: /\.tsx?$/,
          exclude: [/node_modules/],
          use: [
            { loader: 'cache-loader' },
            {
              loader: 'thread-loader',
              options: {
                workers: require('os').cpus().length - 1,
                poolTimeout: isProduction ? 500 : Infinity,
              },
            },
            {
              loader: 'ts-loader',
              options: {
                happyPackMode: true, 
              },
            },
          ],
        },
```

valerio:
I've managed to reproduce this, opened an issue here: https://github.com/TypeStrong/ts-loader/issues/1206

I'll see if I can come up with a fix for this as well!

Issue #1206, opened 2020-11-04
Builds fail when using thread-loader


valerio:
### Expected Behaviour

Using `thread-loader` before `ts-loader`, compilation of a project should succeed.

### Actual Behaviour

Compilation fails with the following error:
```
[webpack-cli] Compilation finished
assets by status 655 bytes [cached] 1 asset
./src/app.ts 39 bytes [built] [code generated] [1 error]

ERROR in ./src/app.ts
Module build failed (from ./node_modules/thread-loader/dist/cjs.js):
Thread Loader (Worker 0)
Invalid value used as weak map key
    at PoolWorker.fromErrorObj (/Users/valeriopipolo/repos/tsloaderleak/node_modules/thread-loader/dist/WorkerPool.js:344:12)
    at /Users/valeriopipolo/repos/tsloaderleak/node_modules/thread-loader/dist/WorkerPool.js:217:29
    at WeakMap.set (<anonymous>)
    at Object.getTypeScriptInstance (/Users/valeriopipolo/repos/tsloaderleak/node_modules/ts-loader/dist/instances.js:37:23)
    at Object.loader (/Users/valeriopipolo/repos/tsloaderleak/node_modules/ts-loader/dist/index.js:16:41)

webpack 5.3.2 compiled with 1 error in 441 ms
```

### Steps to Reproduce the Problem

Use a minimal webpack configuration as follows:
```js
module.exports = {
  devtool: "inline-source-map",
  entry: "./src/app.ts",
  output: {
    filename: "bundle.js",
  },
  module: {
    rules: [
      {
        test: /\.tsx?$/,
        use: ["thread-loader", "ts-loader"],
      },
    ],
  },
};
```

Running builds with `webpack` should fail.
Reproducible with the following dependencies:
```
    "thread-loader": "3.0.1",
    "ts-loader": "8.0.8",
    "typescript": "4.0.5",
    "webpack": "4.13.0"
```

### Location of a Minimal Repository that Demonstrates the Issue.

https://github.com/valerio/tsloader-build-issue-with-thread-loader

valerio:
When using `thread-loader` the `loader._compiler` instance is set to undefined, which makes setting it as a key to a `WeakMap` fail at runtime: https://github.com/TypeStrong/ts-loader/blob/master/src/instances.ts#L63

A couple ideas for workarounds:
- use a "marker" object to use as a key in the `WeakMap` whenever a webpack compiler isn't available passed
[...]

PR #1207, opened 2020-11-04
Fix failed builds when using thread-loader

valerio:
Possible fix for https://github.com/TypeStrong/ts-loader/issues/1206

When using `thread-loader`, the `loader._compiler` field is `undefined`, this causes a runtime error when trying to use it as a key in a WeakMap.

This PR changes the instance cache so that it now uses a dummy "marker" object as reference when the compiler instance is undefined.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## v8.0.9
+* [Fixed build failing when using thread-loader](https://github.com/TypeStrong/ts-loader/pull/1207) - thanks @valerio
+
 ## v8.0.8
 * [Fixed memory leak when using multiple webpack instances](https://github.com/TypeStrong/ts-loader/pull/1205) - thanks @valerio
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "8.0.8",
+  "version": "8.0.9",
   "description": "TypeScript loader for webpack",
   "main": "index.js",
   "types": "dist",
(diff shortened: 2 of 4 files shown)


johnnyreilly (commented):
This looks great! Thank you!

Now I must reveal a prejudice of mine... I don't really like `class`es. It's not an accident that `ts-loader` is pretty much a `class` free codebase.

From the looks of it, `class InstanceCache` is a `class` that is instantiated just the once. I wonder if it would be straightforward to switch over to the revealing module pattern here perhaps?

No worries if you don't have time now, this looks functionally completely correct.

valerio:
> From the looks of it, `class InstanceCache` is a `class` that is instantiated just the once. I wonder if it would be straightforward to switch over to the revealing module pattern here perhaps?

Fair point! Switching to that should be fairly easy, I'll make the change 👌

johnnyreilly:
Awesome - you want to include the `package.json` and `CHANGELOG.md` changes as well? All good and we'll ship!

valerio:
> Awesome - you want to include the `package.json` and `CHANGELOG.md` changes as well? All good and we'll ship!

Done!

I'll also take a look at the other error that I saw when using `thread-loader` and open an issue for it (unless there's something related already).

valerio:
@johnnyreilly I've unlinked this PR from the original issue (https://github.com/TypeStrong/ts-loader/issues/1206), so that it can be closed after users report that it's fixed, but feel free to close it if you prefer handling it that way.

johnnyreilly:
Travis is stuck - going to override and merge.

johnnyreilly:
v8.0.9 should be out in the next 20 mins or so - thanks for the quick turnaround @valerio!

https://github.com/TypeStrong/ts-loader/releases/tag/v8.0.9

PR #1217, opened 2020-12-10
Instead of checking date, check time thats more accurate to see if something has changed

sheetalkamat:
This ensures that there is no unnecesary watcher invoked resulting in full config file reloads and builds

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 # Changelog
 
+## v8.0.12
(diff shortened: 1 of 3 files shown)


johnnyreilly:
nice work @sheetalkamat!  I've updated the `CHANGELOG.md` and `package.json` - once the build passes lets look to get this shipped!

johnnyreilly:
Shipping with https://github.com/TypeStrong/ts-loader/releases/tag/v8.0.12

PR #1223, opened 2020-12-22
Remove @types/webpack due to Webpack 5 update

nmggithub:
According to the [update notes](https://webpack.js.org/blog/2020-10-10-webpack-5-release/#typescript-typings), the package `@types/webpack` is no longer used. It should be removed from `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,6 @@
     "@types/micromatch": "^3.1.0",
     "@types/node": "*",
     "@types/semver": "^6.0.0",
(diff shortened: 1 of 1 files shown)


johnnyreilly:
`ts-loader` is still built with webpack 4. Until we move to build with webpack 5 we can't make this change.

nmggithub:
> `ts-loader` is still built with webpack 4. Until we move to build with webpack 5 we can't make this change.

Ok. Looking forward to it.

Issue #1215, opened 2020-12-01
Incremental Compile Performance Regression 8.0.3 -> 8.0.11


berickson1:
Note: This is in a relatively large repo that uses project references + composite projets
~55 typescript composite packages (and slowly growing)
Webpack + ts-loader for compilation (running in watch mode)

### Expected Behaviour
Incremental compile is a similar duration after upgrade of ts-loader (~11 seconds)

### Actual Behaviour
Incremental compile is 20x longer (~240 seconds)

### Steps to Reproduce the Problem
Unfortunately I'm not able to share the repo - but a simple drop-in replacement of ts-loader caused a huge spike in incremental compile times.  I'd love to provide any relevant debugging output to help diagnose this.

I think I can work around it by using fork-ts-checker-webpack-plugin, but ideally the root cause in ts-loader could be identified and fixed.

johnnyreilly:
Could you identify specifically which version of ts-loader the regression occurred in please?

berickson1:
I grabbed the webpack build times and recorded them below, the clock time on cold builds is actually higher than what webpack reports.
I performed the following test by installing the specified version of ts-loader, executing the webpack build in watch mode, recording first compile time, then modifying 1 file (same file every time) and recording time again

8.0.3  - cold build:   34845ms incremental:    14388ms
8.0.4  - cold build:   32921ms incremental:     13763ms
8.0.5  - cold build:   32908ms incremental:    13753ms
8.0.6  - cold build:   32030ms incremental:    13724ms
8.0.7  - cold build:  135885ms incremental:  176851ms
8.0.8  - cold build:  145522ms incremental:  157949ms
8.0.9  - cold build:  144421ms incremental:  153780ms
8.0.10 - cold build: 143325ms incremental:  154577ms
8.0.11 - cold build:  143119ms  incremental:  151772ms

Note: Sample size of 1 build per version, but there's an obvious step function between 8.0.6 and 8.0.7, both for cold and incremental builds

https://github.com/TypeStrong/ts-loader/pull/1202 seems to be the culprit. Incremental builds are effectively the same cost as cold builds, which is pretty much a dealbreaker for upgrading atm. I'd love to help diagnose however possible.

By comparison, executing `tsc -b` on a fresh repo takes ~262s and a change when running `tsc -b -w` takes ~7s.

johnnyreilly:
[...]

PR #1226, opened 2020-12-29
Increase build performance by skipping node_modules file lookups

berickson1:
Skip attempting to lookup node_modules output files names inside getInputFileNameFromOutput. This should never resolve if the allowTsInNodeModule loader option is not set.

Tested in a closed-source project & showed a large improvement in speed over 8.0.12 (~20% decrease in build times)

Relates to #1215

### DIFF
--- a/src/servicesHost.ts
+++ b/src/servicesHost.ts
@@ -852,6 +852,14 @@ export function makeSolutionBuilderHost(
   function getInputFileNameFromOutput(
     outputFileName: string
   ): string | true | undefined {
+    // Unless we explicitly want to compile files in node_modules, exclude them from lookups
+    if (
+      !instance.loaderOptions.allowTsInNodeModules &&
+      outputFileName.indexOf('node_modules') !== -1
+    ) {
+      return undefined;
+    }
+
     const resolvedFileName = filePathKeyMapper(outputFileName);
     for (const configInfo of configFileInfo.values()) {
       ensureInputOutputInfo(configInfo);
--- a/src/servicesHost.ts
+++ b/src/servicesHost.ts
@@ -855,7 +855,7 @@ export function makeSolutionBuilderHost(
     // Unless we explicitly want to compile files in node_modules, exclude them from lookups
     if (
       !instance.loaderOptions.allowTsInNodeModules &&
-      outputFileName.indexOf('node_modules') !== -1
+      outputFileName.indexOf('/node_modules/') !== -1
     ) {
       return undefined;
     }


johnnyreilly:
Thanks for contributing! This is a pretty discreet change and, on the face of it, it seems pretty safe to roll with. 

@andrewbranch do you have any views on this? I'm mindful of your comments here: https://github.com/TypeStrong/ts-loader/issues/1215#issuecomment-751810688

It would be straightforward to revert this if it turns out to be problematic.

johnnyreilly:
Side note: it's time to drop Travis usage. I believe they've dropped free support for open source projects. Fortunately @g-plane has already sorted out support with GitHub actions and so we're good. 

Funny old year, first appveyor starts consistently failing so we have to move away from that. Now Travis. Good odds on GHAs longterm support though 😉

src/servicesHost.ts
```diff
@@ -852,6 +852,14 @@ export function makeSolutionBuilderHost(
   function getInputFileNameFromOutput(
     outputFileName: string
   ): string | true | undefined {
+    // Unless we explicitly want to compile files in node_modules, exclude them from lookups
+    if (
+      !instance.loaderOptions.allowTsInNodeModules &&
+      outputFileName.indexOf('node_modules') !== -1
```
andrewbranch, line 858:
```suggestion
      outputFileName.indexOf('/node_modules/') !== -1
```

andrewbranch (changes requested):
Small nit to start with. I’ll have to investigate what this is used for in order to understand whether `allowTsInNodeModules` is a sufficient guard for it. Maybe I’ll try making an example project with node_modules symlinks.

src/servicesHost.ts
andrewbranch: Note: in the compiler, filenames get normalized with forward slashes, so I was assuming that here, but if it comes directly from webpack, maybe this is a bad assumption

berickson1: I had originally included slashes, but worried about path normalization. The other point in code that does TsInNodeModules checks without slashes, so I adopted that:
https://github.com/TypeStrong/ts-loader/blob/master/src/index.ts#L109

That being said, I can test on Windows and Mac to double check

berickson1: Unfortunately, it's platform sensitive
Screen Shot 2020-12-29 at 1 08 29 PM

andrewbranch:
I couldn’t observe any bad effects with this—when using a monorepo with symlinked packages through node_modules, I only ever witnessed the realpath being passed through here.

With the `allowTsInNodeModules` flag, it’s probably safe. It does feel like a bit of a hack, though. It seems like this function is analogous to [getSourceOfProjectReferenceRedirect](https://github.com/microsoft/TypeScript/blob/ea93ee6db9441eb51a1d816e114b0468363dcd9c/src/compiler/program.ts#L2722-L2745) in the compiler. There, the first time the program is asked for the input file of some project reference output file, we iterate through all resolved project references and make a map from their outputs to their inputs. From then on, it’s just a map lookup. I’m not sure why this code doesn’t either do the same thing or hook into the program code (maybe because the first time it runs is during `createProgram` itself). I think it would be best to get @sheetalkamat’s input. Even though I think this would probably be safe to merge, it would likely just be covering up a larger flaw.

berickson1:
I think there's a couple holes in the file watch/resolution paths based on some evidence I've seen in profiling. Specifically, there's a few caches that are cached at the beginning of every watch-build (fileExists, directoryExists cache) that result in a ton of filesystem operations to repopulate. These operations end up causing a bunch of additional garbage collection (as seen in #1228 both before and after). I'm guessing that there's a smarter approach that can be taken to invalidate only the portions of the affected cache based on webpack-watch callback. (An investigation like that would require a bit more understanding about the internal workings of the typescript compiler and ts-loader than I currently possess)

src/servicesHost.ts
```diff
@@ -852,6 +852,14 @@ export function makeSolutionBuilderHost(
   function getInputFileNameFromOutput(
     outputFileName: string
   ): string | true | undefined {
+    // Unless we explicitly want to compile files in node_modules, exclude them from lookups
+    if (
```
sheetalkamat, line 856: As @andrewbranch i am not familiar with monorepo projects out there that this will impact. I think somewhere we are missing the cache and that would be better solution in my opinion

berickson1:
Just wanted to check in here again. We've been using a patched version of ts-loader with this change with good success. I realize that there's probably a better solution deals with the multiple different runtime caches that exist, but unfortunately I don't have the familiarity or the bandwidth to work on this now. I'd love to bring this PR to a resolution before it gets too stale.
I'm ok if that's "we need a better approach" & a related issue or if that means moving forwards with this in it's current (or similar) state

johnnyreilly:
Hey @berickson1 

Thanks for checking in. I'm mindful of @sheetalkamat's comment:

> . I think somewhere we are missing the cache and that would be better solution in my opinion

My gut feeling is that if we merged this as is we'd probably get you to a better working place but break others. That makes me a little nervous I'm afraid.

berickson1:
I understand your sentiment and can sympathize. Given there's nobody else clamouring for this, it might be fine to close it out and keep us running on a fork (not ideal, but we will live). Eventually if this is solved "properly" , we could remove the patch in our fork.

If you (or someone else) were able to provide a cut more guidance about how the internal caches are supposed to behave, I may be able to help validate that. From my prior observations, the majority of caches are wiped out on file change callbacks, which seems to be the root of the problem. I'm just nervous to start blindly making changes there :)

berickson1:
FWIW, We've been running with a patched version of ts-loader with this optimization since the PR has been opened with no unexpected effects

johnnyreilly:
I'm open to us adding this - question @berickson1 : are you running on webpack 5?

berickson1:
> question @berickson1 : are you running on webpack 5?
We are :)

johnnyreilly:
How would you feel about implementing the cache approach suggested by @sheetalkamat  / @andrewbranch ?

PR #1227, opened 2020-12-29
Add documentation for the useCaseSensitiveFileNames option

berickson1:
Setting this option on projects with lots of files can reduce compilation time. Case-insensitivty requires normalization of file paths, which involves executing regex's. This shows up in the hot path of compilation profiling.
Switching to case sensitive file names skips the normalization step and increases build speed by ~5%  (closed-source project, sorry)

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@
     + [appendTsSuffixTo](#appendtssuffixto)
     + [appendTsxSuffixTo](#appendtsxsuffixto)
     + [onlyCompileBundledFiles](#onlycompilebundledfiles)
+    + [useCaseSensitiveFileNames](#useCaseSensitiveFileNames)
     + [allowTsInNodeModules](#allowtsinnodemodules)
     + [context](#context)
     + [experimentalFileCaching](#experimentalfilecaching)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -625,6 +626,15 @@ loading only those files that are actually bundled by webpack, as well as any `.
 by the `tsconfig.json` settings. `.d.ts` files are still included because they may be needed for
 compilation without being explicitly imported, and therefore not picked up by webpack.
 
+#### useCaseSensitiveFileNames
+| Type | Default Value |
+|------|--------------|
+| `boolean` | determined by typescript based on platform |
+
+The default behavior of `ts-loader` is to act as a drop-in replacement for the `tsc` command,
+so it respects the `useCaseSensitiveFileNames` set internally by typescript. The `useCaseSensitiveFileNames` option modifies this behavior,
```
johnnyreilly, line 635: Is it worth mentioning here what the default `useCaseSensitiveFileNames` behaviour of TypeScript is here?

berickson1: I stuck `determined by typescript based on platform` in the default above. Windows is case-insensitive and all other platforms are tested at runtime. Happy to include that if you like but I didn't want to tie docs to the internals of typescript
https://github.com/microsoft/TypeScript/blob/master/src/compiler/sys.ts#L1427

johnnyreilly: Oh no - that works well enough I think.

johnnyreilly:
Thanks!

PR #1228, opened 2020-12-29
Add a cache to file path mapping

berickson1:
This greatly reduces the time needed to resolve FilePathKey objects on watch builds. (~20% from 20s to 16s)

Note: The only case where I could see this optimization performing poorly is if a symlink was changed between builds, but I don't believe webpack handles that gracefully anyways.

Relates to #1215

See the performance traces below:

Before:
Screen Shot 2020-12-29 at 1 25 50 PM
Screen Shot 2020-12-29 at 1 26 01 PM

After:
Screen Shot 2020-12-29 at 1 35 58 PM
Screen Shot 2020-12-29 at 1 36 14 PM

### DIFF
--- a/src/instances.ts
+++ b/src/instances.ts
@@ -87,6 +87,8 @@ function createFilePathKeyMapper(
   compiler: typeof typescript,
   loaderOptions: LoaderOptions
 ) {
+  // Cache file path key - a map lookup is much faster than filesystem/regex operations & the result will never change
+  const filePathMapperCache = new Map<string, FilePathKey>();
   // FileName lowercasing copied from typescript
   const fileNameLowerCaseRegExp = /[^\u0130\u0131\u00DFa-z0-9\\/:\-_\. ]+/g;
   return useCaseSensitiveFileNames(compiler, loaderOptions)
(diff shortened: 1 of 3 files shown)


johnnyreilly:
Haha - nice visuals!  Did you make those by following the debugging instructions here: https://github.com/TypeStrong/ts-loader/blob/master/CONTRIBUTING.md#debugging-ts-loader-installed-from-npm-in-your-own-webpack-project ? Or was it something else?

src/instances.ts
```diff
@@ -87,23 +87,35 @@ function createFilePathKeyMapper(
...
   return useCaseSensitiveFileNames(compiler, loaderOptions)
     ? pathResolve
     : toFileNameLowerCase;
 
   function pathResolve(x: string) {
-    return path.resolve(x) as FilePathKey;
+    let cachedPath = filePathMapperCache.get(x);
+    if (!cachedPath) {
+      cachedPath = path.resolve(x) as FilePathKey;
+      filePathMapperCache.set(x, cachedPath);
```
johnnyreilly, line 102: Could we rename `x` to something more meaningful please?  I realise you didn't create that name, but reviewing the PR has made realise how it's not as helpful as could be.

src/instances.ts
```diff
@@ -87,23 +87,35 @@ function createFilePathKeyMapper(
...
   }
 
   function toFileNameLowerCase(x: string) {
-    const filePathKey = pathResolve(x);
-    return fileNameLowerCaseRegExp.test(filePathKey)
-      ? (filePathKey.replace(fileNameLowerCaseRegExp, ch =>
-          ch.toLowerCase()
-        ) as FilePathKey)
-      : filePathKey;
+    let cachedPath = filePathMapperCache.get(x);
```
johnnyreilly, line 108: same comment

johnnyreilly:
I've one naming nit but otherwise this LGTM.  Do you want to add an entry to the `CHANGELOG.md` and increment the `package.json` please?

berickson1:
> Haha - nice visuals! Did you make those by following the debugging instructions here: https://github.com/TypeStrong/ts-loader/blob/master/CONTRIBUTING.md#debugging-ts-loader-installed-from-npm-in-your-own-webpack-project ? Or was it something else?

Oh neat, I didn't realize that you had debugging instructions anywhere. I used a combination of `node --inspect` and [ndb](https://github.com/GoogleChromeLabs/ndb) to profile and do some initial investigations

src/instances.ts
berickson1: Will do. I wasn't sure how scoped you'd like changes to be in individual PRs

johnnyreilly:
Yeah I think @andrewbranch put the debug instructions together. Looking at this PR makes me think it would be great to have some perf testing instructions as well... Could you share what your workflow looked like to produce the above?

src/instances.ts
andrewbranch: Are the inputs relative paths? I’m surprised this has such an impact. Nice investigating 👍

johnnyreilly: That's a great question - can you confirm @berickson1?

berickson1: They are absolute paths. But path.resolve still does string normalization on every call (which results in one string allocation per character in the path and at least one string concat - and string allocation - operation per folder in the path)

berickson1:
> Yeah I think @andrewbranch put the debug instructions together. Looking at this PR makes me think it would be great to have some perf testing instructions as well... Could you share what your workflow looked like to produce the above?

Added in #1230

johnnyreilly (approved):

johnnyreilly:
Thanks!  https://github.com/TypeStrong/ts-loader/releases/tag/v8.0.13 is shipping now. 🚢

PR #1237, opened 2021-01-04
chore(deps): upgrade most production deps

AviVahl:
adjusted `chalk` import/initialization to new major

### DIFF
--- a/src/instances.ts
+++ b/src/instances.ts
@@ -1,4 +1,4 @@
-import chalk = require('chalk');
+import * as chalk from 'chalk';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as typescript from 'typescript';
--- a/src/instances.ts
+++ b/src/instances.ts
@@ -1,4 +1,4 @@
-import chalk, { Chalk } from 'chalk';
+import chalk = require('chalk');
 import * as fs from 'fs';
 import * as path from 'path';
 import * as typescript from 'typescript';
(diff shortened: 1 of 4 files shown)


src/instances.ts
```diff
@@ -1,4 +1,4 @@
-import chalk, { Chalk } from 'chalk';
+import chalk = require('chalk');
```
johnnyreilly, line 1: Could we avoid using `require` imports please?

AviVahl: done.

johnnyreilly:
Awesome. Could you increment the version the `package.json` and add an entry to the `CHANGELOG.md` please?

AviVahl:
Done. I avoided adding the `thanks @<myself>` to the changelog. Otherwise, I get flooded with Dependabot notifications including that `@`. XD

johnnyreilly:
I didn't think dependabot read `CHANGELOG.md` - could be wrong.

It's fine to not have your handle in there if that's the case. Please could you put "thanks Avi Vahl" instead then please? That's human readable and as such I'm fine with it.

AviVahl:
Done.

AviVahl:
Oh, and Dependabot picks up the release notes from the GitHub releases page. I already got spammed from a contribution to handlebars once, as they creditted my handle.

johnnyreilly:
> Oh, and Dependabot picks up the release notes from the GitHub releases page

Interesting - I would have imagined it used the related commits and the authors there instead. Go figure

johnnyreilly:
Shipped - thanks for the contribution!

Stanzilla:
@AviVahl this broke https://github.com/TypeStrong/ts-loader/issues/1240

PR #1244, opened 2021-01-31
Update README.md for latest webpack 5 syntax

e56:
The latest webpack WatchIgnorePlugin has updated syntax that expects an options object

https://webpack.js.org/plugins/watch-ignore-plugin/


johnnyreilly:
Thanks so much!

Issue #1243, opened 2021-01-25
Definition files are not updated after initial build when using watch mode in Webpack 5.


JonWallsten:
### Expected Behaviour
When using watch mode in Webpack 5 and I'm changing a file, the index.js and index.d.ts bundle files should be updated with the new code/definitions and I should be able to import the code in my app that consumes the code.
### Actual Behaviour
Only the index.js file is updated when adding new code. All d.ts files as well as the index.d.ts bundle remains untouched and I get compiler error when importing the new code (in runtime the code actually runs fine).
![image](https://user-images.githubusercontent.com/5993909/105712145-b75bcd00-5f19-11eb-865b-a561b0156a55.png)

### Steps to Reproduce the Problem
1. Run Webpack 5 in watch mode: `node node_modules\webpack\bin\webpack.js --config config/webpack.dev.ts --watch`
2. Change any code in the project
3. Check the dist folder and compare the times between the `.js` and the `d.ts` files
### Location of a Minimal Repository that Demonstrates the Issue.
(Working on it)

**webpack.dev.ts**
```
module: {
    rules: [
        {
            test: /\.ts$/,
            loader: 'ts-loader',
            options: {
                configFile: '../tsconfig.build.json'
            },
            include: [
                projectRootPath('typings')
                rootPath('src'),
                rootPath('typings')
]
        }
    ]
},
```
**tsconfig.build.json**
```
{
  "extends": "../../tsconfig.json",
  "compilerOptions": {
    "rootDir": "./src",
    "declaration": true,
    "declarationMap": true,
    "declarationDir": "./dist",
    "outDir": "./dist"
  },
  "files": [
    "./src/index.ts",
    "./typings/global.d.ts",
    "../../typings/global.d.ts"
  ]
}
```
**tsconfig.json**
```
{
  "compilerOptions": {
    "rootDir": ".",
    "baseUrl": ".",
    "target": "es5",
    "module": "commonjs",
    "moduleResolution": "node",
    "allowSyntheticDefaultImports": true,
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "removeComments": false,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "noEmitHelpers": true,
    "noEmitOnError": true,
    "importHelpers": true,
    "noImplicitAny": false,
    "skipLibCheck": true,
    "strictNullChecks": false,
    "strictPropertyInitialization": false,
    "pretty": true,
    "sourceMap": true,
[...]

PR #1249, opened 2021-02-03
Update definition files in watch mode in webpack@5

appzuka:
See https://github.com/TypeStrong/ts-loader/issues/1243 for discussion of problem and this fix.

### DIFF
--- a/src/instances.ts
+++ b/src/instances.ts
@@ -336,20 +336,32 @@ const addAssetHooks = !!webpack.version!.match(/^4.*/)
         makeAfterCompile(instance, false, true, instance.configFilePath)
       );
 
-      // Emit the assets at the afterProcessAssets stage
-      loader._compilation.hooks.afterProcessAssets.tap(
-        'ts-loader',
-        (_: any) => {
-          makeAfterCompile(
-            instance,
-            true,
-            false,
-            instance.configFilePath
-          )(loader._compilation, () => {
-            return null;
-          });
-        }
+      // makeAfterCompile is a closure.  It returns a function which closes over the variable checkAllFilesForErrors
+      // We need to get the function once and then reuse it, otherwise it will be recreated each time
+      // and all files will always be checked.
+      const cachedMakeAfterCompile = makeAfterCompile(
+        instance,
+        true,
+        false,
+        instance.configFilePath
       );
+
+      // compilation is actually of type webpack.compilation.Compilation, but afterProcessAssets
+      // only exists in webpack5 and at the time of writing ts-loader is built using webpack4
+      const makeAssetsCallback = (compilation: any) => {
+        compilation.hooks.afterProcessAssets.tap('ts-loader', () =>
+          cachedMakeAfterCompile(compilation, () => {
+            return null;
+          })
+        );
+      };
+
+      // We need to add the hook above for each run.
+      // For the first run, we just need to add the hook to loader._compilation
+      makeAssetsCallback(loader._compilation);
+
+      // For future calls in watch mode we need to watch for a new compilation and add the hook
+      loader._compiler.hooks.compilation.tap('ts-loader', makeAssetsCallback);
 
       // It may be better to add assets at the processAssets stage (https://webpack.js.org/api/compilation-hooks/#processassets)
       // This requires Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL, which does not exist in webpack4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 # Changelog
+
+## v8.0.15
+* [Update definition files in watch mode in webpack@5](https://github.com/TypeStrong/ts-loader/pull/1249) - thanks @appzuka,@JonWallsten,@alexander-akait
+* [Add afterDeclarations to getCustomTransformers in README.md](https://github.com/TypeStrong/ts-loader/pull/1248) - thanks @appzuka
+
 ## v8.0.14
 * [Upgrade `chalk`, `loader-utils`, and `semver` to latest stable versions](https://github.com/TypeStrong/ts-loader/pull/1237) - thanks Avi Vahl
 
--- a/README.md
+++ b/README.md
@@ -344,7 +344,7 @@ These options should be functions which will be used to resolve the import state
 #### getCustomTransformers
 | Type |
 |------|
-| ` (program: Program) => { before?: TransformerFactory<SourceFile>[]; after?: TransformerFactory<SourceFile>[]; } ` |
+| ` (program: Program) => { before?: TransformerFactory<SourceFile>[]; after?: TransformerFactory<SourceFile>[]; afterDeclarations?: TransformerFactory<SourceFile>[]; } ` |
 
 Provide custom transformers - only compatible with TypeScript 2.3+ (and 2.4 if using `transpileOnly` mode). For example usage take a look at [typescript-plugin-styled-components](https://github.com/Igorbek/typescript-plugin-styled-components) or our [test](test/comparison-tests/customTransformer).
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "8.0.14",
+  "version": "8.0.15",
   "description": "TypeScript loader for webpack",
   "main": "index.js",
   "types": "dist",


src/instances.ts
```diff
@@ -336,21 +336,33 @@ const addAssetHooks = !!webpack.version!.match(/^4.*/)
...
+        instance,
+        true,
+        false,
+        instance.configFilePath
       );
 
+      // compilation is actually of type webpack.compilation.Compilation, but afterProcessAssets
+      // only exists in webpack5 and at the time of writing ts-loader is built using webpack4
+      const makeAssetsCallback = (compilation: any) => {
+        compilation.hooks.afterProcessAssets.tap('ts-loader', () =>
```
JonWallsten, line 352: Since this code block is only accessible for Webpack 5 you can use the new hook:
```
compilation.hooks.processAssets.tap({ name: 'ts-loader', stage: webpack.Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL }, (_assets) => {
...
}
```

Webpack is already required, so no other changes needed.

johnnyreilly:
Thanks @appzuka - could you update the `package.json` and `CHANGELOG.md` please? Also could you respond to @JonWallsten's suggestion?

Cheers!

src/instances.ts
appzuka: ts-loader is currently built using webpack@4.5.0, which does not define webpack.Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL, so although the code would run under webpack 4 and 5 it cannot be built under webpack4 if we include this.

Also, I think it is OK to use the afterProcessAssets hook.  The various processAssets hooks provide a list of the assets that are being processed at that stage, but we are not using the list of assets.  We just need a place to hook into where we can emit the assets and afterProcessAssets seems to work fine.

alexander-akait: No, please use `compilation.hooks.processAssets`, you can check webpack version using `compiler.webpack` (if exists it is webpack@5, if no it is webpack@4), also using `compilation.hooks.processAssets` I can implement cache, so it will speed build in watch mode

appzuka: I tried building ts-loader using <code>compilation.hooks.processAssets</code> but it will not build because <code>webpack.Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL</code> is not defined in webpack 4. If you try to build using webpack 5 there are other errors.  I assume at some point in the future ts-loader will upgrade to webpack 5 but until then we cannot use anything not defined in webpack 4. afterProcessAssets is a valid webpack 5 hook so I'm not sure there is any reason to avoid using it.

alexander-akait: Example of usage https://github.com/webpack-contrib/terser-webpack-plugin/blob/v4.2.3/src/index.js#L647 (webpack@4 and webpack@5), on this hook I can't help with cache, so if you want speedup your build, if you don't use cache and do developers happy - OK

JonWallsten: @alexander-akait: I think he means that the loader itself is built with webpack@4.5.0, so he would get an error while building. Is it possible to bump Webpack to 4.44.0 for ts-loader?

appzuka: Yes, the file @alexander-akait linked to in terser-webpack-plugin is a Javascript file.  This issue is that ts-loader is written in TypeScript and you can't build it with webpack5 constants using webpack 4.

We are already branching based on the webpack version:
https://github.com/TypeStrong/ts-loader/blob/6816735404350ece29b9bce40b2addb396dcba8b/src/instances.ts#L321

I agree this code can be cleaned up in the future but I was aiming to fix the current code today without introducing a breaking change.

alexander-akait: We can use `@ts-ignore` and add comment, I think it is not bad idea for compatibility reasons

alexander-akait: `ts-loader` already slow in some situations, doing it you really will get issues about problems with perf and need to implement cache to solve it, I don't understand why do bug fix with perf regressions, my opinion

appzuka: @alexander-akait, could you explain what performance regression is being introduced here?  The original goal was to stop emitting assets in the afterCompile hook because that it deprecated in Webpack 5.  The code where the assets are emitted is being moved to a different hook but apart from that it runs the same.  I have confirmed that the same assets are emitted in the old and new code.

I would need to look into the webpack caching used in Terser although if I understand correctly that is avoiding reprocessing assets unnecessarily.  I believe ts-loader uses tsc intelligently to only re-compile files as needed so it may already be doing the same thing.  Subsequent builds in watch mode using ts-loader are already very fast.

alexander-akait: @appzuka oh, it is hard to explain, but I will try, the main problem what you need **always** emit assets otherwise it will create when `clean-webpack-plugin` or `output.clean` (new features to replace `clean-webpack-plugin`) will not work, it will remove assets between compilations, but emitting assets on each compilation will create big regressions, especial if you have many files, to avoid this you need to use cache, there is no problems with `tsc`, here problem how webpack works with new assets (https://github.com/TypeStrong/ts-loader/blob/484e47f9f900e91680f22587bc072feb688ae1c2/src/after-compile.ts#L406), to be honestly problem with perf already exists, which is why I recommend using the right hook so that we can solve the problem in the next iteration. Also emitting assets using https://github.com/TypeStrong/ts-loader/blob/484e47f9f900e91680f22587bc072feb688ae1c2/src/after-compile.ts#L406 is invalid in webpack@5 and can create bad situations, you need `compiler.webpack.sources`.

appzuka: @alexander-akait,  many thanks for taking the time to explain.  I have also thought that ts-loader could be simplified and streamlined.  I would worry about making changes at this level because I don't fully understand the interactions between ts-loader, tsc and webpack. Perhaps there is a case for looking at the ts-loader's architecture as part of a major overhaul for webpack 5.

I submitted this PR based on the solution you and @JonWallsten proposed so that an immediate bug can be fixed and I don't believe it introduces a performance regression.  I don't think anyone has the time today to delve deeper so I'll look at using your suggestion of using <code>@ts-ignore</code> and the new hook so we can get the bug fixed.

johnnyreilly: @appzuka - we're building `ts-loader` with webpack 4 - of which the types are located here I think: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/webpack/index.d.ts

If we get `PROCESS_ASSETS_STAGE_ADDITIONAL` added to the types does that help?  (I'm a maintainer of these types BTW)

johnnyreilly: If you were to raise a PR against Definitely Typed, I could review it https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/webpack/index.d.ts

JonWallsten: @johnnyreilly I think they were introduced in Webpack@4.44.0. Is that correct @alexander-akait?

alexander-akait: We can `extends` some types to solve problem between versions

johnnyreilly: I think just adding new portions to the definition should be fine.  People will only use it if they're using that version of webpack.  Should be no drama.

johnnyreilly: Do we just add `PROCESS_ASSETS_STAGE_ADDITIONAL` in the `Compilation` defined here?:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/webpack/index.d.ts#L1269

alexander-akait: I think it is wrong, because real `Compilation` for v4 doesn't have this properties

johnnyreilly: Oh okay - I misunderstood

JonWallsten: Should we just accept this PR as it is to fix the immediate issue and start working on a PR with Webpack > 4.40.0 with the help of @alexander-akait, and also upgrade the Webpack version for ts-loader itself? Seems like there's little middle ground here. Or am I missing something?

johnnyreilly: What's your view @appzuka?

appzuka: I think it makes sense to release this now as it fixes a bug and I believe it does not affect performance.  I have updated the code to change package.json and Changelog.  I have also changed the hook and used <code>@ts-ignore</code> to allow building:
```
        // @ts-ignore compilation.hooks.processAssets is only defined in webpack 5
        compilation.hooks.processAssets.tap({name: 'ts-loader', stage: webpack.Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL}, () =>
          cachedMakeAfterCompile(compilation, () => {
            return null;
          })
        );
```

I believe webpack 4.40 will not be enough to drop the <code>@ts-ignore</code>.  We will need to get to webpack 5.

I am running the tests now and then I'll push the new version.

appzuka: husky won't let me use <code>@ts-ignore</code> when I try to commit:
```
✖ yarn lint found some errors. Please fix them and try committing again.
yarn run v1.22.4
$ eslint -c .eslintrc.js --ext .ts ./src /home/nick/scratch/ts-loader/src/instances.ts

/home/nick/scratch/ts-loader/src/instances.ts
352:9  error  Do not use "// @ts-ignore" because it alters compilation errors  @typescript-eslint/ban-ts-comment

✖ 1 problem (1 error, 0 warnings)
```

I suggest to go back to using the afterProcessAssets hook so I there are no errors to ignore. I can include a comment that this should be cleaned up in future and I don't believe there is any downside to today's release. @johnnyreilly - let me know what you would like me to do.

johnnyreilly: Thanks @appzuka - could you update the `package.json` and `CHANGELOG.md` please? Then let's ship

alexander-akait: Just note: maybe to do new major release with dropping webpack@4 is not bad idea

johnnyreilly: Yeah - I'm in favour

johnnyreilly: (not in this PR obviously)

alexander-akait: Also I can help with cache and optimizations :star:

johnnyreilly: Thanks in advance!

johnnyreilly (approved):

johnnyreilly:
Thanks all - shipping now https://github.com/TypeStrong/ts-loader/releases/tag/v8.0.15

johnnyreilly:
I've started work on the webpack 5 migration: 

https://github.com/TypeStrong/ts-loader/pull/1251

Help gratefully received!  I'm already stuck on a few API renames :smile:

JonWallsten:
@johnnyreilly: I'll do whatever I can to help. How do we coordinate?

johnnyreilly:
Let's collaborate on the PR!

PR #1251, opened 2021-02-04
webpack 5 migration

johnnyreilly:
This PR will be where the migration of ts-loader to webpack 5 takes place.  Help is appreciated!

- [x] dependencies upgraded to webpack 5 (`enhanced-resolve`, `webpack-cli`, `webpack`)
- [x] drop support for webpack 4
- [x] make node 12 the minimum version
- [x] code compiles
- [ ] ~~drop support for experimental watch API~~ - see https://github.com/TypeStrong/ts-loader/pull/1251#issuecomment-806195840
- [x] upgrade comparison test pack running against webpack 5
- [x] execution tests pass
- [x] increment major version to 9.0 and update `CHANGELOG.md`
- [x] update `README.md`

cc @alexander-akait @appzuka @JonWallsten - all advice and contributions towards this are gratefully received.

Feel free to PR or push into this branch to work on this.

Some discussion here: https://github.com/webpack/webpack/discussions/12475#discussioncomment-347051

## UPDATE

Okay I think we're ready to ship.  I'm going to:

- [x] create a branch for webpack 4 - so if there's a need to ship another change to support webpack 4 we can (I don't expect to but you never know)
- [x] write up a blog post to detail this
- [ ] ship


appzuka:
Would this be a good point to drop support for experimentalWatchApi?  As far as I can see it has never worked and it adds complexity to the code.

JonWallsten:
Don't forget to update webpack-cli to 4.5.0 (I think it's the latest).

johnnyreilly:
> Would this be a good point to drop support for experimentalWatchApi? As far as I can see it has never worked and it adds complexity to the code.

Go for it @appzuka

johnnyreilly:
> Don't forget to update webpack-cli to 4.5.0 (I think it's the latest).

Good shout

JonWallsten:
@johnnyreilly How do you normally test the loader when you are working? Do you have an example project you are running locally?

JonWallsten:
@johnnyreilly: How do you wanna handle these scenarios?
image

Potential fix:
image

johnnyreilly:
That seems like a legitimate workflow @JonWallsten - to your question about the fix: yeah maybe. Question: is it legitimate that a module may be a `NormalModule`? If so then seems reasonable

JonWallsten:
@johnnyreilly / @appzuka:
Getting back to this:
image

image

Am I correctly to assume that the first images shows where we ask TypeScript to compile and the second we pass those assets to Webpack?

JonWallsten:
> That seems like a legitimate workflow @JonWallsten - to your question about the fix: yeah maybe. Question: is it legitimate that a module may be a `NormalModule`? If so then seems reasonable

According to the debugger, the current module I'm looking at is a `NormalModule`.
image
And accroding to the types:
image
`NormalModule` is the only Module related class that has the property `resources`. So we can assume that it has to be a `NormalModule`? I don't see any other way of getting the types to work besides setting it as any. And that would be sad! :(

johnnyreilly:
If that's what the debugger says, I trust it!

JonWallsten:
> If that's what the debugger says, I trust it!

I can't promise that all modules are NormalModules, but if we trust the types, the `resource` prop only exists on `NormalModule` and since we do a nullcheck on that property in extension we should be able to check if it's a `NormalModule` so we get the correct Type information in the editor.

appzuka:
@JonWallsten,

I put some logging statements to clarify where assets are transpiled and emitted.  When running on a small project (index.ts with sub.ts as a dependency) I get the following output:
```
Processing /home/nick/scratch/ts-loader-1243/src/index.ts
getEmitOutput:Sourcefile: /home/nick/scratch/ts-loader-1243/src/index.ts
getEmitOutput:Output: /home/nick/scratch/ts-loader-1243/dist/index.js
getEmitOutput:Output: /home/nick/scratch/ts-loader-1243/dist/index.d.ts.map
getEmitOutput:Output: /home/nick/scratch/ts-loader-1243/dist/index.d.ts
Returning transpiled file to webpack: /home/nick/scratch/ts-loader-1243/src/index.ts

Processing /home/nick/scratch/ts-loader-1243/src/sub.ts
getEmitOutput:Sourcefile: /home/nick/scratch/ts-loader-1243/src/sub.ts
getEmitOutput:Output: /home/nick/scratch/ts-loader-1243/dist/sub.js
getEmitOutput:Output: /home/nick/scratch/ts-loader-1243/dist/sub.d.ts.map
getEmitOutput:Output: /home/nick/scratch/ts-loader-1243/dist/sub.d.ts
Returning transpiled file to webpack: /home/nick/scratch/ts-loader-1243/src/sub.ts

** Running processAssets hook
getEmitOutput:Sourcefile: /home/nick/scratch/ts-loader-1243/src/index.ts
getEmitOutput:Output: /home/nick/scratch/ts-loader-1243/dist/index.js
getEmitOutput:Output: /home/nick/scratch/ts-loader-1243/dist/index.d.ts.map
getEmitOutput:Output: /home/nick/scratch/ts-loader-1243/dist/index.d.ts
>> outputFileToAsset: index.d.ts.map
>> outputFileToAsset: index.d.ts
getEmitOutput:Sourcefile: /home/nick/scratch/ts-loader-1243/src/sub.ts
getEmitOutput:Output: /home/nick/scratch/ts-loader-1243/dist/sub.js
getEmitOutput:Output: /home/nick/scratch/ts-loader-1243/dist/sub.d.ts.map
getEmitOutput:Output: /home/nick/scratch/ts-loader-1243/dist/sub.d.ts
>> outputFileToAsset: sub.d.ts.map
>> outputFileToAsset: sub.d.ts
```

I believe the program.emit statement in getEmitOutput is where the output from the TypeScript compiler is obtained.  The input is a single .ts file and the output is a .js file and .d.ts and .d.ts.map files, if required.

The .js file is returned to webpack as the result of the compilation.  This happens until all .ts files have been processed.

Then, during the processAssets hook, makeAfterCompile goes through all the files the TypeScript instance viewed, gets the outputs files from program.emit again, and adds .d.ts and .d.ts.map files as assets.

It seems inefficient that program.emit is called multiple times to get the output from .ts files, but I believe the output is cached so on subsequent calls it is just hitting the cache.

On this small example, each asset is only emitted to webpack once. So it seems that the asset output caching recommended by @alexander-akait may not be required.  However, when I run this in watch mode the assets are emitted even if their contents have not changed, so there could be a benefit in implementing a cache around the asset output in watch mode. There could also be a benefit in more complex projects.

JonWallsten:
@appzuka: Thanks for the explanation! Do we know why the files are emitted in different steps? Is it supposed to be like that or just legacy?

johnnyreilly:
Hey @sokra

I hope you are keeping well!

I seem to remember you put together a doc which detailed the API changes between webpack 4 and 5 targeted at maintainers of loaders and plugins. We're finally getting round to building `ts-loader` with webpack 5 and bumping on a few missing pieces.

Do you happen to remember where that doc lives? For example we're trying to track down `webpack.loader.LoaderContext` and not finding it.

All help gratefully received!

appzuka:
@JonWallsten 
> Do we know why the files are emitted in different steps? Is it supposed to be like that or just legacy?

I understand why the js is returned separately from the .d.ts files but not why they are returned at different stages. A simple loader as described at https://webpack.js.org/contribute/writing-a-loader/ just takes an input file and transforms it to a string that is valid Javascript.  In the case of ts-loader the input file is a .ts file, which is transformed into Javascript but also .d.ts and other files need to be added to the assets in the output directory.  The Javascript output by typescript is not emitted as an asset but rather passed back to webpack for inclusion in the bundle.

What is not clear to me is why the .d.ts and other files are emitted together when all the .ts files have been processed rather than being emitted as assets during the getEmitOutput. It would make things simpler as we would not need the compilation/processAssets hooks at all, but perhaps there is a reason it was done this way.

It seems to me that processAssets is not really the right place to generate the assets.  In the Terser example processAssets was used but then Terser is a plugin which processes assets generated by other plugins and loaders.  In ts-loader we are not processing the assets, we are creating them.

I'm wary of changing the architecture in case we are repeating a mistake someone fixed several years ago, although simplifying ts-loader may also help improve performance and reduce errors.

johnnyreilly:
> I'm wary of changing the architecture in case we are repeating a mistake someone fixed several years ago, although simplifying ts-loader may also help improve performance and reduce errors.

I have a feeling that there was work done around this but it was a long time ago and frankly I don't remember the details. I'm open to changes being made but it's possible there's a reason things are the way they are now and I'm just not recalling

JonWallsten:
It kind of make sense though that the d.ts and d.ts.map files are concidered additional files since they are not part of the bundle and Webpack is not built around TypeScript. So even though they are vital to us, it's not vital to Webpack in general. Thoughts?

johnnyreilly:
> It kind of make sense though that the d.ts and d.ts.map files are concidered additional files since they are not part of the bundle and Webpack is not built around TypeScript. So even though they are vital to us, it's not vital to Webpack in general. Or am i missing something?

Sorry I missed this @JonWallsten - I don't quite follow the question.

Incidentally, it turns out that webpack isn't exporting all types... So some we'll have to hold onto ourselves

johnnyreilly:
Okay - the code compiles now; I borrowed a bunch of stuff from your branch @JonWallsten - thanks!

No tests pass alas

johnnyreilly:
Execution tests passing! (or it looks like it!)

JonWallsten:
> > It kind of make sense though that the d.ts and d.ts.map files are concidered additional files since they are not part of the bundle and Webpack is not built around TypeScript. So even though they are vital to us, it's not vital to Webpack in general. Or am i missing something?
> 
> Sorry I missed this @JonWallsten - I don't quite follow the question.
> 
> Incidentally, it turns out that webpack isn't exporting all types... So some we'll have to hold onto ourselves

This is like jeopardy. You give the answer I ask the question! 😂 Why is ts-files and d.ts/d.ts.map files emitted in different places in the code and Webpack lifecycle. I would say it's because declaration files are not part of the "code" in that sense, it's additional bonus stuff that will end up as separate files, not part of the bundle, in contrast to the transpiled/compiled (which is it?) code and the sourcemaps.

JonWallsten:
> Okay - the code compiles now; I borrowed a bunch of stuff from your branch @JonWallsten - thanks!
> 
> No tests pass alas

What about this little guy? 🙄

image

johnnyreilly:
Oh yeah - I saw that but I wasn't sure if it was necessary.

johnnyreilly:
Seems to be working without it

johnnyreilly:
Current problem is that in the comparison tests, the `bundle.js` files aren't being copied correctly.  Not sure why. Anyway the kids are up now so I've got to check out.

JonWallsten:
> Seems to be working without it

Yeah, it is working without it, but it was the preferred/new way according to @alexander-akait. We discussed it in the old PR I think. It would make the code a little bit easier to understand at least. Maybe we can try it later and measure if it's better in some way.

JonWallsten:
> Current problem is that in the comparison tests, the `bundle.js` files aren't being copied correctly. So tired - been doing this since 3am. Not sure why. Anyway the kids are up now so I've got to check out.

Jesus man! What an effort! 🎖 No wonder there were so many changes. I'll help you look later today if you want to! :)

johnnyreilly:
> Yeah, it is working without it, but it was the preferred/new way according to @alexander-akait

Feel free to PR off this branch and add it back. If it works then awesome!

johnnyreilly:
Just regenerated the test output - let's see if the comparison tests are a touch happier now...

JonWallsten:
@johnnyreilly Whats up with the execution tests on windows? They all seems to be stuck.

JonWallsten:
Running them locally: 
image

JonWallsten:
image
Maybe these needs to be updated or removed? Webpack 2/3 as a peerDependency?

JonWallsten:
image
Is this change fine? Not sure why it started complaining now and not before?

johnnyreilly:
I think those messages come from the tests that use babel-loader.  Our dependency is pretty old.  Still works though.

Yeah the tests are super weird; all the Windows tests seem to be stuck.  I switched to my Windows laptop and I can confirm all works fine on Windows.  I think that GitHub Actions is having an off day. I've seen that previously - it usually fixes up in the end.

johnnyreilly:
> Is this change fine? Not sure why it started complaining now and not before?

It's fine I think.  It's reporting the same error, but twice.  That doesn't seem to be generally repeated so I think it's okay.

JonWallsten:
> > Is this change fine? Not sure why it started complaining now and not before?
> 
> It's fine I think. It's reporting the same error, but twice. That doesn't seem to be generally repeated so I think it's okay.

Not exactly though. It complains about not finding component/myComponent, which is the component resolved through an alias in the webpack config in that test. If Webpack is rewriting the path based on the alias I'm not sure why we get an error. But if it builds I guess it doesn't matter. The original path doesn't exist, so the error is warranted. Maybe it's a separate typecheck and not an error from the build os something. My knowledge is not good enough here.

johnnyreilly:
> Not exactly though. It complains about not finding component/myComponent, which is the component resolved through an alias in the webpack config in that test. If Webpack is rewriting the path based on the alias I'm not sure why we get an error. But if it builds I guess it doesn't matter. The original path doesn't exist, so the error is warranted. Maybe it's a separate typecheck and not an error from the build os something. My knowledge is not good enough here.

You might be onto something - feel free to dig into it.  

I think that GitHub Actions is experiencing issues with Windows which is stopping Windows tests from running.  Do you know where you can report this?  Been looking but can't find anything

JonWallsten:
> > Not exactly though. It complains about not finding component/myComponent, which is the component resolved through an alias in the webpack config in that test. If Webpack is rewriting the path based on the alias I'm not sure why we get an error. But if it builds I guess it doesn't matter. The original path doesn't exist, so the error is warranted. Maybe it's a separate typecheck and not an error from the build os something. My knowledge is not good enough here.
> 
> You might be onto something - feel free to dig into it.
> 
> I think that GitHub Actions is experiencing issues with Windows which is stopping Windows tests from running. Do you know where you can report this? Been looking but can't find anything

I can try to do the exact same thing in my local test app and see what happens. It's easier if I can remove the testing tools from the mix. 

Sorry, I have literally zero knowledge about GitHub CI. Ask me about Jenkins though! 😂

JonWallsten:
**Webpack config:**
image

**Build succeeds but type checking fails:**
image

**Output looks correct:**
image
image

My opinion is that this is not `ts-loader`'s responsibility. You can use "paths" in tsconfig if you want to do this properly. The only reason for using alias in webpack is for non-ts-stuff (like sass imports etc) and then type checking is not an issue.

You can also fix this by supressing the error:
image
image

So I would say this is "working as expected" and fine. I'm more surprised it didn't throw an error before.

johnnyreilly:
So this test predates my involvement with `ts-loader` (yup it's that old!) And predates the paths support etc. It dates back to TypeScript 1.4 I think. It may even be worth dropping as a test case as I'm not sure it represents good practice anymore

JonWallsten:
> So this test predates my involvement with `ts-loader` (yup it's that old!) And predates the paths support etc. It dates back to TypeScript 1.4 I think. It may even be worth dropping as a test case as I'm not sure it represents good practice anymore

Wow! That's old! Before I started working with TypeScript. Yeah, what we are testing is that Webpack's modules resolution works? I guess Webpack has there own tests for that? It's not ts-loaders responsibility. We could however convert to test to use paths if you don't have test for that? Otherwise, definitely remove.

johnnyreilly:
> We could however convert to test to use paths if you don't have test for that? Otherwise, definitely remove.

Yeah a paths execution test instead of the alias comparison test. Execution tests are more reliable than comparison tests so are preferred. However they don't cater for expected failure cases and so aren't always appropriate

johnnyreilly:
Hey @andrewbranch 

We're slowly in the process of migrating `ts-loader` to webpack 5.  It's going pretty well, but we're bumping on something that I wonder if you can help with.

On Ubuntu, all execution tests / comparison tests are passing and looking good.  On Windows we're getting some strange behaviour.  Weirdly execution tests aren't running. I think may be a GitHub Actions outage issue or something and I've validated that they do pass on a Windows machine of mine.  That's not what I'm messaging about though; the weirdness we're encountering is with a number of Project References related tests on Windows which you originally wrote I think.  You can see failures showing up in GitHub Actions for Windows here:

https://github.com/TypeStrong/ts-loader/runs/1933294427?check_suite_focus=true

Whilst on Ubuntu everything is great:

https://github.com/TypeStrong/ts-loader/runs/1933293599?check_suite_focus=true

We had to update all the comparison tests with new output for webpack 5.  Unfortunately given the nature of the change it relies on on eyeballs to validate that behaviour hasn't changed that much.  (Eyeballs don't automate 👀 😄  ).  But when you look at the diff here:

https://github.com/TypeStrong/ts-loader/pull/1251/files#diff-eefd388bdbd140de5006b084bc0e913931bd7f204b0d3f3220b5238952474c44

You should see that there are no radical changes.  (BTW - that's a giant page because of all the diffs; may take 10 seconds before it jumps to the right place)

Can you think of any reason why Windows would be choking on Project References tests and Ubuntu would be fine?  It's very strange and I'm drawing a blank.

JonWallsten:
> > We could however convert to test to use paths if you don't have test for that? Otherwise, definitely remove.
> 
> Yeah a paths execution test instead of the alias comparison test might be good. Execution tests are more reliable than comparison tests so are preferred. However they don't cater for expected failure cases and so aren't always appropriate

I'll try to write one then! 👍

andrewbranch:
@johnnyreilly no, unfortunately I can’t make sense of what’s happening there. I don’t know that those statuses mean—“compared for emit” vs. “cached”? Does that ring a bell to you?

src/interfaces.ts
```diff
@@ -318,3 +276,260 @@ export interface ResolvedModule {
...
+  _compiler: any;
+
+  /**
+   * Hacky access to the Module object being loaded.
+   */
+  _module: any;
+
+  /** Flag if HMR is enabled */
+  hot: boolean;
+}
```
alexander-akait, line 535: I think better to avoid using this, because it is invalid interface, just keep copy only necessary properties, I think we need improve this on webpack side

johnnyreilly:
> @johnnyreilly no, unfortunately I can’t make sense of what’s happening there. I don’t know that those statuses mean—“compared for emit” vs. “cached”? Does that ring a bell to you?

Yeah - it's not super intuitive as messages go @andrewbranch 😄 

@alexander-akait are you able to confirm what “compared for emit” vs. “cached” mean in the context of webpack output messages?

alexander-akait:
`compared for emit` - means webpack checks content of file to avoid extra writing
`cached` - means it was used from cache :smile:

johnnyreilly:
> `compared for emit` - means webpack checks content of file to avoid extra writing

Okay cool - have you any idea why on Windows we'd see `compared for emit` and on Linux `cached`?

alexander-akait:
Do you have caches?

johnnyreilly:
Internal to ts-loader, yes.

https://github.com/TypeStrong/ts-loader/blob/953358eb1b090dd9add0824c4746220b0ebe56c1/src/servicesHost.ts#L1201

Is that what you mean? (It can be turned off by toggling the `experimentalFileCaching` off)

alexander-akait:
Should not affect on `compared for emit` and `cached`, where you see this problem? On CI or locally? Maybe something with timestamps or content, what is `EOL`?

johnnyreilly:
> where you see this problem? On CI or locally

Both - the difference in behaviour is between Linux and Windows. It didn't exist with webpack 4 - it started presenting when running our test pack against webpack 5.

alexander-akait:
Maybe you can create small example? I will look at this

johnnyreilly:
> Maybe you can create small example? I will look at this

Thanks! Simple way to reproduce:

- clone `ts-loader` and checkout this branch
- `yarn`
- `yarn build`
- `yarn run comparison-tests -- --single-test projectReferencesOutDirWithPackageJsonAlreadyBuilt`

This will run one of the project references tests that displays the curious behaviour.

I did this on my Ubuntu laptop and tests passed. On my Windows one it failed. I'd imagine that using WSL on Windows will allow you to repro both passing and failing on a single Windows machine (but haven't tested).

Thanks for your time! 🌻❤️

sokra:
For both `cached` and `compared for emit` content on disk is identical with the content that should be written.

For `cached` webpack already knows that it and it has to do nothing.
For `compared for emit` webpack doesn't know that it's identical and had to read it's content to check that.

`cached` will happen when:
- file is using any type of hash and the file already exists, or
- during watch mode when the written Source object has already been written to that file (same identity of the object)

johnnyreilly:
Thanks for the clarification @sokra - appreciate it

JonWallsten:
@johnnyreilly: 
I don't have an issue running these tests of Windows locally:
image

johnnyreilly:
I think we've two different threads going on here... @JonWallsten on execution tests and me / @alexander-akait on comparison tests for anyone reading this who is puzzled

johnnyreilly:
I haven't been working on it @alexander-akait 

I thought you might be from the comment above:  https://github.com/TypeStrong/ts-loader/pull/1251#issuecomment-782891134

Essentially puzzled as to why we're seeing different behaviour with webpack 5 between Linux and Windows

johnnyreilly:
> We are fixed some windows problems in the latest path can you try?

So if I upgrade to the latest version of webpack 5 you think behaviour might be different?

alexander-akait:
Maybe, I'm afraid I might have missed something here, you can show me how to reproduce

johnnyreilly:
> Maybe, I'm afraid I might have missed something here, you can show me how to reproduce

Do these instructions not work? https://github.com/TypeStrong/ts-loader/pull/1251#issuecomment-782891134

I think this should allow you to reproduce? You'll need to run on Windows and Linux to compare

alexander-akait:
Thanks, I will look at this tomorrow

johnnyreilly:
Thanks so much!

alexander-akait:
@johnnyreilly sorry for delay, how I can update tests, maybe command, because after new install a lot of changed happens

alexander-akait:
And I don't have problem too with old version on windows and linux...

johnnyreilly:
> how I can update tests, maybe command, because after new install a lot of changed happens

Thanks @alexander-akait! Instructions here: https://github.com/TypeStrong/ts-loader/tree/master/test/comparison-tests#regenerating-test-data

alexander-akait:
@johnnyreilly i.e. Are these tests broken https://github.com/TypeStrong/ts-loader/runs/1933294427?check_suite_focus=true? Sorry for stupid questions :smile: 

But here no problems, output is different because in github actions line end is `CR LF` on widnows so we have different size of assets and here logic https://github.com/webpack/webpack/blob/master/lib/stats/DefaultStatsFactoryPlugin.js#L1653 which decides to print it on one line (i.e. `assets by status A-NUMBER-OF KiB [cached] 5 assets`) or split it into several lines (i.e. each assets on own line with own information), if you remove logic for replacing `A-NUMBER-OF` you will see different size.

johnnyreilly:
> @johnnyreilly i.e. Are these tests broken https://github.com/TypeStrong/ts-loader/runs/1933294427?check_suite_focus=true? Sorry for stupid questions 😄

They appear to be "broken" as of webpack 5, yes.  But crucially only on Windows; not Ubuntu.  It's important to be clear here, these are a type of snapshot test and so we're just looking for the same behaviour regardless of whether it's Windows or Ubuntu.  We do some normalisation to handle different types of carriage return between Windows / Mac / Linux etc anyway

So with webpack 4, post normalisation behaviour was identical.  For these tests with webpack 5 that seems not to be the case.  And that's the mystery.

alexander-akait:
Weird, very weird, maybe ts generate different contents... Can we disable fast fail for test and check other broken tests?

johnnyreilly:
> maybe ts generate different contents

No - TypeScript output is not affected.  What's strange is that this works fine with webpack 4 and doesn't with webpack 5. 

As far as I can tell, there's an change with webpack 5 such that Ubuntu and Windows behave differently.

> Can we disable fast fail for test and check other broken tests?

It doesn't fail fast - it tries all the tests.  Most failures are project references related.  Here's the ones that behave differently between Ubuntu and Windows with webpack 5:

- projectReferencesOutDirWithPackageJsonAlreadyBuilt
- projectReferencesOutDirWithPackageJson
- projectReferencesOutDir
- projectReferencesNotBuilt_WatchApi
- projectReferencesNotBuilt_SemanticErrorInReference_WatchApi
- projectReferencesNotBuilt_SemanticErrorInReference_Composite_WatchApi
- projectReferencesNotBuilt_SemanticErrorInReference
- projectReferencesNotBuilt_ErrorInProject_WatchApi
- projectReferencesNotBuilt_ErrorInProject_Composite_WatchApi
- projectReferencesNotBuilt_ErrorInProject
- projectReferencesNotBuilt_Composite_WatchApi
- projectReferencesNotBuilt
- projectReferencesNoSourceMap
- projectReferencesMultipleDifferentInstance
- projectReferencesMultiple
- projectReferences
- declarationOutputWithMaps
- declarationOutput

alexander-akait:
> No - TypeScript output is not affected. What's strange is that this works fine with webpack 4 and doesn't with webpack 5. As far as I can tell, there's a change with webpack 5 such that Ubuntu and Windows behave differently.

Because you don't test webpack with the `futureAssetsEmit` option. So I think we have this bug in v4 too.

I think bug in path for emitting assets on windows we use `\` in asset name, Do you have windows? We can check it, let's open `provideErrorsToWebpack`  and look result `filesToCheckForErrors.entries()` on windows and linux

johnnyreilly:
> Because you don't test webpack with the futureAssetsEmit option. So I think we have this bug in v4 too.

So you think that if we ran the test pack with `futureAssetsEmit: true` for webpack 4 it would error in the same way?

> I think bug in path for emitting assets on windows we use `\` in asset name

Can you explain what you mean by this? I don't quite follow.  We do normalise `\` to `/` for the purposes of comparison already:

https://github.com/TypeStrong/ts-loader/blob/656d6b89aa9904b06e35de1d84efb0b560b440f0/test/comparison-tests/create-and-execute-test.js#L243

alexander-akait:
> So you think that if we ran the test pack with futureAssetsEmit: true for webpack 4 it would error in the same way?

Should yes, but webpack v5 has more complex logic, need to test, but I think it will be same

> Can you explain what you mean by this? I don't quite follow. We do normalise \ to / for the purposes of comparison already:

We should normalize it for `emitAsset`, `stats.compilation.assets` should always use `/` (all relative paths should use `/`)

johnnyreilly:
> We should normalize it for emitAsset, stats.compilation.assets should always use / (all relative paths should use /)

We normalise all output from tests and so I don't think this is an issue.  Unless I'm misunderstanding you?

alexander-akait:
> We normalise all output from tests and so I don't think this is an issue. 

We should not normalize it in tests, we should **always** pass it using `/` inside `loader`/`plugin` (i.e. ``).
```ts
function outputFileToAsset(
  outputFile: ts.OutputFile,
  compilation: webpack.Compilation
) {
  // We should use `/` here
  const assetPath = path.relative(
    compilation.compiler.outputPath,
    outputFile.name
  );
  // We should use `compiler.webpack.sources.RawSource` here
  compilation.assets[assetPath] = {
    source: () => outputFile.text,
    size: () => outputFile.text.length,
  } as any;
}
```

Otherwise cache is broken as you can see in tests

johnnyreilly:
Oh wait - I think I see what you mean!  I have to break off now - but yeah this might be worth pursing.  Do you want to have a go at making that change?

johnnyreilly:
Hey @alexander-akait - I've implemented your suggestion; let's see what happens in the build.

BTW I didn't know what you meant by:

```
// We should use `compiler.webpack.sources.RawSource` here
```

And so haven't done that.

JonWallsten:
> Hey @alexander-akait - I've implemented your suggestion; let's see what happens in the build.
> 
> BTW I didn't know what you meant by:
> 
> ```
> // We should use `compiler.webpack.sources.RawSource` here
> ```
> 
> And so haven't done that.

@johnnyreilly He's talking about the fix in my branch:
image

So if we combine his path relative with my fix for the assets it's should be what he wants.
The relative path however will not fix our projectReference-problem.

image
image

Still working on that. Although it would be nice if it was fixed in Webpack I can buy the fact that it could be the responsibility of the user to not pass files from another drive, or somehow fix the path before sending them. Node doesn't seem to care since it's not mainly for Windows. Using path relative just makes the path even worse.

johnnyreilly:
Ah - thanks @JonWallsten - I've added your change in as well.  Let's see if it's good enough to get tests passing (although right now GitHub Actions seems to be having an off day....)

alexander-akait:
> Still working on that. Although it would be nice if it was fixed in Webpack I can buy the fact that it could be the responsibility of the user to not pass files from another drive, or somehow fix the path before sending them. Node doesn't seem to care since it's not mainly for Windows. Using path relative just makes the path even worse.

It will not work if you are using linux

JonWallsten:
> > Still working on that. Although it would be nice if it was fixed in Webpack I can buy the fact that it could be the responsibility of the user to not pass files from another drive, or somehow fix the path before sending them. Node doesn't seem to care since it's not mainly for Windows. Using path relative just makes the path even worse.
> 
> It will not work if you are using linux

I will find the source of the issue with projectReferences. Seems to be the only appropriate thing to do. But would have been SO MUCH easier if it could be recreated locally. Can only use console.log to debug on Github Workflow 😐

johnnyreilly:
Doesn't look like the changes have changed test success 🥺

johnnyreilly:
I've upgraded webpack and regenerated test output

johnnyreilly:
ug I hadn't realised that execution tests were failing on Windows

```
Error: EINVAL: invalid argument, mkdir 'C:\Users\RUNNER~1\AppData\Local\Temp\_karma_webpack_789216\D:\a\ts-loader\ts-loader\test\execution-tests\3.0.1_projectReferences\lib'
```

JonWallsten:
> ug I hadn't realised that execution tests were failing on Windows
> 
> ```
> Error: EINVAL: invalid argument, mkdir 'C:\Users\RUNNER~1\AppData\Local\Temp\_karma_webpack_789216\D:\a\ts-loader\ts-loader\test\execution-tests\3.0.1_projectReferences\lib'
> ```
> 
> Oh gosh I see what you mean @JonWallsten - it fails in GitHub Actions on Windows, but it works locally. This is the _worst_

Haha, yeah, I've been fighting with that one for a while. I saw your commit, you could have moved them to C: this whole time?😂

I would however like to get to the bottom of why we emit and try write the project reference files to the output folder when using Github Actions? It doesn't happen locally, and they should just be used internally in webpack while generating the bundle, and then be discarded, right?

johnnyreilly:
> Haha, yeah, I've been fighting with that one for a while. I saw your commit, you could have moved them to C: this whole time?😂

Yes. Better late than never!

> I would however like to get to the bottom of why we emit and try write the project reference files to the output folder when using Github Actions? It doesn't happen locally, and they should just be used internally in webpack while generating the bundle, and then be discarded, right?

Yeah that sounds right. If you can get to the bottom of it that'd be amazing. I wonder if it's something to do with the caching behaviour of webpack 5, but that's just a wild guess really.

JonWallsten:
> > Haha, yeah, I've been fighting with that one for a while. I saw your commit, you could have moved them to C: this whole time?😂
> 
> Yes. Better late than never!
> 
> > I would however like to get to the bottom of why we emit and try write the project reference files to the output folder when using Github Actions? It doesn't happen locally, and they should just be used internally in webpack while generating the bundle, and then be discarded, right?
> 
> Yeah that sounds right. If you can get to the bottom of it that'd be amazing. I wonder if it's something to do with the caching behaviour of webpack 5, but that's just a wild guess really.

I'll take a new look tomorrow. It's just so time consuming when all you can do is logging stuff. Some remote debugging with Chrome Dev Tools would have been awesome! 😎

johnnyreilly:
Yeah if you could look that'd be amazing. I've got the build "green" by ignoring the following tests on Windows:

```js
const testsWithDifferentOutputOnWindows = [
    'projectReferencesOutDirWithPackageJsonAlreadyBuilt',
    'projectReferencesOutDirWithPackageJson',
    'projectReferencesOutDir',
    'projectReferencesNotBuilt_WatchApi',
    'projectReferencesNotBuilt_SemanticErrorInReference_WatchApi',
    'projectReferencesNotBuilt_SemanticErrorInReference_Composite_WatchApi',
    'projectReferencesNotBuilt_SemanticErrorInReference',
    'projectReferencesNotBuilt_ErrorInProject_WatchApi',
    'projectReferencesNotBuilt_ErrorInProject_Composite_WatchApi',
    'projectReferencesNotBuilt_ErrorInProject',
    'projectReferencesNotBuilt_Composite_WatchApi',
    'projectReferencesNotBuilt',
    'projectReferencesNoSourceMap',
    'projectReferencesMultipleDifferentInstance',
    'projectReferencesMultiple',
    'projectReferences',
    'declarationOutputWithMaps',
    'declarationOutput',
    'projectReferencesWatchRefWithTwoFilesAlreadyBuilt',
    'projectReferencesRootDirInvalidConfig',
    'projectReferencesSymLinks',
    'projectReferencesSymLinksPreserve',
    'projectReferencesSymLinksPreserve_WatchApi',
    'projectReferencesSymLinks_WatchApi',
    'projectReferencesWatch',
    'projectReferencesWatchRefWithTwoFiles',
    'projectReferencesWatchRefWithTwoFilesAlreadyBuilt_Composite_WatchApi',
    'projectReferencesWatchRefWithTwoFilesAlreadyBuilt_WatchApi',
    'projectReferencesWatch_Composite_WatchApi',
    'projectReferencesWatch_WatchApi',
    'projectReferencesRootDir',
    'projectReferencesWatchRefWithTwoFiles_WatchApi',
    'projectReferencesWatchRefWithTwoFiles_Composite_WatchApi'
];
```

As we can see, something about project references and webpack 5 is curious. With webpack 4, project references behaved the same between Linux and Windows. With webpack 5 that doesn't hold true and I don't know why.

JonWallsten:
> Yeah if you could look that'd be amazing. I've got the build "green" by ignoring the following tests on Windows:
> 
> ```js
> const testsWithDifferentOutputOnWindows = [
>     'projectReferencesOutDirWithPackageJsonAlreadyBuilt',
>     'projectReferencesOutDirWithPackageJson',
>     'projectReferencesOutDir',
>     'projectReferencesNotBuilt_WatchApi',
>     'projectReferencesNotBuilt_SemanticErrorInReference_WatchApi',
>     'projectReferencesNotBuilt_SemanticErrorInReference_Composite_WatchApi',
>     'projectReferencesNotBuilt_SemanticErrorInReference',
>     'projectReferencesNotBuilt_ErrorInProject_WatchApi',
>     'projectReferencesNotBuilt_ErrorInProject_Composite_WatchApi',
>     'projectReferencesNotBuilt_ErrorInProject',
>     'projectReferencesNotBuilt_Composite_WatchApi',
>     'projectReferencesNotBuilt',
>     'projectReferencesNoSourceMap',
>     'projectReferencesMultipleDifferentInstance',
>     'projectReferencesMultiple',
>     'projectReferences',
>     'declarationOutputWithMaps',
>     'declarationOutput',
>     'projectReferencesWatchRefWithTwoFilesAlreadyBuilt',
>     'projectReferencesRootDirInvalidConfig',
>     'projectReferencesSymLinks',
>     'projectReferencesSymLinksPreserve',
>     'projectReferencesSymLinksPreserve_WatchApi',
>     'projectReferencesSymLinks_WatchApi',
>     'projectReferencesWatch',
>     'projectReferencesWatchRefWithTwoFiles',
>     'projectReferencesWatchRefWithTwoFilesAlreadyBuilt_Composite_WatchApi',
>     'projectReferencesWatchRefWithTwoFilesAlreadyBuilt_WatchApi',
>     'projectReferencesWatch_Composite_WatchApi',
>     'projectReferencesWatch_WatchApi',
>     'projectReferencesRootDir',
>     'projectReferencesWatchRefWithTwoFiles_WatchApi',
>     'projectReferencesWatchRefWithTwoFiles_Composite_WatchApi'
> ];
> ```
> 
> As we can see, something about project references and webpack 5 is curious. With webpack 4, project references behaved the same between Linux and Windows. With webpack 5 that doesn't hold true and I don't know why.

Yeah, and not just Windows in general. Windows 2019 server it seems since that's what Github Workflow is running! I can't reproduce on Win 10. But PR is a TypeScript feature, right? Webpack shouldn't have anything to do with that?

johnnyreilly:
>  I can't reproduce on Win 10.

Well you can't now as I've made the failing tests "pass".

If you exclude one the tests above, commit that change (Since we `git clean` before running tests) and then run say:

```
Yarn comparison-tests --single-test projectReferencesRootDir
```

Then it should fail on Windows 10

JonWallsten:
> > I can't reproduce on Win 10.
> 
> Well you can't at the moment as I've made the failing tests "pass". False sense of security ☺️
> 
> If you exclude one the tests above, commit that change (Since we `git clean` before running tests) and then run say:
> 
> ```
> Yarn comparison-tests --single-test projectReferencesRootDir
> ```
> 
> Then it should fail on Windows 10

I tried on two different machines a few weeks ago with Win 10, even with the project on another drive letter. It doesn't emit the Project reference files like on GitHub, so the tests passes since it doesn't tried to write to the invalid directory you mentioned a few posts up.

johnnyreilly:
> I still have my old branch so will continue debugging on that one.

Okay cool - it may be relevant, when deving `ts-loader` on Windows, I'm doing it in the location `C:\source\ts-loader`

alexander-akait:
@johnnyreilly oh, looks problem solved, something else from me?

johnnyreilly:
Far from it alas! We've still got divergent behaviour between Linux and Windows for webpack 5. At the moment the tests that cause errors have been set to ignore for Windows. See the conversation above to see more details - but I'm not feeling too good about this!

sokra:
That's probably a problem in ts-loader. I'm on windows and running webpack and typescript test suites are completely deterministic and behave equal on windows, linux and mac. Typescript has many snapshot tests for the behavior and that was equal on windows.

Anyway let me know if I can help you in any way.

johnnyreilly:
Thanks @sokra - that's helpful.

Probably it'll be easiest if I summarise where we are, and what's puzzling.

This PR ported `ts-loader` to both build against webpack@5 and run testpacks against webpack@5.  Previously `ts-loader` has been built / run against webpack@4.  Most of the code changes associated with this change were related to interfaces.  So not runnable code.  The few exceptions were moving to use `compilation.emitAsset` here:

https://github.com/TypeStrong/ts-loader/pull/1251/files#diff-913660dd83fada2ccec7f01dd2e5b486537c5ea232ffa64f367655fd4cf72d58R393

and directly using `new webpack.WebpackError(message)` here:

https://github.com/TypeStrong/ts-loader/pull/1251/files#diff-39b2554fd18da165b59a6351b1aafff3714e2a80c1435f2de9706355b4d32351R154

The final change (and perhaps this is relevant) is here in `watchRun`:

https://github.com/TypeStrong/ts-loader/pull/1251/files#diff-ca0521bf4e14ccf353114dea5c548f8c90f9c29fb1a0e5ecd8db64cd95177275R35

Previously we had:

```ts
        if (date <= lastTime) {
          continue;
        }

        lastTimes.set(key, date);
```

And we moved to instead:

```ts
        if (
          !date ||
          date === 'ignore' ||
          (date.timestamp || date.safeTime) <= lastTime
        ) {
          continue;
        }

        lastTimes.set(key, date.timestamp || date.safeTime);
```

What we've started to notice is divergent behaviour in our comparison tests between Windows and Linux.  Not all tests; most behave the same on Windows and Linux (as has historically always been the case).  Of the ones that fail, the majority of them are project references tests, with only two exceptions:

-     `declarationOutputWithMaps`
-     `declarationOutput`

We're mystified as to why we're seeing this difference in behaviour.  I'd previously excluded the failing tests from running on Windows so I could be certain that things were running entirely correctly on Linux.  I've just reincluded them so you'll see tests start to appear red in 20 minutes or so.

To read details of the sort of differences we're seeing, take a look at the chat earlier in the discussion here: https://github.com/TypeStrong/ts-loader/pull/1251#issuecomment-781967959 with @andrewbranch 

If you could provide some insight as to why we're getting this divergent behaviour between Windows and Linux that would be amazing.  I'm somewhat out of ideas.  If you'd like to hack on this branch you'd be welcome to; I've sent you an invite.

To replicate the issue you should be able to:

```
git clone https://github.com/TypeStrong/ts-loader.git
git checkout feat/webpack-5
yarn
yarn build
yarn run comparison-tests -- --single-test projectReferencesOutDirWithPackageJsonAlreadyBuilt
```

Hopefully this all makes sense?  Bit a brain dump so please let me know if anything isn't clear!

sokra:
The problem is introduced due to some normalization logic in the test case: see https://github.com/TypeStrong/ts-loader/pull/1273

While the PR fixes the problem, I think the paths should be normalized earlier in the pipeline to make this normalization code unnecessary. Note that asset names should have only `/` as they are filenames and not paths. Only absolute paths have `\`.

johnnyreilly:
This is amazing - thanks @sokra I really appreciate it!

>  I think the paths should be normalized earlier in the pipeline to make this normalization code unnecessary. Note that asset names should have only / as they are filenames and not paths. Only absolute paths have \.

By this do you mean that ideally the fix should be in ts-loader itself rather than in the test pack? And if that was implemented then the normalisation you've implemented would no longer be necessary?

sokra:
> By this do you mean that ideally the fix should be in ts-loader itself rather than in the test pack?

yes

> And if that was implemented then the normalisation you've implemented would no longer be necessary?

Yes the whole block shouldn't be necessary.

sokra:
I guess that's a easy fix. How does declaration files from typescript become webpack assets?

JonWallsten:
@sokra : I think that would be here:
https://github.com/TypeStrong/ts-loader/blob/235747cf607cfdd79f17f4b1c95f2658eb728010/src/after-compile.ts#L388

johnnyreilly:
https://github.com/TypeStrong/ts-loader/blob/235747cf607cfdd79f17f4b1c95f2658eb728010/src/after-compile.ts#L358

JonWallsten:
> I think it might actually be this:
> 
> https://github.com/TypeStrong/ts-loader/blob/235747cf607cfdd79f17f4b1c95f2658eb728010/src/after-compile.ts#L358

My line is the end of the line for your line! ;)

johnnyreilly:
I wonder if this:

https://github.com/TypeStrong/ts-loader/blob/235747cf607cfdd79f17f4b1c95f2658eb728010/src/after-compile.ts#L391

Should actually be:

```js
.replace(/\\/g, '/');
```

?

JonWallsten:
> I wonder if this:
> 
> https://github.com/TypeStrong/ts-loader/blob/235747cf607cfdd79f17f4b1c95f2658eb728010/src/after-compile.ts#L391
> 
> Should actually be:
> 
> ```js
> .replace(/\\/g, '/');
> ```
> 
> ?

Yes! The code above replaces slash with slash. We probably should have noticed that before! Haha.

johnnyreilly:
Okay, so applying the `.replace(/\\/g, '/')` fix, @sokra's remedy no longer is required for tests to pass! Can't believe one wrong character caused so much pain!

JonWallsten:
> Okay, so applying the `.replace(/\\/g, '/')` fix, @sokra's remedy no longer is required for tests to pass! Can't believe one wrong character caused so much pain!

That's great! Does it also fix the projektReference issue? Or is it just the difference in output?

johnnyreilly:
> Does it also fix the projektReference issue?

Do you mean the failing project references comparison tests? Yes they're fixed.

JonWallsten:
I don't really understand how it can fix the issues with the projectReferences. Even if the path was changed from D:\ to D:/ it should still cause issues. And it should still output the files. I guess we just have to accept that something is going on under the hood in TypeScript or Webpack that we don't understand.
Feels like I wasted two days chasing down a rabbit hole though! 😅

JonWallsten:
I will try it out in our project tomorrow!

johnnyreilly:
I encourage you to satisfy your curiosity @JonWallsten!

@appzuka if you're still keen to pull out the experimental watch API, feel free!

Otherwise the port to webpack 5 is complete I think.

appzuka:
@johnnyreilly, when I looked I got the impression there is quite a lot of code that seems to be connected with the watch API and could be ripped out but I don't have enough time to do this carefully and I would not want to break the code after all the work you have put in. Perhaps I'll start a separate branch when I have time to dedicate to the task.

johnnyreilly:
Okay that's cool - @appzuka

test/comparison-tests/aliasResolution/expectedOutput-4.1/bundle.js
```diff
@@ -1,99 +1,21 @@
...
-/******/ 	__webpack_require__.p = "";
-/******/
-/******/
-/******/ 	// Load entry module and return exports
-/******/ 	return __webpack_require__(__webpack_require__.s = "./app.ts");
-/******/ })
-/************************************************************************/
-/******/ ({
+/*
+ * ATTENTION: The "eval" devtool has been used (maybe by default in mode: "development").
```
JonWallsten, line 2: @johnnyreilly Is this deliberate?

johnnyreilly: The code comes from webpack and our comparison tests run in development mode.

johnnyreilly: Ultimately, whatever webpack pumps out will end up in our comparison tests

JonWallsten: I get that. The question was more if using a devtool in the builds are deliberate and something you want?

johnnyreilly: Sorry I missed the reply - yes I'm happy with using `development` mode in the tests. This mirrors the typical webpack usecase of developing an application with a dev server running locally in watch mode.

johnnyreilly:
Okay I think it's nearly time to ship - see update at the top of this PR.

JonWallsten:
@johnnyreilly Sorry for being MIA. Swamped at work. I'll try this with out project today!

johnnyreilly:
Thanks @JonWallsten!

johnnyreilly:
Did you have a chance to try it out @JonWallsten?  All feedback welcome!

JonWallsten:
> Did you have a chance to try it out @JonWallsten? All feedback welcome!

Sorry! Got stuck with build issues in our project yesterday. I will try it out today instead.

JonWallsten:
I have tried to build a prod build, dev builds with watch and changed a new files. Seems to work fine. Not slower nor faster then previously. No errors.

JonWallsten:
I guess we will get additional feedback once released! :)

appzuka:
@johnnyreilly , @JonWallsten,  I don't believe I have contributed to this branch but I built a couple of my projects with the latest feat/webpack-5 branch today. They worked fine and I did not notice any performance or size issues.  Well done.

johnnyreilly:
Thanks @appzuka

src/instances.ts
```diff
@@ -316,51 +304,39 @@ function getExistingSolutionBuilderHost(key: FilePathKey) {
...
-      loader._compiler.hooks.compilation.tap('ts-loader', makeAssetsCallback);
+  // For future calls in watch mode we need to watch for a new compilation and add the hook
+  loader._compiler.hooks.compilation.tap('ts-loader', makeAssetsCallback);
 
-      // It may be better to add assets at the processAssets stage (https://webpack.js.org/api/compilation-hooks/#processassets)
-      // This requires Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL, which does not exist in webpack4
-      // Consider changing this when ts-loader is built using webpack5
-    };
+  // It may be better to add assets at the processAssets stage (https://webpack.js.org/api/compilation-hooks/#processassets)
+  // This requires Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL, which does not exist in webpack4
```
JonWallsten, line 334: @johnnyreilly: I forgot about this. I think it was part of my PR when I sidetracked with the GitHub Workflow thing.

johnnyreilly:
Cool - I'll finish writing up the blog post and we'll ship ⛴️

src/instances.ts
JonWallsten: Let's fix that in a minor later!

JonWallsten:
@johnnyreilly: Woho! Super good work @johnnyreilly! 🏆Sorry I couldn't be more helpful the last couple of weeks!

johnnyreilly:
Thanks @JonWallsten - I appreciate the help that you gave!

PR #1252, opened 2021-02-05
Minor improvements

JonWallsten:
Updated webpack-cli
Added types for Webpack (Do we want this?)
Added skipLibCheck to supress errors caused by webpack dependencies (needed for now due to https://github.com/webpack/webpack/issues/12185, not sure if we should verify external packages typings though?)
Implemented compilation.emitAssets for additional resources
Replaced afterProcessAssets with processAssets hook

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "8.0.15",
+  "version": "9.0.0",
   "description": "TypeScript loader for webpack",
   "main": "index.js",
   "types": "dist",
@@ -96,7 +96,7 @@
     "rimraf": "^2.6.2",
     "typescript": "^4.0.0",
     "webpack": "^5.20.0",
-    "webpack-cli": "^3.1.1"
+    "webpack-cli": "^4.5.0"
   },
   "peerDependencies": {
     "typescript": "*",
--- a/src/instances.ts
+++ b/src/instances.ts
@@ -343,11 +343,11 @@ const addAssetHooks = !!webpack.version!.match(/^4.*/)
       // compilation is actually of type webpack.Compilation, but afterProcessAssets
       // only exists in webpack5 and at the time of writing ts-loader is built using webpack4
       const makeAssetsCallback = (compilation: any) => {
-        compilation.hooks.afterProcessAssets.tap('ts-loader', () =>
+        compilation.hooks.processAssets.tap({ name: 'ts-loader', stage: webpack.Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL }, (_assets: webpack.Asset[]) => {
           cachedMakeAfterCompile(compilation, () => {
             return null;
-          })
-        );
+          });
+        });
       };
 
       // We need to add the hook above for each run.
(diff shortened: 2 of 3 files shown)


JonWallsten:
@johnnyreilly

src/instances.ts
```diff
@@ -343,11 +343,11 @@ const addAssetHooks = !!webpack.version!.match(/^4.*/)
       // compilation is actually of type webpack.Compilation, but afterProcessAssets
       // only exists in webpack5 and at the time of writing ts-loader is built using webpack4
       const makeAssetsCallback = (compilation: any) => {
-        compilation.hooks.afterProcessAssets.tap('ts-loader', () =>
+        compilation.hooks.processAssets.tap({ name: 'ts-loader', stage: webpack.Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL }, (_assets: webpack.Asset[]) => {
```
JonWallsten, line 346: I took a gamle on the webpack.Asset. Guess I need to double check with the debugger.

JonWallsten: Not even close 😅
image

JonWallsten: @johnnyreilly: Since Webpack doesn't provide types for everything, are we just supposed to add what we know and improve/change later if we find more information?
Right now I know that it's a `Record<string, CachedSource>`

But could potentially be other sources as well:
image

johnnyreilly: > Since Webpack doesn't provide types for everything, 

It doesn't? What's missing? It's built with TypeScript (albeit derived from JSDoc) so I'd imagined it all should be there

package.json
```diff
@@ -65,6 +65,7 @@
     "@types/semver": "^7.3.4",
     "@typescript-eslint/eslint-plugin": "^4.0.0",
     "@typescript-eslint/parser": "^4.0.0",
+    "@types/webpack": "^4.41.26",
```
johnnyreilly, line 68: Why include `@types/webpack`? That's webpack 4 - don't we want to remove this?

JonWallsten: Hmm, I had a hard time locating the types in webpack, so wasn't sure if they included types for everything yet. I'll take another look on Monday. I also assumed that most of the types for 4.4x.x will be the same for Webpack 5. Most things hasn't changed. Have a nice weekend!

johnnyreilly: Yeah, either way we should not be depending on `@types/webpack` as it will clash with the types shipped with webpack 5. Have a nice weekend!

JonWallsten: Totally agree! I jumped to some conclusions there. Good catch!

JonWallsten: @johnnyreilly: Reverted those changes:
webpack.Loader.LoaderContext doesn't seem to exist.
There seems to be limited typings for Loaders. They use `any` internally. 
Do you want to add custom typings or not use them at all?

image
image

johnnyreilly: There's a missing piece. Somewhere out there is a doc that detailed the changes to webpacks APIs between 4 and 5. We need to track it down

johnnyreilly:
I've lifted the changes directly as this wasn't mergable - thanks!

PR #1255, opened 2021-02-10
Include standard webpack start/end locations in emitted errors

### LINKED ISSUE #1199: Errors/warnings added by ts-loader do not respect webpack API
I'm processing webpack errors/warnings in the compilation results.

### Expected Behaviour

The errors and warnings reported by ts-loader through webpack should implement the `WebpackError` interface.
Starting from webpack 5 the interface is part of webpack's official typings.

### Actual Behaviour

Instead of including the `loc` property with start/end positions of the error, the object added by ts-loader includes a `location` property with just the starting position. This is easily seen in the webpack `stats` object produced by the compilation.

```javascript
stats.compilation.errors; // each error misses the 'loc' property
stats.compilation.warnings; // each warning misses the 'loc' property
```

Probably as a consequence of this, when I transform the stats into JSON, there is no location information for the errors.

```javascript
const json = stats.toJson({ all: false, errors: true, warnings: true });
json.errors; // 'loc' missing in errors
json.warnings; // 'loc' missing in warnings
```

lorenzodallavecchia:
This PR adds the `loc` property to all errors coming from TypeScript diagnostics.
The property is the webpack "API" way of indicating the precise location of an error, including its _end_ location.
In this way, the errors produced by `ts-loader` can be processed from the webpack `Stats` object, with no knowledge of `ts-loader`.

The previous location property is still available, so the change is backward-compatible.

Webpack will print its own error location string in the string output, in addition to the line and character that `ts-loader` already includes in the error message. All comparison tests are updated to expect the new output.

Closes #1199

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## v8.0.17
+* [Included correct webpack source location in emitted errors](https://github.com/TypeStrong/ts-loader/issues/1199) - thanks @lorenzodallavecchia
+
 ## v8.0.16
 * [Re-Fixed missing errors in watch mode in webpack5](https://github.com/TypeStrong/ts-loader/issues/1204) - thanks @appzuka
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "8.0.16",
+  "version": "8.0.17",
   "description": "TypeScript loader for webpack",
   "main": "index.js",
   "types": "dist",
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -13,12 +13,25 @@ export interface ErrorInfo {
   context: string;
 }
 
-export type FileLocation = { line: number; character: number };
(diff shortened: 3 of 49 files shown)


johnnyreilly:
This looks good - thanks for contributing!

Could you update the `package.json` and the `CHANGELOG.md` please? I'll do another review later, but I think this is probably good to ship.

lorenzodallavecchia:
> Could you update the package.json and the CHANGELOG.md please?

Done!

johnnyreilly:
Keep getting a failing test.. rerunning

lorenzodallavecchia:
@johnnyreilly yeash, I was thinking it was a known instability problem.
By looking at the details it seems that the comparison tests on Ubuntu are being aborted after 15 minutes. Maybe they are taking too much time? On Windows the complete after 20 minutes.

johnnyreilly:
Passing now - go figure

johnnyreilly:
https://github.com/TypeStrong/ts-loader/releases/tag/v8.0.17 - released!

PR #1259, opened 2021-03-02
feat: remove top-level typescript import statements

ulivz:
## Background

In some deployment systems, the builder and source code are placed in different paths, and we also expect the typescript's version to be completely controlled by the user, if the builder does not depend on `typescript`, then we will get error containing `cannot found "typescript"` caused by the top-level typescript import statement at `ts-loader`, In other words, **_we SHOULD respect the resolved compiler_**.

Prior to TypeScript 3.8, you can import a type using import. With TypeScript 3.8, you can import a type using the import statement, or using `import type`, and import type is always guaranteed to be removed from your JavaScript, this project has stopped testing TypeScript 3.6 and 3.7 at [953358e](https://github.com/TypeStrong/ts-loader/commit/953358eb1b090dd9add0824c4746220b0ebe56c1), so usage of `import type` is allowed here.

This pull request replaced all remove top-level typescript `import` statements with `import type`, and leverage constants from resolved compiler.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## v8.1.0
+* [feat: remove top-level typescript import statements](https://github.com/TypeStrong/ts-loader/pull/1259) - thanks @ulivz
+
 ## v8.0.18
 * [Perf: Optimize fileExists callback path](https://github.com/TypeStrong/ts-loader/issues/1266) - thanks @berickson1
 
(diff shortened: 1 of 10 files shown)


ulivz:
@johnnyreilly Could you help review this pull request? Thanks

johnnyreilly:
Sorry about the delay - have been busy.  I like this change.  Thanks for raising it.

I've just hit the update branch button.  Please could you increment the version in the `package.json` and add an entry to the `CHANGELOG.md`?

johnnyreilly:
Let's 🚢

johnnyreilly:
Releasing with https://github.com/TypeStrong/ts-loader/releases/tag/v8.1.0

PR #1266, opened 2021-03-10
Optimize fileExists callback path

berickson1:
Instead of looking up filePathKey twice, cache in in the local scope. This isn't a hugely expensive lookup path, but is a hot path on large project recompiles (especially after caches are blown away). Profiling shows that even the Map lookup shows non-trivial cache lookup aggregate time. (Less than a non-cached implementation though :) )
Screen Shot 2021-03-10 at 1 32 07 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## v8.0.18
+* [Perf: Optimize fileExists callback path](https://github.com/TypeStrong/ts-loader/issues/1266) - thanks @berickson1
+
 ## v8.0.17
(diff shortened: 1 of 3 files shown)


johnnyreilly:
looks good!  Can you update the `CHANGELOG.md` and `package.json` too please and we'll ship.

berickson1:
@johnnyreilly - any idea what's up with the travis test failure? It was clean before I bumped package.json & I don't see any reason why my change should cause the listed failure

johnnyreilly:
travis has been flaky for a while and I want to remove it entirely - can't find the relevant setting in GitHub alas.  It's intentionally not required so don''t worry.

johnnyreilly:
shipping now https://github.com/TypeStrong/ts-loader/releases/tag/v8.0.18

PR #1270, opened 2021-03-12
Fixes #1260 by generating declaration files for js files if allowJs is set to true

### LINKED ISSUE #1260: No .d.ts Declaration Files for JS Files Are Emitted (allowJs: true)
When setting `allowJs: true` in `tsconfig.json`, typescript files can import javascript files.
Typescript is also clever about the types used in javascript and infers as much as possible.
Such inferred javascript types might even be exported by typescript files and become part of the library API.

### Expected Behaviour

`ts-loader` should emit declaration files for javascript files when `allowJs` is set to `true` and `ts-loader` is configured to handle javascript files.

### Actual Behaviour

`ts-loader` does not emit declaration files for javascript files that are imported in typescript files. This results in broken declaration files.

1. clone https://github.com/hediet/webpack-allowjs-declaration-issue-repro
2. run `yarn`
3. run `yarn build`
4. Notice this error (no declaration files for javascript):

image

https://github.com/hediet/webpack-allowjs-declaration-issue-repro

Btw., thanks for creating and maintaining ts-loader!

hediet:


johnnyreilly:
Thanks for the PR!  It looks like you've a failing test.  On Windows you're generating:

```
      -      TS2507: Type 'typeof import("sub/dep")' is not a constructor function type.
```

But on Linux you're getting this:

```
      +      TS2507: Type 'typeof import("/declarationOutputAllowJs/sub/dep")' is not a constructor function type.
```

hediet:
> Thanks for the PR! It looks like you've a failing test

Do you have an idea what causes this? Seems to be unrelated of my change and an issue with the test itself, especially as it is OS dependent.

johnnyreilly:
The nature of the difference between the two output is `"sub/dep"` and `"/declarationOutputAllowJs/sub/dep"`.  Which OS did you use when you wrote this test? It could be worth comparing your test to the existing `declarationOutput` one and see if you can spot a difference.

This may be surfacing an actual issue though.  If memory serves there have been historic issues with paths and declaration output. See:

https://github.com/TypeStrong/ts-loader/pull/822

cc @JonWallsten

hediet:
I'm a Windows user!

johnnyreilly:
What do you get if you generate the test output in the devcontainer?

JonWallsten:
I'll have a look today. Since the difference is between systems I guess the relative path is the culprit?

johnnyreilly:
Seems plausible - I wonder if the `replace(/\\/g, '/')` in afterCompile would be the fix?

JonWallsten:
> Seems plausible - I wonder if the `replace(/\\/g, '/')` in afterCompile would be the fix?

My initial thought was that TS handles js files differently, because it already works with TS files. But since it works on one system and not the other I was also thinking about that fix we did! We should probably try that!

johnnyreilly:
Try all the things!

lastthyme (approved):

johnnyreilly:
I think this was done prior to the webpack 5 port. Could be worth regenerating comparison test output and see if the issue is resolved by the webpack 5 work

as-zlynn-philipps:
Dang, this should be merged. Whatever happened with the failing test?

johnnyreilly:
I'm not sure - it all went quiet. If you'd like to pick this up and work on it I'll happily look at a PR!

hediet:
Sorry, I moved on from [the project I needed this for](https://github.com/hediet/pdf.js-viewer) and then got very busy.

For this PR, the neccessary fix is in `src/after-compile.ts`. Everything else is about the tests.

johnnyreilly:
Awesome - I'm happy to assist with the test if someone else wants to pick this up

mvilliger:
https://github.com/TypeStrong/ts-loader/issues/1260 has been fixed in [PR 1483](https://github.com/TypeStrong/ts-loader/pull/1483) and published with release 9.3.1.
Therefore, this PR is obsolete.

hediet:
Awesome, thanks!

Issue #1025, opened 2019-10-20
Custom transformers not working for referenced projects


adam-stanek:
Hi. I am trying to build a mono-repo project consisting from webpack demo application and separately published library. The library is defined as referenced typescript project (`projectReferences: true`).
In order to build the library correctly I need custom transformer ([@zerollup/ts-transform-paths](https://www.npmjs.com/package/@zerollup/ts-transform-paths) to transform paths). It seems to be ignored when building the library.

### Expected Behaviour

Library code gets emitted while being properly transformed.

### Actual Behaviour

Library code gets emitted but no transformation is applied. It is worth mentioning that `@zerollup/ts-transform-paths` run through `ttypescript` is working fine.

### Steps to Reproduce the Problem

*ts-loader options*

```
{
  projectReferences: true,
  getCustomTransformers: require.resolve('./webpack.tsTransformers'),
}
```

*./webpack.tsTransformers.js*

```
const createTsTransformPath = require('@zerollup/ts-transform-paths')

function getCustomTransformers(program) {
  const tsTransformPath = createTsTransformPath(program)

  return {
    before: [tsTransformPath.before],
    // after: [tsTransformPath.afterDeclarations],
  }
}

module.exports = getCustomTransformers
```

### Location of a Minimal Repository that Demonstrates the Issue.

https://github.com/adam-stanek/chobot/commit/8bf5dd2c69ce0437494d55ac7e9524b0f60c6469

unstubbable:
This is also true for [typescript-plugin-styled-components](https://github.com/Igorbek/typescript-plugin-styled-components), which is mentioned in the REAMDE as an example for `getCustomTransformers `. I'd like this issue to be reopened.

PR #1274, opened 2021-03-27
buildSolutionReferences with instance transformers

gasnier:
Fixes https://github.com/TypeStrong/ts-loader/issues/1025

I first tried a solution inspired from https://stackoverflow.com/questions/62026189/typescript-custom-transformers-with-ts-createwatchprogram/62132983#62132983
Where I used getNextInvalidatedProject from compiler api that is defined here :
https://github.com/microsoft/TypeScript/blob/master/src/compiler/tsbuildPublic.ts
This worked for standalone builds but was not working in watch mode.
In watch mode, it is SolutionBuilder that handles the emit and today there is no way to pass customTransformers to SolutionBuilder.

The final solution overrides the emit function and is based on https://github.com/nonara/ts-patch


johnnyreilly:
Heads up, we've just renamed from `master` to `main` - https://twitter.com/johnny_reilly/status/1376156935414820869

This will have updated this PR but not your local branch.

gasnier:
@johnnyreilly 
This PR is ready for review

johnnyreilly:
Thanks - I'll try and take a look this weekend

johnnyreilly:
Sorry - I forgot about this. Just updated the branch. Will take a look.

johnnyreilly:
Is `transformers` on instance now unused? If so, should it be removed?

https://github.com/gasnier/ts-loader/blob/6bb61d4088fcae03feca700e59f6e96c77f6244c/src/interfaces.ts#L212

src/transformers.ts
```diff
@@ -0,0 +1,140 @@
...
+}
+
+/**
+ * Patches typescript program to pass custom transformers to the program.emit function.
+ * In 2021, this is the only way since it is not possible to pass transformers to SolutionBuilder
+ * SolutionBuilder is used when watch is enabled
+ */
+let patched = false;
+export function installTransformers(loaderOptions: LoaderOptions) {
+  if (!loaderOptions.getCustomTransformers) {
```
johnnyreilly, line 79:
```suggestion
  if (loaderOptions.getCustomTransformers === undefined) {
```

src/transformers.ts
```diff
@@ -0,0 +1,140 @@
...
+ * Patches typescript program to pass custom transformers to the program.emit function.
+ * In 2021, this is the only way since it is not possible to pass transformers to SolutionBuilder
+ * SolutionBuilder is used when watch is enabled
+ */
+let patched = false;
+export function installTransformers(loaderOptions: LoaderOptions) {
+  if (loaderOptions.getCustomTransformers === undefined) {
+    return; //no need to patch if there is no transformer
+  }
+  if (patched) {
```
johnnyreilly, line 82:
```suggestion
  if (patched === true) {
```

johnnyreilly:
Looking good I think. Two things:

- this PR is effectively working around the fact there is no way to pass customTransformers to SolutionBuilder. It would be good to raise that absence against the TypeScript project itself so they are aware of this. Perhaps in future a different approach could be used if something is implemented in TypeScript. Could you either link to the issue if it already exists, or raise it and link to it? Thanks. cc @sheetalkamat 
- could you update `package.json` and `CHANGELOG.md`? Thanks!

src/transformers.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    function newEmit(
+      targetSourceFile?: ts.SourceFile,
+      writeFile?: ts.WriteFileCallback,
+      cancellationToken?: ts.CancellationToken,
+      emitOnlyDtsFiles?: boolean,
+      customTransformers?: ts.CustomTransformers
+    ): ts.EmitResult {
+      /* Invoke TS emit */
+      const result: ts.EmitResult = originalEmit(
```
sheetalkamat, line 123: This will have some bad effects .. emit is used for emitting dts files to determine shape of the file as well and take internal parameters and that signature is internal to typescript .. also I don’t think you want transformers for that reason .. 

I have been thinking about your issue and was wondering if build could take getTransformers as function that returns transformers based on project path.. haven’t had chance to investigate that part yet but this approach is definitely not correct

sheetalkamat (changes requested):

src/transformers.ts
johnnyreilly: Oh thanks for that @sheetalkamat - very helpful!

gasnier: Yes, if we can pass customTransformers to build in SolutionBuilder, this solves the problem.
the done function on invalidatedProject already has the customTransformers parameter.
image

A function by projet path is ok, anyway in ts-loader there is only one global configuration so we would always pass the same transformers. A good solution would be to do like ttypescript (https://github.com/cevek/ttypescript) and read the configuration from tsconfig.json

sheetalkamat: https://github.com/microsoft/TypeScript/pull/43984 add the typescript side of the API

johnnyreilly:
Should we close this now?

gasnier:
Yes, I close it, I didn't know Typescript 4.3.1 was out...
Thanks @sheetalkamat

I will create another issue / PR to solve this using SolutionBuilder.

PR #1286, opened 2021-04-19
fix: Use correct stage for adding assets during compilation

JonWallsten:


JonWallsten:
@johnnyreilly: I extracted the fix from my previous webpack 5 branch and added on top of the v5. Seems to work fine!

src/instances.ts
```diff
@@ -314,10 +314,16 @@ function addAssetHooks(loader: WebpackLoaderContext, instance: TSInstance) {
...
 
   const makeAssetsCallback = (compilation: webpack.Compilation) => {
-    compilation.hooks.afterProcessAssets.tap('ts-loader', () =>
-      cachedMakeAfterCompile(compilation, () => {
-        return null;
-      })
+    compilation.hooks.processAssets.tap(
+      {
+        name: 'ts-loader',
+        stage: webpack.Compilation.PROCESS_ASSETS_STAGE_ADDITIONAL,
```
johnnyreilly, line 320: What's the significance of the specified stage @JonWallsten?

johnnyreilly: "add additional assets to the compilation." I guess?

JonWallsten: This is just the preferred way of doing it in Webpack 5. @alexander-akait metioned it in the beginning of the PR. It shouldn't affect anything at all. It's just more future proof.

alexander-akait: Yep, it is right usage, we can implement cache here in future

johnnyreilly:
Cool - do you want to update the `CHANGELOG.md` and `package.json` ?

JonWallsten:
> Cool - do you want to update the `CHANGELOG.md` and `package.json` ?

Done!

johnnyreilly:
https://github.com/TypeStrong/ts-loader/releases/tag/v9.0.1

alexander-akait:
@JonWallsten We can avoid using `loader-utils` https://github.com/TypeStrong/ts-loader/blob/main/src/index.ts#L2, it will be deprecated in near future:
https://github.com/TypeStrong/ts-loader/blob/main/src/index.ts#L177:
```
// No need fallback on `{}`, when not option it will be `{`}
const options = this.getOptions(schema); // schema is optional 
```

https://github.com/TypeStrong/ts-loader/blob/main/src/index.ts#L649, just change on:
```
loaderContext.remainingRequest
```

Because webpack v5 is minimum supported version, no breaking changes here, just less code and less deps

johnnyreilly:
Hey @alexander-akait 

Are you saying that

```ts
  const loaderOptions =
    loaderUtils.getOptions<LoaderOptions>(loaderContext) ||
    ({} as LoaderOptions);
```

can become

```ts
const loaderOptions = this.getOptions(); // schema is optional 
```

?

alexander-akait:
Yes

JonWallsten:
I can do a PR if you'd like @johnnyreilly!

hassenc:
I'm getting a `TypeError: loaderContext.getOptions is not a function` on start, after updating to 9.0.2
Is it linked to this update ?

PR #1290, opened 2021-04-22
fix: times is not iterable bug

elf-mouse:


johnnyreilly:
Under what circumstances would `compiler.fileTimestamps` not be present?

elf-mouse:
For webpack middlewares usage:

```js
const webpack = require('webpack');
const middleware = require('webpack-dev-middleware');
const compiler = webpack({
  // webpack options
});

console.log(compiler.fileTimestamps);
// Output:
// webpack@4: Map(0) {}
// webpack@5: undefined
```

johnnyreilly:
Interesting - has `compiler.fileTimestamps` been replaced by something else in webpack

sokra:
Yes with FileSystemInfo and the snapshot system.

Avoid comparing timestamps directly and use a snapshot instead

johnnyreilly:
> Yes with FileSystemInfo and the snapshot system.
>
> Avoid comparing timestamps directly and use a snapshot instead

@elf-mouse would you care to take a look at @sokra's suggestion around migrating to use these APIs instead?

This feels like a better approach.

elf-mouse:
Hi @johnnyreilly ,

It is recommended to update the patch first to fix this bug to ensure online security.

I would migrate to use new APIs when I have time later.

THX~

johnnyreilly:
Thanks @elf-mouse - but I'm not very keen on the current workaround; going for the new APIs feels like the better approach.

kaytrance:
do I understand correctly that the only workaround for now is not to use webpack 5?

johnnyreilly:
If someone would like to create a minimal reproduction repo that illustrates this issue, it might help someone work on this.

user84441:
> If someone would like to create a minimal reproduction repo that illustrates this issue, it might help people to work on this.

There is a simple example:
https://stackoverflow.com/questions/67699117/error-when-trying-to-run-dev-server-react-ssr

Any progress in fixing this issue?

imdavidmin:
This error seems to have popped out of nowhere for me. Have been using Next / webpack 5 for a while and previously compiled fine.

johnnyreilly:
If someone would like to submit a PR which makes use of the advice provided by @sokra we'll happily take a look:

> Yes with FileSystemInfo and the snapshot system.
>
> Avoid comparing timestamps directly and use a snapshot instead

PR #1287, opened 2021-04-20
Use caches for module resolution and type reference directives when using compiler default functions

sheetalkamat:
Uses #microsoft/TypeScript#43700

PR #1291, opened 2021-04-22
Backport #1287 to v8

JasonKleban:
Backport "Use caches for module resolution and type reference directives when using compiler default functions" #1287 to v8

@johnnyreilly is this the right branch for any version 8 patches?

CC @sheetalkamat


johnnyreilly:
> @johnnyreilly is this the right branch for any version 8 patches?

It is yes - I've never tried publishing from a branch that isn't main using the GitHub Actions workflow.  Wonder if it works....

johnnyreilly:
Failing comparison tests - you may want to regenerate the output for the failing tests. See https://github.com/TypeStrong/ts-loader/tree/main/test/comparison-tests#regenerating-test-data

johnnyreilly:
I think you may have regenerated all of the test output, rather than just the failing test cases?

JasonKleban:
I regenerated the tests, but I don't think they'll succeed on the server check happening now.  There are trivial differences and root path differences.

image

johnnyreilly:
Yeah - I think there's something wrong with your backport

JasonKleban:
@johnnyreilly gonna need some help please.  I don't know what I'm looking at.

johnnyreilly:
So to regenerate the failing test I did this: `yarn run comparison-tests -- --save-output --single-test projectReferencesSymLinks`

JasonKleban:
Yes, I ran exactly that (although I also ran "instance" since that one was also listed as failing, from my reading).

johnnyreilly:
I did it in the context of `ts-loader`s devcontainer (but that shouldn't matter).  Looks like bad things are happening in GitHub Actions at the moment; seeing random failures

johnnyreilly:
`yarn run comparison-tests -- --single-test projectReferencesSymLinks_WatchApi --save-output`

JasonKleban:
good, right?

johnnyreilly:
Yup - good.  This is where we find out if the automation works for the non-`main` branch...

johnnyreilly:
https://github.com/TypeStrong/ts-loader/actions/runs/778506165

johnnyreilly:
https://www.npmjs.com/package/ts-loader/v/8.2.0

PR #1293, opened 2021-04-25
UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'fileName' of undefined

gasnier:
When using webpack devserver I often get an Unhandled promise rejection when a file is removed (generally because I switch to another git branch)
I get UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'fileName' of undefined
    at provideErrorsToWebpack (after-compile.js:128:29)

 (I assume it may also happen while simply using watch mode)

Without the provided fix, I need to stop and restart the devserver to solve the problem

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v9.1.2
+
+* [Fix removed files handling in watch mode](https://github.com/TypeStrong/ts-loader/pull/1293) - thanks @gasnier
+
 ## v9.1.1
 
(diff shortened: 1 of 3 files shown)


johnnyreilly:
Thanks! Do you want to add an entry to the `CHANGELOG.md` and increment the `package.json`?

gasnier:
> Thanks! Do you want to add an entry to the `CHANGELOG.md` and increment the `package.json`?

Done

johnnyreilly:
Shipping with https://github.com/TypeStrong/ts-loader/releases/tag/v9.1.2

PR #1296, opened 2021-05-02
Update to babel7

sa4k5hi:
Updated to babel7 in devDependencies package.json and files both in execution and comparison tests


johnnyreilly:
The tests seem to be failing?

lastthyme (approved):

sa4k5hi:
It seems that 2 execution test- 2.1.4_babel-react and 2.1.4_react are failing. But that is strange, because everything is working fine in local. Though I'm not sure, but I think this problem is not because of babel.

johnnyreilly:
Seems like comparison tests are failing also? Will requeue just to be sure

sa4k5hi:
Files only from babel-es6resolveParent, babel-issue81 and babel-issue92 comparison test has been edited. Rest is all same for it.

sa4k5hi:
For all comparison test, it is the error-
`AssertionError [ERR_ASSERTION]: The expected output does not exist; there is nothing to compare against! Has the expected output been created?`

It seems like it is not able to identify the expected output file but is able to give actual output

johnnyreilly:
If you can get execution tests working, I'll take a look at the comparison tests

johnnyreilly:
Closing due to lack of activity

PR #1317, opened 2021-05-12
Fixed generating cache for loading options to handle objects

### LINKED ISSUE #1316: Incorrect instance hash for different compiler options
Let's say you have 2 packages, and one of them you want to build with target=es5, another one with target=es6. If you set up ts-loader options to override compiler options (in loader options, not in tsconfig) it wont work properly.

There is a test https://github.com/TypeStrong/ts-loader/tree/main/test/execution-tests/loaderOptions for similar situation, but it doesn't cover this case. Let's say instead of overriding the only [`getCustomTransformers`](https://github.com/TypeStrong/ts-loader/blob/94480fbdf2db2c4060706ee8bf7d51294647f4da/test/execution-tests/loaderOptions/webpack.config.js#L25-L27) option you want to override something in `compilerOptions`. In this case [this code](https://github.com/TypeStrong/ts-loader/blob/94480fbdf2db2c4060706ee8bf7d51294647f4da/src/index.ts#L165]) will return `[object Object]` for both _different_ objects (different ref and a value itself) which causes the issue (also this code doesn't cover a case when a value is `false`, thus if a value is falsy but enabled by default it will also failed).

Not too sure about the fix here. I see different approaches to solve it:
1. Use `JSON.stringify` for non-functions and `value.toString()` otherwise
2. Allow to provide a `instanceKey` or something like that so people can declare their own cache key and refer to them in different places

Maybe you have thoughts how to solve it in better way. Let me know you what do you think. I'm ready to make a fix for this.

timocov:
Fixes #1316

### DIFF
--- a/test/comparison-tests/errorFormatter/expectedOutput-4.2/output.txt
+++ b/test/comparison-tests/errorFormatter/expectedOutput-4.2/output.txt
@@ -5,11 +5,11 @@ asset bundle.js 2.6 KiB [emitted] (name: main)
 ERROR in app.ts
 ./app.ts 1:29-53
 Does not compute.... [1m[31mcode: 2307,severity: error,content: Cannot find module 'components/myComponent' or its corresponding type declarations.,file: app.ts,line: 1,character: 30,context: .test/errorFormatter[39m[22m
-ts-loader-default_6241146af0584e95
+ts-loader-default_85b0565984bbe8dd
 
 ERROR in app.ts
 ./app.ts 2:30-55
 Does not compute.... [1m[31mcode: 2307,severity: error,content: Cannot find module 'components/myComponent2' or its corresponding type declarations.,file: app.ts,line: 2,character: 31,context: .test/errorFormatter[39m[22m
-ts-loader-default_6241146af0584e95
+ts-loader-default_85b0565984bbe8dd
 
 webpack compiled with 2 errors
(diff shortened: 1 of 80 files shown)


johnnyreilly:
This looks potentially like a reasonable approach. I think this has broken some of the comparison tests - please could you take a look?

timocov:
@johnnyreilly I think it should be better now. Let me what if you want to change the check somehow.

test/comparison-tests/errorFormatter/expectedOutput-4.2/output.txt
```diff
@@ -4,12 +4,12 @@ asset bundle.js 2.6 KiB [emitted] (name: main)
 
 ERROR in app.ts
 ./app.ts 1:29-53
-Does not compute.... [1m[31mcode: 2307,severity: error,content: Cannot find module 'components/myComponent' or its corresponding type declarations.,file: app.ts,line: 1,character: 30,context: .test/errorFormatter[39m[22m
-ts-loader-default_6241146af0584e95
+Does not compute.... [1m[31mcode: 2307,severity: error,content: Cannot find module 'components/myComponent' or its corresponding type declarations.,file: app.ts,line: 1,character: 30,context: .test\errorFormatter[39m[22m
```
timocov, line 7: oh, missed these changes, I'll update it later

timocov:
@johnnyreilly now it should be fine I guess 🙂

johnnyreilly:
Do you want to update the `package.json` and `CHANGELOG.md`?

Issue #1323, opened 2021-05-21
Limit webpack version in peerDependencies


afdev82:
I think that the `webpack` version in the `peerDependencies` section of `package.json` should be restricted to only v5, since from v9 `ts-loader` is not compatible anymore with `webpack` v4.

https://github.com/TypeStrong/ts-loader/blob/613027f1cd6d5e68088d77617ce849d6d64aa479/package.json#L102

Or am I wrong?
I tried to update `ts-loader` while still using `webpack` v4 and `yarn` didn't complain, but I got the error in the console:

```
Module build failed: TypeError: loaderUtils.getOptions is not a function
```
And found the explanation in [this issue](#595)

johnnyreilly:
Not a bad idea - fancy raising a PR?

afdev82:
Yes, I will do it

PR #1324, opened 2021-05-21
Add webpack minimum version in peerDependencies

afdev82:
I added the minimum webpack version in the peerDependencies section of package.json.
See also #1323

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,6 +99,6 @@
   },
   "peerDependencies": {
     "typescript": "*",
-    "webpack": "*"
+    "webpack": "^5.0.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "9.2.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -99,6 +99,6 @@
   },
   "peerDependencies": {
     "typescript": "*",
-    "webpack": "*"
+    "webpack": "^5.20.0"
```
johnnyreilly, line 102: Thanks!  Why specifically 5.20.0 as the minimum?

afdev82: Because it's the version in the `devDependencies` section, so I assume the plugin is tested against this specific version.

johnnyreilly: It is - but it's not expressing a minimum version for users. To my knowledge, anything 5+ should be fine

afdev82, line 102:
```suggestion
    "webpack": "^5.0.0"
```

johnnyreilly:
Thanks! Should ship with v9.2.2

PR #1334, opened 2021-06-03
getCustomTransformers error message fix

blaky:
The error message here was misleading, it wasn't reporting on the same variable that it checks in line 360. If ts-loader was configured with a `getCustomTransformers` function that returned a module path which didn't had a `module.exports = function(){..}` in it, the error message displayed was the following: `... should export a function, got function`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v9.2.3
+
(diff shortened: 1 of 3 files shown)


johnnyreilly:
Thanks - do you want to add an entry to the `CHANGELOG.md` and increment the `package.json`?

blaky:
@johnnyreilly version number and changelog updated

johnnyreilly (approved):

johnnyreilly:
Shipping with https://github.com/TypeStrong/ts-loader/releases/tag/v9.2.3 - thanks!

PR #1346, opened 2021-07-24
fix: undefined configPath now falls back to default

### LINKED ISSUE #1345: Allow options.configFile to be undefined
### Expected Behaviour

`options.configFile` should allow passing in `undefined` to fallback the standard config file 

We hit this with a webpack config generator function that looks like:

```ts
function createWebpackConfig(configFileName: string | undefined) {
 ...
}
```

Sometimes we want to override the config file but usually we do not 

### Actual Behaviour

Passing in `undefined` produces the following build error:

```
Module build failed (from ./node_modules/ts-loader/index.js):
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at validateString (internal/validators.js:124:11)
    at Object.isAbsolute (path.js:1133:5)
    at findConfigFile (/workspaces/ts-loader/examples/vanilla/node_modules/ts-loader/dist/config.js:57:14)
    at Object.getConfigFile (/workspaces/ts-loader/examples/vanilla/node_modules/ts-loader/dist/config.js:8:28)
    at successfulTypeScriptInstance (/workspaces/ts-loader/examples/vanilla/node_modules/ts-loader/dist/instances.js:74:40)
    at Object.getTypeScriptInstance (/workspaces/ts-loader/examples/vanilla/node_modules/ts-loader/dist/instances.js:42:12)
    at Object.loader (/workspaces/ts-loader/examples/vanilla/node_modules/ts-loader/dist/index.js:16:41)
```

Again, I'd expect `undefined` to be treated as if the `configFile` property has been omitted entirely 

1. In `examples/vanilla`, update the rules to:

```js
rules: [
    {
        test: /\.tsx?$/,
        loader: 'ts-loader',
        options: {
            configFile: undefined
        }
    }
]
```

2. Now build: `yarn run build`

Apply this diff in this repo:

```diff
diff --git a/examples/vanilla/webpack.config.js b/examples/vanilla/webpack.config.js
index 32e454fe..ae9fb98d 100644
--- a/examples/vanilla/webpack.config.js
+++ b/examples/vanilla/webpack.config.js
@@ -13,7 +13,10 @@ module.exports = {
         rules: [
             {
                 test: /\.tsx?$/,
-                loader: 'ts-loader'
+                loader: 'ts-loader',
+                options: {
+                    configFile: undefined
+                }
             }
         ]
     },
```

johnnyreilly:
Fixes #1345 by ensuring that `configPath` is set to the default value if `undefined` is supplied.

There's some prettier reformatting in this that makes the PR a little noisier - but this effectively comes down to a 2 line change.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -267,7 +267,6 @@ function makeLoaderOptions(instanceName: string, loaderOptions: LoaderOptions) {
       logLevel: 'WARN',
       logInfoToStdOut: false,
       compiler: 'typescript',
-      configFile: 'tsconfig.json',
       context: undefined,
       transpileOnly: false,
       compilerOptions: {},
@@ -289,6 +288,7 @@ function makeLoaderOptions(instanceName: string, loaderOptions: LoaderOptions) {
   options.ignoreDiagnostics = arrify(options.ignoreDiagnostics).map(Number);
   options.logLevel = options.logLevel.toUpperCase() as LogLevel;
   options.instance = instanceName;
+  options.configFile = options.configFile || 'tsconfig.json';
 
   // happypack can be used only together with transpileOnly mode
   options.transpileOnly = options.happyPackMode ? true : options.transpileOnly;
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -289,6 +288,7 @@ function makeLoaderOptions(instanceName: string, loaderOptions: LoaderOptions) {
   options.ignoreDiagnostics = arrify(options.ignoreDiagnostics).map(Number);
   options.logLevel = options.logLevel.toUpperCase() as LogLevel;
   options.instance = instanceName;
+  options.configFile = options.configFile || 'tsconfig.json';
```
johnnyreilly, line 291: This line here (and the deletion of 270) is the essential fix

PR #1352, opened 2021-08-03
Add function to get the latest program

### LINKED ISSUE #1216: Program instance for custom transformers becomes stale
### Expected Behaviour

This should not throw an error:

```
export function transformer(program: ts.Program) {
  return (context: ts.TransformationContext) => (file: ts.SourceFile) => {
    if (program.getSourceFile(file.fileName).getText() !== file.getText()) {
      throw new Error();
    }
  };
}
```

### Actual Behaviour

The files in the `program` instance passed in becomes stale when you have a watch running, and `program.getSourceFile()` will return the file as it looked when the watch started.

I was the one who added the `program` argument to `getCustomTransformers` so I'm the one to blame here. 😄 To avoid a breaking change I think we could change it from:
```
(program: Program) => { before?: TransformerFactory<SourceFile>[]; after?: TransformerFactory<SourceFile>[]; }
```
to:
```
(program: Program, getProgram: () => Program) => { before?: TransformerFactory<SourceFile>[]; after?: TransformerFactory<SourceFile>[]; }
```

And in a new major remove the passed in program and just pass a function that gets the latest program instance.

Zn4rK:
Closes #1216.

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -270,7 +270,8 @@ export interface LoaderOptions {
   getCustomTransformers:
     | string
     | ((
-        program: typescript.Program
+        program: typescript.Program,
+        getProgram: () => typescript.Program
       ) => typescript.CustomTransformers | undefined);
   experimentalWatchApi: boolean;
   allowTsInNodeModules: boolean;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v9.2.5
+
+* [Add function to get the latest program](https://github.com/TypeStrong/ts-loader/pull/1352) - thanks @Zn4rK
+
 ## v9.2.4
 
 * [Fix undefined configPath now falls back to default](https://github.com/TypeStrong/ts-loader/pull/1346) - thanks @johnnyreilly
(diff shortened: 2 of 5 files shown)


src/interfaces.ts
```diff
@@ -270,7 +270,8 @@ export interface LoaderOptions {
   getCustomTransformers:
     | string
     | ((
-        program: typescript.Program
+        program: typescript.Program,
+        getProgram: () => typescript.Program
```
johnnyreilly, line 274: Is it worth having `program` and `getProgram` given that `getProgram` returns a `program`?

Zn4rK: I just went with the suggestion @andersekdahl had in #1216. Removing the program in a minor or patch would be a breaking change...

Zn4rK: Most users will not have a problem with the original implementation, since most transformers don't rely on the sourceFiles in the program.

johnnyreilly: Makes sense - thanks!

johnnyreilly:
Could you update the docs please: https://github.com/TypeStrong/ts-loader#getcustomtransformers

Zn4rK:
Sure - I knew I forgot something - updated :)

johnnyreilly:
Nice work - thanks!

PR #1367, opened 2021-08-29
Allow source maps to be merged in loader pipeline

bojanv55:
Allows source maps to be passed in as an argument and mapped to final source map.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,8 @@
     "chalk": "^4.1.0",
     "enhanced-resolve": "^5.0.0",
     "micromatch": "^4.0.0",
-    "semver": "^7.3.4"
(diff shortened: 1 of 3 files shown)


johnnyreilly:
Thanks for contributing! Could you tell us more about your use case please?

bojanv55:
> Thanks for contributing! Could you tell us more about your use case please?

https://github.com/aurelia/aurelia/issues/1052

johnnyreilly:
Sorry about the delay in responding.  This is an interesting PR - I like the the idea of improving source map support.  Can you advise on what the impact on performance of this is please?  It looks like it's a potentially expensive operation?

PR #1375, opened 2021-09-11
Feature: allow for custom writeFile method

bmarti44:
Disclaimer: what I'm doing here is probably wrong, but any guidance and direction is greatly appreciated!

I'm working on a project that uses typescript's compilerHost.writeFile functionality to evaluate `d.ts` files and adds//modifies additional typings. In order to use ts-loader, I need to be able to use into the same writeFile functionality.

The below "works", but I've seen quite a few other places where compilers get created. I'm sure I've missed a few things. If you could point me in the correct direction for anything else that needs to change, or any recommendations, I'll be glad to adjust and make the changes.

Once there's more of an agreement on direction, I will put in tests and update documentation.

Thanks for any and all help!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 !index.js
 /src/*.js
 /src/.vscode
+/.idea
 /*.d.ts
 /*.log
 *.js.map
--- a/src/index.ts
+++ b/src/index.ts
@@ -218,6 +218,7 @@ const validLoaderOptions: ValidLoaderOptions[] = [
   'appendTsxSuffixTo',
   'onlyCompileBundledFiles',
   'happyPackMode',
+  'writeFile',
   'getCustomTransformers',
   'reportFiles',
   'experimentalWatchApi',
--- a/src/instances.ts
+++ b/src/instances.ts
@@ -786,7 +786,10 @@ export function getEmitOutput(instance: TSInstance, filePath: string) {
       fileName: string,
       text: string,
       writeByteOrderMark: boolean
-    ) => outputFiles.push({ name: fileName, writeByteOrderMark, text });
+    ) => {
+      outputFiles.push({ name: fileName, writeByteOrderMark, text });
+      instance.loaderOptions?.writeFile?.(fileName, text, writeByteOrderMark);
+    };
     const outputFilesFromWatch = getEmitFromWatchHost(instance, filePath);
     if (outputFilesFromWatch) {
       return outputFilesFromWatch;
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -267,6 +267,11 @@ export interface LoaderOptions {
   appendTsSuffixTo: (RegExp | string)[];
   appendTsxSuffixTo: (RegExp | string)[];
   happyPackMode: boolean;
+  writeFile: (
+    fileName: string,
+    text: string,
+    writeByteOrderMark: boolean
+  ) => void;
   getCustomTransformers:
     | string
     | ((


johnnyreilly:
Heya!

My time is a bit limited I'm afraid but I'll help where I can. It sounds like you're writing a kind of tool to update Type definitions. How does ts-loader connect with that?

bmarti44:
thanks @johnnyreilly i really appreciate any and all help! i'm looking into an integration with Oracle's JET framework with ts-loader, specifically modifying type definitions as the file gets written out. we have a type, `GlobalProps`, that if found in a type definition being written to the file system, we in turn modify the type definition file to in include a few more imports of JSX intrinsic types and a few other types. i did not create this file so i'm not super familiar with it, but i can get you more details if needed, let me know.

Here is the compiler bit where we wire up the writeFile logic ->
https://github.com/oracle/oraclejet/blob/master/dist/custom-tsc/compile.js#L45

And then the writeFile callback where we modify the types ->
https://github.com/oracle/oraclejet/blob/master/dist/custom-tsc/JetTypeGenerator.js#L35

given the above, i'm also open to a different solution if necessary. i'm guessing we'd prefer access to the writeFile mechanism to keep the current workflow, if possible.

johnnyreilly:
> specifically modifying type definitions as the file gets written out

I find this detail really interesting / curious. Generally type definitions are static, as they represent a specific instance of a JavaScript program. So modifying them on the fly seems like quite an unusual use case.

I'm curious, is there a reason you don't generate the typedefs up front? I'm unfamiliar with oraclejet I'm afraid

bmarti44:
to tell you the truth, i'm (somewhat) new to Oracle, and learning a bit about this code as i work on this prototype for converting a JET project from requirejs to webpack...

from what i see in the code, it looks like we are using this to define properties/methods (both default and custom) that either get decoratored with something like a `@method` decorator, or defined in comment annotations (via an `ojmetadata` marker), or defined in a separate component.json file.

via the above methods, we then take this information and define the additional properties in the typings for a custom element. let me try to provide a clearer example...

so you can decorate a preact based component with a `@customElement` decorator, like the below ->

```
@customElement('app-root') // defines this component as a custom element, and applies additional helper methods
export class App extends Component<ExtendGlobalProps<Props>, State> {
  static defaultProps: Props = {
    appName: 'VDOM Training',
    userLogin: 'john.hancock@oracle.com',
  };

  @method
  someExampleMethodOnTheCustomElement() {
    // available to call this method on the custom element with the @method decorator
  }
  
  render() {
    return <div>something</div>;
  }
}
```

the `@customElement` decorator defines both the component as a custom element and additional methods that can be accessed from the element, for example a `setProperty` method that allows you to access or set any public property on the component.

since this decorator uses something like `Object.define` to define these methods and properties, we lose the typing in the process. the writeFile mechanism you see here makes the lost types accessible by writing out new typings and importing them into the already defined types, allowing us to have somewhat dynamic typing (at least for custom defined properties).

that's more on our use case, but i think there's a more interesting general case of a developer wanting to know what files are being written by the typescript compiler, and performing some action based on what is being written.

johnnyreilly:
Interesting. And custom transformers don't cover your use case at all?

> do you want to allow ts-loader to provide a functionality where files can be analyzed before they are written to the file system?

To be honest, you're the first person that's asked for this, and it does seem like an unusual use case. Apart from your own situation, I'm not sure if this is generally useful.

bmarti44:
i'm not sure a custom transformer is ideal? what i'd like to enable is creating a new file, based on the current files being written. from looking at the transformers, they are passed a node representing an AST of the file that allows for modifying the file output (of course not telling you anything you don't already know, just being pedantic). this is less about the pipeline for modifying files, and more about providing awareness around what files are ultimately being written to a consumer of ts-loader. my use case is indeed a bit different (and as an Angular developer who is used to a different workflow, i will concede this is a bit strange) but i think there might still be room for applying this idea to general use cases.

1. much like in ts-loader itself, perhaps someone would want to generate/modify a file (like sourcemaps) based upon the output? i see that the current writeFile implementation in ts-loader saves a list of output files for working with sourcemaps?
2. anytime someone would like to perform some action based on the generated output, be it related to sourcemaps, types, logging, watching, or another integration.

please don't take the above in the wrong tone, i'm extremely grateful for any help. i'm just pleading my case a bit. if ultimately the decision is our use case is too niche, and allowing users to perform an action based on the output file is not something ts-loader wishes to support, i respect and support that. i think there are other avenues that we can take (like running this same function over what gets generated in our dist folder) but i wanted to, as closely as possible, follow what is being done by the Oracle JET team as possible (I work on the architecture team with their Global Business Units, outside of the JET team).

finally, thank you for your time!

bmarti44:
i was able to run this by our team and received some more information. it looks like we might be able to use the `afterDeclarations` hook to accomplish the same thing we're doing with the `writeFile` method. our team is continuing to investigate that avenue (would require a little bit of a refactor on our end, but not the end of the world).

johnnyreilly:
That's good to hear. A goal for ts-loader is to be, as much as is practical, a drop in webpack aware replacement for tsc. Generally we want to avoid adding APIs that turn ts-loader into a special purpose build tool. Ideally projects can migrate to and away from ts-loader with minimum friction. So unless the reason is super compelling we want to avoid adding custom hooks.

bmarti44:
so it looks like your original take was the correct one - i think what we're doing can be done with an `afterDeclarations` custom transformer instead of writeFile. a quick example i hacked together ->

```
function jetTypeTransformer(typeGenerator) {
  return (context) => {
    const visit = (node) => {
      if (node && node.fileName && node.text) {
        typeGenerator(node.fileName, node.text);
      }
      return ts.visitEachChild(node, (child) => visit(child), context);
    };
    return (node) => ts.visitNode(node, visit);
  };
}
```

we have some more work to do, but yeah we can just pull the filename and content and pass it along to the type generator. this will be a good interim solution while we migrate the type generator over to be an AST instead of regex's on the content.

thank you so much for bearing with me here!

johnnyreilly:
Happy you got where you needed to!

PR #1405, opened 2022-01-19
Use webpack's inputFileSystem for the resolver, instead of a seperately cached version

markjm:
Creating a resolver without passing in a filesystem creates an all new cached file system, both causing more cache misses and double caching of other entries

https://github.com/webpack/enhanced-resolve/blob/d1a3cb65cdf90754871a263172006e2902a2f86e/lib/index.js#L20

### DIFF
--- a/src/servicesHost.ts
+++ b/src/servicesHost.ts
@@ -59,7 +59,8 @@ function makeResolversAndModuleResolutionHost(
   const getCurrentDirectory = () => loader.context;
 
   // make a (sync) resolver that follows webpack's rules
-  const resolveSync = makeResolver(loader._compiler!.options);
+  const webpackOptions = loader._compiler!.options;
+  const resolveSync = makeResolver(webpackOptions, loader.fs);
 
   const moduleResolutionHost: ModuleResolutionHostMayBeCacheable = {
     trace: logData => instance.log.log(logData),
@@ -175,9 +176,10 @@ export function makeServicesHost(
         return file.version.toString();
       }
 
-      const outputFileAndKey = instance.solutionBuilderHost?.getOutputFileAndKeyFromReferencedProject(
-        fileName
-      );
+      const outputFileAndKey =
+        instance.solutionBuilderHost?.getOutputFileAndKeyFromReferencedProject(
+          fileName
+        );
       if (outputFileAndKey !== undefined) {
         instance.solutionBuilderHost!.outputAffectingInstanceVersion.set(
           outputFileAndKey.key,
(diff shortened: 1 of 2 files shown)


src/servicesHost.ts
```diff
@@ -175,9 +176,10 @@ export function makeServicesHost(
         return file.version.toString();
       }
 
-      const outputFileAndKey = instance.solutionBuilderHost?.getOutputFileAndKeyFromReferencedProject(
-        fileName
-      );
+      const outputFileAndKey =
```
markjm, line 179: all changed below this were auto-changed by lint-staged

johnnyreilly:
Thanks! @alexander-akait does this look like an approach that makes sense to you?

alexander-akait (commented):
Make sense, fs calls will be cachable

johnnyreilly:
This looks good @markjm!  Do you have any stats on what difference this makes to performance?

markjm:
I am currently in an employment transition period, so don't have the opportunity to do the benchmarking on a live project that I could before. This came up in my efforts to push the perf improvements across the ecosystem from https://github.com/webpack/enhanced-resolve/issues/316.

I can try and make some dummy project to get numbers if youd like

johnnyreilly:
Yeah that'd be very helpful indeed!

markjm:
@johnnyreilly haha, I am glad you asked me to check on this. In testing, I learned that we _never_ hit this code path, which seems like a bug in itself.

Calling resolveSync with `undefined` context immediately throws,
https://github.com/TypeStrong/ts-loader/blob/5c030bfdedbffac475530f1c90086ecef9a5b2b7/src/servicesHost.ts#L1230

so we always fall back to ts resolution
https://github.com/TypeStrong/ts-loader/blob/5c030bfdedbffac475530f1c90086ecef9a5b2b7/src/servicesHost.ts#L1244

Fixing this logic is pretty trivial (as context is passed in via `loader.context`), but need to do some digging into performance and breakingness of that change

johnnyreilly:
Awesome! Thanks for letting me know ☺️

markjm:
So if we could use the caching from the NormalModule resolver, then I think that would be faster, but since that is an async resolver we can't use it here. Fastest option may actually just be removing the sync resolver from this loader all together to stop the Error spam (even if its immediately caught). I am not sure, though, if that code path is required for older webpack versions

johnnyreilly:
Maybe experiment and see what you find?

alexander-akait:
I think we should finish it

johnnyreilly:
If someone would like to work on this I'll happily review a PR. Sounds like you are keen @alexander-akait ?

alexander-akait:
Do we have an issue for it, I will put it in TODO list?

johnnyreilly:
Just raised #1450 @alexander-akait!

PR #1439, opened 2022-03-08
Bug fix: support webpack 5 in ts-loader

### LINKED ISSUE #1438: Bug fix: support webpack 5 in ts-loader
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `ts-loader@9.2.6` for the project I'm working on.

Error when running ts-loader with webpack 5 :  "times is not iterable"
Previous work was done here by @elf-mouse #1290 but was not accepted.

In my solution I used the new webpack 5 API inside the compilation object, that exposes the _fileTimestamps through fileSystemInfo

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/ts-loader/dist/watch-run.js b/node_modules/ts-loader/dist/watch-run.js
index a2f65a7..5e91edc 100644
--- a/node_modules/ts-loader/dist/watch-run.js
+++ b/node_modules/ts-loader/dist/watch-run.js
@@ -25,26 +25,30 @@ function makeWatchRun(instance, loader) {
             instance.reportTranspileErrors = true;
         }
         else {
-            const times = compiler.fileTimestamps;
-            for (const [filePath, date] of times) {
-                const key = instance.filePathKeyMapper(filePath);
-                const lastTime = lastTimes.get(key) || startTime;
-                if (!date ||
-                    date === 'ignore' ||
-                    (date.timestamp || date.safeTime) <= lastTime) {
-                    continue;
-                }
-                lastTimes.set(key, date.timestamp || date.safeTime);
-                promises.push(updateFile(instance, key, filePath, loader, loaderIndex));
-            }
-            // On watch update add all known dts files expect the ones in node_modules
-            // (skip @types/* and modules with typings)
-            for (const [key, { fileName }] of instance.files.entries()) {
-                if (fileName.match(constants.dtsDtsxOrDtsDtsxMapRegex) !== null &&
-                    fileName.match(constants.nodeModules) === null) {
-                    promises.push(updateFile(instance, key, fileName, loader, loaderIndex));
-                }
-            }
+            compiler.hooks.compilation.tap('ts-loader', compiliation => {
+                compiliation.fileSystemInfo._fileTimestamps.stack.forEach(times => {
+                    for (const [filePath, date] of times) {
+                        const key = instance.filePathKeyMapper(filePath);
+                        const lastTime = lastTimes.get(key) || startTime;
+                        if (!date ||
+                          date === 'ignore' ||
+                          (date.timestamp || date.safeTime) <= lastTime) {
+                            continue;
+                        }
+                        lastTimes.set(key, date.timestamp || date.safeTime);
+                        promises.push(updateFile(instance, key, filePath, loader, loaderIndex));
+                    }
+
+                    // On watch update add all known dts files expect the ones in node_modules
+                    // (skip @types/* and modules with typings)
+                    for (const [key, { fileName }] of instance.files.entries()) {
+                        if (fileName.match(constants.dtsDtsxOrDtsDtsxMapRegex) !== null &&
+                          fileName.match(constants.nodeModules) === null) {
+                            promises.push(updateFile(instance, key, fileName, loader, loaderIndex));
+                        }
+                    }
+                })
+            })
         }
         // Update all the watched files from solution builder
         if (instance.solutionBuilderHost) {
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

einatbar:
Fixes Error when running ts-loader with webpack 5 : "times is not iterable"

closes #1438

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
+## v9.2.8
+
+* [Bug fix: support webpack 5 in ts-loader](https://github.com/TypeStrong/ts-loader/pull/1439) [#1438] - thanks @einatbar
+
 ## v9.2.7
 
-* [cater for change in resolveTypeReferenceDirective API in TypeScript 4.7](https://github.com/TypeStrong/ts-loader/pull/1422)  [#1421] - thanks @johnny_reilly and @cspotcode for inspiration in ts-node work here: https://github.com/TypeStrong/ts-node/pull/1648
+* [cater for change in resolveTypeReferenceDirective API in TypeScript 4.7](https://github.com/TypeStrong/ts-loader/pull/1422) [#1421] - thanks @johnny_reilly and @cspotcode for inspiration in ts-node work here: https://github.com/TypeStrong/ts-node/pull/1648
 
 ## v9.2.6
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "9.2.7",
+  "version": "9.2.8",
   "description": "TypeScript loader for webpack",
   "main": "index.js",
   "types": "dist",
--- a/src/watch-run.ts
+++ b/src/watch-run.ts
@@ -29,36 +29,38 @@ export function makeWatchRun(
     if (instance.loaderOptions.transpileOnly) {
       instance.reportTranspileErrors = true;
     } else {
-      const times = compiler.fileTimestamps;
+      compiler.hooks.compilation.tap('ts-loader', compiliation => {
+        compiliation.fileSystemInfo._fileTimestamps.stack.forEach(times => {
+          for (const [filePath, date] of times) {
+            const key = instance.filePathKeyMapper(filePath);
+            const lastTime = lastTimes.get(key) || startTime;
 
-      for (const [filePath, date] of times) {
-        const key = instance.filePathKeyMapper(filePath);
-        const lastTime = lastTimes.get(key) || startTime;
+            if (
+              !date ||
+              date === 'ignore' ||
+              (date.timestamp || date.safeTime) <= lastTime
+            ) {
+              continue;
+            }
 
-        if (
-          !date ||
-          date === 'ignore' ||
-          (date.timestamp || date.safeTime) <= lastTime
-        ) {
-          continue;
-        }
+            lastTimes.set(key, date.timestamp || date.safeTime);
+            promises.push(updateFile(instance, key, filePath, loader, loaderIndex));
+          }
 
-        lastTimes.set(key, date.timestamp || date.safeTime);
-        promises.push(updateFile(instance, key, filePath, loader, loaderIndex));
-      }
-
-      // On watch update add all known dts files expect the ones in node_modules
-      // (skip @types/* and modules with typings)
-      for (const [key, { fileName }] of instance.files.entries()) {
-        if (
-          fileName.match(constants.dtsDtsxOrDtsDtsxMapRegex) !== null &&
-          fileName.match(constants.nodeModules) === null
-        ) {
-          promises.push(
-            updateFile(instance, key, fileName, loader, loaderIndex)
-          );
-        }
-      }
+          // On watch update add all known dts files expect the ones in node_modules
+          // (skip @types/* and modules with typings)
+          for (const [key, { fileName }] of instance.files.entries()) {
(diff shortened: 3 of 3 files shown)


johnnyreilly:
Thanks! Just kicked off the tests

johnnyreilly:
The build is failing at the moment?

johnnyreilly:
`compiliation.fileSystemInfo._fileTimestamps.stack.forEach(times`

I notice this approach depends upon `_fileTimestamps` - is this API for public consumption? @alexander-akait / @sokra can you advise?

alexander-akait:
@vankop friendly ping, I think it is regression

vankop:
hm.. maybe some watch bug.. could you create a small reproducible repo? From code this could happen (`fileTimestamps=undefined`). so fast fix could be just check that `fileTimestamps` exists (`if (!fileTimestamps) return;`)

johnnyreilly:
This is actually a question about types; we're getting this compilation error:

> src/watch-run.ts(33,37): error TS2551: Property '_fileTimestamps' does not exist on type 'FileSystemInfo'

It suggests that there is no available `_fileTimestamps` property. The fact there is an `_` at the start suggests it may be an internal property and intentionally not exposed with a type - is that right?

vankop:
yes, `_fileTimestamps` is internal property and is not presented in types.

johnnyreilly:
Okay cool. Is there a public API that can be used? Is there an `fileTimestamps` for instance?

johnnyreilly:
Just looking at the code and I can see there's a `getDeprecatedFileTimestamps()` method. But the name suggests we shouldn't rely upon it... Is there a non deprecated one by any chance?

https://github.com/webpack/webpack/blob/770dea1fb46c7c5e04016dbab1f0854037b00d81/lib/FileSystemInfo.js#L3537

johnnyreilly:
@einatbar would you like to update your pr to use `getDeprecatedFileTimestamps()`? Assuming that works, we could use that as a basis for a conversation with the webpack team about a non deprecated endpoint.

@alexander-akait / @vankop if you have thoughts then please do share!

vankop:
I think `compiler.fileTimestamps` should work.. need @sokra to take a look.

alexander-akait (commented):
I think we should use snapshot API here, there is small example how we do it in copy plugin https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/src/index.js#L569

vankop:
If I understand correctly this code snippet does:
- watch N files
- if one file changed => regenerate definitions

So you can use 2 api methods [createSnapshot](https://github.com/webpack/webpack/blob/main/lib/FileSystemInfo.js#L1886) (example above) and on watch callback [check that snapshot is valid](https://github.com/webpack/webpack/blob/main/lib/FileSystemInfo.js#L2385)
But this idea seems wrong, since you will need 1 snapshot per file.. (this will decrease build performance)

So I think you can use [`compiler.modifiedFiles`](https://github.com/webpack/webpack/blob/main/lib/Compiler.js#L234) and [`compiler.removedFiles`](https://github.com/webpack/webpack/blob/main/lib/Compiler.js#L236)

einatbar:
> and it still unclear to me why compiler. fileTimestamps does not work

@vankop `compiler. fileTimestamps` is sometimes undefined, which causes the error "times is not iterable".
I am not sure what causes it to be undefined

johnnyreilly:
> > and it still unclear to me why `compiler.fileTimestamps` does not work
> @vankop `compiler.fileTimestamps` is sometimes undefined, which causes the error "times is not iterable".
I am not sure what causes it to be undefined

Would it be enough to guard against `compiler.fileTimestamps` being `undefined`?

johnnyreilly:
> > Would it be enough to guard against compiler.fileTimestamps being undefined?
> 
> @johnnyreilly looks like it's working as well if I just guard against compiler.fileTimestamps

Cool - fancy updating the PR?

einatbar:
@johnnyreilly sure, on it

johnnyreilly:
Awesome, do you want to update the `package.json` and the `CHANGELOG.md`?  Then I think we're ready to ship!

elf-mouse:
@einatbar , thank you very much :)

einatbar:
@johnnyreilly Thank you! :)

johnnyreilly:
My pleasure @einatbar !

PR #1451, opened 2022-04-28
feat: add zero-config support for fork-ts-checker-webpack-plugin

piotr-oles:
Hi! :wave:
I'm a maintainer of the [fork-ts-checker-webpack-plugin](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin).
Currently, we have to set `transpileOnly: true` to skip type-checking on the loader side.
With this change it's no longer the case - the loader will detect if there is a checker plugin in the configuration and set default transpileOnly: true if that's the case. I hope this will improve DX and make webpack config files cleaner and smarter :)

#### Before:
```javascript
// webpack.config.js
const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');

module.exports = {
  context: __dirname, // to automatically find tsconfig.json
  entry: './src/index.ts',
  resolve: {
    extensions: [".ts", ".tsx", ".js"],
  },
  module: {
    rules: [
      {
        test: /\.tsx?$/,
        loader: 'ts-loader',
        options: {
          // disable type checker - we will use it in fork plugin
          transpileOnly: true
        }
      }
    ]
  },
  plugins: [new ForkTsCheckerWebpackPlugin()]
};
```

#### After:
```javascript
// webpack.config.js
const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');

module.exports = {
  context: __dirname, // to automatically find tsconfig.json
  entry: './src/index.ts',
  resolve: {
    extensions: [".ts", ".tsx", ".js"],
  },
  module: {
    rules: [
      {
        test: /\.tsx?$/,
        loader: 'ts-loader'
      }
    ]
  },
  plugins: [new ForkTsCheckerWebpackPlugin()]
};
```

I tested this manually, please let me know if you need some automatic tests for this.


johnnyreilly:
Hey @piotr-oles - I know who you are! 😅

I think this is a great improvement! Could we make an enhancement? With the current implementation, you can't explicitly control `transpileOnly` - it's always derived from whether the plugin is present.

If someone wants to run with `transpileOnly` and not use a checker, we'd still want to support that.

So passing `transpileOnly: false` should be possible in the absence of checker usage. Would you be up for making that change?

johnnyreilly:
Actually, I read that wrong - your PR already implements the behaviour I mentioned. Could you update the `README.md` to document new usage of `fork-ts-checker-webpack-plugin`, increment the version in the `package.json` and add an entry to the `CHANGELOG.md` please?

piotr-oles:
Done! ✅

johnnyreilly (approved):
Let's ship it!

johnnyreilly:
https://github.com/TypeStrong/ts-loader/releases/tag/v9.3.0

PR #1477, opened 2022-06-11
chore: add vscode extension settings

unional:
for making the VSCode environment stable so that the `format document` will keep the code style as expected.


johnnyreilly (approved):
Nice work!

PR #1478, opened 2022-06-11
chore: upgrade to yarn 3

unional:
A testing PR to see if there are any issue about it for this project


johnnyreilly:
Looks like this isn't working out?

johnnyreilly:
That said, it looks like we're getting arbitrary failures now which I would guess to be new npm related

johnnyreilly:
Wonder if this --legacy-peer-deps will fix

https://github.com/TypeStrong/ts-loader/pull/1479

unional:
Look like logistic issue. Let me check

unional:
Yeah, it's the `yarn.lock` file changed during test issue.

unional:
You are right. It is actually that. It's about that `npm link --legacy-peer-deps` fix.

I read it wrong. I through it was the lockfile frozen issue. But the fail is actually during test, not during install

johnnyreilly:
I've updated with the --legacy-peer-deps fix. That said, I'm not sure if we want to move to yarn 3 - may switch back to npm instead. We originally moved to yarn as (then) npm didn't have meaningful lock files. Now it does we may want to switch to that instead given yarn 3 doesn't seem to be widely used

unional:
Sure. The yarn 2/3 usage rate is a difficult topic.
What happen is that the usage of yarn 2/3 is based off of yarn 1:

```sh
# install yarn 1 (note that the latest version is actually yarn 1. 1.22.19)
npm install -g yarn

# or by using corepack
corepack enable

# enable yarn 3
yarn set version stable
```

It actually downloads the bundled `yarn 3` file to `./.yarn/releases/yarn-3.2.1.cjs`.
So it doesn't go through the registry and the usage rate is not measured in a traditional way.
I don't know where to find that info.

I have also tried using `pnpm` in one of my projects.

In general, `pnpm` > `yarn 3` > `npm`.
But evil is in the detail, as always, such as the strict peer deps, and some weird bug due to certain expectation of the `node_modules` structure (https://github.com/storybookjs/presets/issues/242)

johnnyreilly:
I also like pnpm - as much as anything else I'm keen on using stuff that's already installed on CI machines though. With pnpm I'd have to manually install. Not the end of the world - just another thing to do

unional:
> With pnpm I'd have to manually install.

You don't have to. `corepack` takes care of it. The only thing need is `packageManager: pnpm@<version>` in the `package.json`

unional:
btw, some CI tools doesn't work with `yarn 3`. I'm sure `synk` doesn't work and I recall `dependabot` doesn't either. So that might be a deal breaker.

johnnyreilly:
Yeah we'll either go with npm or pnpm

johnnyreilly:
Must also get round to looking at renovate in place of dependabot

unional:
> Must also get round to looking at renovate in place of dependabot

I'm using that, and `mergify` :)

PR #1481, opened 2022-06-11
chore: try to fix the peer deps issue by upgrading babel

unional:
trying to see if that fixes the issue instead of using `legacy-peer-deps`.

Currently, this is blocking the usage of `pnpm` as I'm trying it out.

I'm not sure will this be a good idea or if `ts-loader` wants to use the oldest version of `babel-loader` to ensure compatibility (or should work with the latest version, depends on the argument. 😛 )


johnnyreilly:
Happy for Babel to be upgraded - looks like this didn't succeed though. I haven't used Babel forever - and it's possible these aren't meaningful failures. They relate to comparison tests which will be affected by different output from Babel - which a new version will almost certainly supply.

johnnyreilly:
> node_modules/babel-loader/lib/index.js):/nError: Plugin/Preset files are not allowed to export objects, only functions. In /home/runner/work/ts-loader/ts-loader/node_modules/babel-preset-es2015/lib/index.js/n    at createDescriptor (/home/runner/work/ts-loader/ts-loader/node_modules/@babel/core/lib/config/config-descriptors.js:211:11)/n

Actually it looks like they are meaningful errors - Babel unhappy

unional:
Those `babel-preset-es2015` are deprecated. But I didn't change anything because I don't know what is the purpose of the code. :)

johnnyreilly:
So Babel is only in the repo for testing purposes. People can chain babel-loader and ts-loader and these tests covered that use case. I'd be happy to delete all comparison tests which are Babel related given that there are execution tests that cover that use case

PR #1503, opened 2022-09-18
Add Support for Resolving `.cjs`, `.mjs`, `.cts` and `.mts` Files

### LINKED ISSUE #1383: ts-loader with Typescript 4.5 beta: "Error: TypeScript emitted no output for .../index.mts"
Anyone have luck webpack-5-ing with the new `.mts` extension?  I'm able to output from `tsc -p .` fine, but through webpack I'm getting "Error: TypeScript emitted no output for .../index.mts".

Thanks!

```
    resolve: {
      // Add `.ts` and `.tsx` as a resolvable extension.
      extensions: [".mts", ".ts", ".tsx", ".js"]
    },
    module: {
      rules: [
        {
            test: /(\.mts$)|(\.tsx?$)/,
            use: [
                {
                    loader: 'ts-loader',
                    options: {
                        //transpileOnly: true
                        compilerOptions: {
                            noEmit: false
                        }
                    }
                }
            ]
        }
      ]
    }
```

manuth:
Changes made in this PR will fix #1383
Currently, `ts-loader` does not consider files with extensions like `.cts` and `.mts` as valid source files. Changes made in this PR change the regular expressions in order to make them match these new file extensions.

Furthermore, changes made in this PR update the resolve logic to automatically resolve imports performed in `.mjs` files using `import` and `.cjs` files using `require` (introduced in 75a4dfdfa8f1b78914bcf3e21656b0553b37144f).

Also, I have noticed that the `enhanced-resolver` call is broken. Passing `undefined` as the first argument throws an error. Thus, the use of the `enhanced-resolver` has not been working up until this point. This change has been introduced in 8dda7d3ed1c821b811c8e9052cccc315681b0e22.

To verify that passing `undefined` as the first argument triggers an error, you might want to have a look at these pieces of code:
https://github.com/webpack/enhanced-resolve/blob/651e57863e46afeccf1946b4275d7b2339152ed1/lib/index.js#L51-L58
https://github.com/webpack/enhanced-resolve/blob/651e57863e46afeccf1946b4275d7b2339152ed1/lib/Resolver.js#L231-L237
https://github.com/webpack/enhanced-resolve/blob/651e57863e46afeccf1946b4275d7b2339152ed1/lib/Resolver.js#L260-L268

I might have to mention, that changes made in previously mentioned commit 8dda7d3ed1c821b811c8e9052cccc315681b0e22 break the current tests. In some test cases, there are less errors reported than expected.


johnnyreilly:
Thanks for contributing. I've approved the test run so let's see what errors out. The comparison test pack is effectively a set of snapshot tests. A failing test there means "something has changed" - that could be legitimate

> Also, I have noticed that the enhanced-resolver call is broken. Passing undefined as the first argument throws an error. Thus, the use of the enhanced-resolver has not been working up until this point. 

Interesting. If it wasn't working, didn't that cause issues? I wasn't aware of any being experienced.

src/instances.ts
```diff
@@ -682,7 +682,7 @@ export function getInputFileNameFromOutput(
   instance: TSInstance,
   filePath: string
 ): string | undefined {
-  if (filePath.match(tsTsxRegex) && !fileExtensionIs(filePath, '.d.ts')) {
+  if (filePath.match(tsTsxRegex) && !/\.d\.([cm]?ts|tsx)$/.test(filePath)) {
```
johnnyreilly, line 685: Could we give this regex a name and site it in `constants.ts`?

manuth:
> Thanks for contributing. I've approved the test run so let's see what errors out. The comparison test pack is effectively a set of snapshot tests. A failing test there means "something has changed" - that could be legitimate
> 
> > Also, I have noticed that the enhanced-resolver call is broken. Passing undefined as the first argument throws an error. Thus, the use of the enhanced-resolver has not been working up until this point. 
> 
> Interesting. If it wasn't working, didn't that cause issues? I wasn't aware of any being experienced.

If I'm interpreting the test logs right, using "resolve"-options such as "alias" has only been possible using `transpileOnly` as this would leave `webpack` taking care about the rest, I guess...?

With `enhanced-resolve` working now inside `ts-loader`, I assume that stuff like the `alias`-option now work without being forced to use `transpileOnly`.

Not quite sure about it, though. But it does look like `enhanced-resolve` not working only affected a small number of edge cases.

johnnyreilly:
image

Is the `yarn.lock` change necessary? The diff is too large for me to review it.

manuth:
> image
> 
> Is the `yarn.lock` change necessary? The diff is too large for me to review it.

No, I didn't mean to include this file

johnnyreilly:
Cool - could you revert that please?

manuth:
I reverted the change made to `yarn.lock`. Also, I changed this code passage a little to ensure the signature of `resolveModuleNames` matches the one expected by `typescript`:
https://github.com/TypeStrong/ts-loader/blob/c218be48e2db044b2b8917dfb373ff0c9712d545/src/servicesHost.ts#L263-L271

johnnyreilly:
Cool, could you regenerate the snapshots please?  Instructions here: https://github.com/TypeStrong/ts-loader/tree/main/test/comparison-tests#regenerating-test-data

Just this will do I think:

```yarn run comparison-tests -- --save-output```

Make sure you `yarn` before you do the regeneration to ensure that what you have installed matches the `yarn.lock` file.

**Comment by manuth** [comment]
@johnnyreilly I added a remark on how to regenerate tests to the Dockerfile:
https://github.com/TypeStrong/ts-loader/blob/8dc552ec1af481520e9259ba16934143daca0293/Dockerfile#L37-L40

Hope that's fine

**Comment by johnnyreilly** [comment]
I'm slightly surprised by the alias test change - I wouldn't have expected different behaviour there. But I'm also not worried as I think this *doesn't* represent broken behaviour.  I think we should roll with it, if there are issues when it reaches the outside world we can revert and look again. But hopefully that wont happen.

Could you update the `package.json` and `CHANGELOG.md` to document the new feature please?

**Comment by manuth** [comment]
Sure! What needs to be changed inside the `package.json` file?

**Comment by johnnyreilly** [comment]
New version number: 9.4.0!

**Comment by manuth** [comment]
Done! Thanks for guiding me through this 😄

File: CHANGELOG.md [hunk 2]
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v9.4.0
+
+* [Add Support for Resolving `.cjs`, `.mjs`, `.cts` and `.mts` Files](https://github.com/TypeStrong/ts-loader/pull/1503) [#1503]
```
johnnyreilly, line 5:
```suggestion
* [Add Support for Resolving `.cjs`, `.mjs`, `.cts` and `.mts` Files](https://github.com/TypeStrong/ts-loader/pull/1503) [#1503] - thanks @manuth 
```

johnnyreilly:
It's a pleasure! Thanks for contributing!

manuth:
> It's a pleasure! Thanks for contributing!

You're most welcome 😄 
Glad I could help

johnnyreilly:
Once this pipeline runs this should be live: https://github.com/TypeStrong/ts-loader/actions/runs/3080976806

johnnyreilly:
This is released in https://github.com/TypeStrong/ts-loader/releases/tag/v9.4.0

johnnyreilly:
It looks as though this change has caused an issue https://github.com/TypeStrong/ts-loader/issues/1504

@manuth would you be able to take a look please? If we can resolve this quickly that'd be great - if we can't then we'll likely have to revert 9.4.0 for now.

manuth:
@johnnyreilly the issue is caused by the fact that `enhanced-resolve` is working now.
Because of the settings, `enhanced-resolve` will resolve certain imports to their corresponding `.ts` file now.

As seen here:
https://github.com/TypeStrong/ts-loader/blob/a810470ffe0f02e916190f82961a2b88580ee8f3/src/servicesHost.ts#L1292-L1297

The result of `enhanced-resolve` has a higher priority than the result of `typescript`.
I'm not quite sure what to do about it.
Instead of checking the `typescript` result using `isJsImplementationOfTypings` we could just check whether it's a type declaration using our newly introduced constant `RegExp`.

What are your thoughts?

johnnyreilly:
I think it's worth trying your idea first to see if it resolves things.  You okay with trying that out?  Should be able to test with the example raised in the issue?

johnnyreilly:
Can you also look at the issue @falsandtru raises here please @manuth?

https://github.com/TypeStrong/ts-loader/issues/1505

It's possible we may need to consider rolling back 9.4.0 if it doesn't look promising

manuth:
I think we need to disable enhanced-resolve for the time being until I managed it to look into it a little further.

johnnyreilly:
Sounds like a plan

PR #1506, opened 2022-09-20
Hotfix: Disable `enhanced-resolve`

manuth:
As discussed in #1503, #1505 and #1504, `enhanced-resolve` causes issues in case both `typescript` and `enhanced-resolve` are able to resolve a module specifier.

Changes made in this PR will disable `enhanced-resolve`. This change might be reverted once a proper workflow has been worked out.

In doing so, this PR will fix #1505 and fix #1504.

This is a follow-up to PR #1503 which introduced the changes causing these issues.

PR #1509, opened 2022-09-21
Enable `enhanced-resolve`  for a Small Amount of Cases

manuth:
Changes made in this PR will enable `enhanced-resolve` in case `alias` or `fallback` is set and affects `enhanced-resolve`s result.
This is an alternative to #1506


johnnyreilly:
very interesting.... It does change behaviour.  What's not obvious is whether it's useful... Thanks for doing this - super helpful

PR #1514, opened 2022-10-10
Raise an error when config file is specified but not found

Y--:
Hi!

A small contribution to explicitly raise an error if user specify a configuration file that cannot be resolved/found.

Please let me know:
* if it makes sense :-)
* if it is the right approach (I've tried to use the `formatErrors` utils but didn't seem to work in this case)
* how I can test it (not sure where to put the test)

Thanks!!

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -50,6 +50,10 @@ export function getConfigFile(
       )[0];
     }
   } else {
+    if (loaderOptions.configFile !== undefined) {
+      configFileError = new Error(`Cannot find '${loaderOptions.configFile }' in folder '${path.dirname(loader.resourcePath)}'`);
+    }
+
     if (compilerCompatible) {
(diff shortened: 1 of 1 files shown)


johnnyreilly:
Thanks for contributing! Is the logic for your check definitely correct?

Y--:
> Thanks for contributing! Is the logic for your check definitely correct?
> 
> ```js
> if (loaderOptions.configFile !== undefined) { // should this be === undefined?
>   configFileError = new Error(`Cannot find '${loaderOptions.configFile }' in folder '${path.dirname(loader.resourcePath)}'`);
> }
> ```

Well I want to throw an error if there is a file specified by the user (i.e. not `undefined`) but it is not found. 

Basically I spent quite some time understanding why my config changes were not taken into account until I realized the file referenced in the loader did not exist, and the loader was instead creating a default configuration :-)

PR #1547, opened 2022-11-16
Typescript 4 9

johnnyreilly:


HaimCandiTech:
@johnnyreilly 
Is there a chance this somehow got into the published version?

We got an error on our vue app:
-  "Uncaught TypeError: Cannot assign to read only property 'exports' of object '#<Object>'

Looking around, we found that the vue build is warning us that Typescript version 4.9.3 is not supported
We tried to specify the typescript version in the package.json but it didn't help
Looking at the package-locks.json we found out that ts-loader uses Typescript 4.9.3
We had to find an old package-lock.json file in order to control it

Where does it specify the exact version of typescript that ts-loader uses?

johnnyreilly:
ts-loader uses "*" - there shouldn't be any dependency forced through ts-loader itself. What's more, we haven't published a new version and so you shouldn't be picking that up from us

https://github.com/TypeStrong/ts-loader/blob/620ee7926bb9384317caeafaa01478baf7ae87d3/package.json#L101

HaimCandiTech:
You are super right
It was wrong for me to assume that
Just some weird warning regarding ts 

Some other package were to blame
Sorry for wasting your time

PR #1550, opened 2022-11-21
Use custom transformer when building solution references

### LINKED ISSUE #1025: Custom transformers not working for referenced projects

feosuna1:
This PR adds support for custom transformers in project references.

While attempting to use custom transformers, I realized that the transformers were only being applied to the main target and not to any of my referenced projects.  After a bit of spelunking, I found that #1025 was a similar issue to my own.  #1274 attempted to solve the problem, but some changes needed to be made in Typescript to allow for custom transformers in project references. All the necessary changes to support are available in Typescript v4.3+.

I've modeled my changes after @gasnier's original PR (#1025). This change set:
1. Moves the `getCustomTransformers` logic out of `initializeInstance` and into a dedicated function, so that it can be shared.
2. Refactors `makeSolutionBuilderHost` to include the new `getCustomTransformer` field in the `solutionBuilderHost` and have it return the transformers made available in the loader options.

Fixes #1025


johnnyreilly:
Thanks for this - little snowed right now. Will look when I can

src/servicesHost.ts
```diff
@@ -789,12 +794,28 @@ export function makeSolutionBuilderHost(
       );
       instance.typeReferenceResolutionCache?.update(instance.compilerOptions);
       instance.moduleResolutionCache?.update(instance.compilerOptions);
+
+      if (options) {
+        // The `configFilePath` is the same value that is used as the `project` parameter of
+        // `getCustomtransformers` below.
+        const project = options.configFilePath?.toString();
```
johnnyreilly, line 801: is the `toString` necessary?

johnnyreilly: isn't it always a string?

johnnyreilly:
Have updated your branch - it's possible you might need to do some repairs

src/servicesHost.ts
feosuna1: You're right.  I don't know what I was thinking.  It's a string.

johnnyreilly:
Failing tests following merge I'm afraid - could you take a look please?

feosuna1:
> Failing tests following merge I'm afraid - could you take a look please?

It started failing because the typescript compiler is complaining that options can be a bunch of different types.

CleanShot 2022-11-26 at 09 24 22@2x

johnnyreilly:
I think we might be nearly there. Could you update the version in the package.json and put an entry in the changeling.md please?

johnnyreilly (approved):

johnnyreilly:
Thanks!

johnnyreilly:
https://github.com/TypeStrong/ts-loader/releases/tag/v9.4.2

PR #1611, opened 2023-05-16
fix: add config file as build dependency

alexander-akait:
Make configuration more DX, we already known typescript configuration, so let's add it as a build dependency and not need provide to add this in build dependency like here https://github.com/TypeStrong/ts-loader/issues/1332

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 9.4.3
+* [Bug fix: add config file as build dependency](https://github.com/TypeStrong/ts-loader/pull/1611) - thanks @alexander-akait
+
 ## 9.4.2
 * [Bug fix: Use custom transformer when building solution references](https://github.com/TypeStrong/ts-loader/pull/1550) [#1025] - thanks @feosuna1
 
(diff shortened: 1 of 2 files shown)


johnnyreilly:
Builds passed - do you want to update the package.json / changelog.md?

alexander-akait:
@johnnyreilly If need I can do, just don't know contirbution process, can you describe?

alexander-akait:
> Change the version in the package.json to represent a new version - I'd say this is a patch rather than anything else.

> The changelog.md is hopefully self explanatory - put a version entry in, linking back to this PR. Nothing more

Okay, I will do it, just thought that the release process is separate from the fixes/features, because we have https://github.com/TypeStrong/ts-loader/commit/5f28651df20ecd77cc8b1c70b4e8c33e18cee73b and I don't know what I should put in Changelog

johnnyreilly:
So the changelog addition would probably be:

```
## 9.4.3
* [fix: add config file as build dependency](https://github.com/TypeStrong/ts-loader/pull/1611) - thanks @alexander-akait 
```

alexander-akait:
@johnnyreilly Done :+1:

johnnyreilly:
Publishing: https://github.com/TypeStrong/ts-loader/releases/tag/v9.4.3

PR #1614, opened 2023-06-14
Set webpack module type to `"javascript/esm"` when TS `impliedNodeFormat` is `ESNext`

andrewbranch:
Or in other words, make Webpack give the same special treatment to `.mts` files as it does `.mjs` files. (Same for `.ts` in scope of a package.json `"type": "module"`.) In short, Webpack changes its ESM/CJS module interop strategy to match Node’s when it sees these file extensions. In order for TypeScript to accurately type check this behavior, it needs to be consistent between JS and TS files. See https://github.com/webpack/webpack/issues/17288 and https://andrewbranch.github.io/interop-test/#synthesizing-default-exports-for-cjs-modules for more of an explanation. (This PR will give Webpack + ts-loader + `"module": "nodenext"` a 💙🌟 in the linked table.)

This is technically a breaking change, but it only affects users who are using `node16`/`nodenext` for their `moduleResolution`, which I suspect is rare in combination with Webpack. In the near future, I’m planning to make a new `module` or other TypeScript option that enables this Node-like interop behavior in `.mts`/`.mjs` files, but can be used with `moduleResolution: "bundler"`. (This is tracked by https://github.com/microsoft/TypeScript/issues/54102.)

I think this should stay open until https://github.com/microsoft/TypeScript/issues/54102 has a clear resolution; I just wanted to make sure it was doable before moving on. Thanks @alexander-akait for the pointers in https://github.com/webpack/webpack/issues/17288.


src/index.ts
```diff
@@ -147,9 +148,19 @@ function setModuleMeta(
     // a previously cached version the TypeScript may be different and therefore should be
     // treated as new
     loaderContext._module!.buildMeta.tsLoaderFileVersion = fileVersion;
+    const impliedNodeFormat = getImpliedNodeFormat(
+      loaderContext._module!.resource,
```
alexander-akait, line 152: Please use `loaderContext.resourcePath`, resource can contains `?foo=bar` and `#hash`

src/index.ts
```diff
@@ -147,9 +148,19 @@ function setModuleMeta(
...
+      instance,
+      loaderContext,
+      instance.program || instance.languageService?.getProgram() || instance.builderProgram?.getProgram()
+    ); 
+    if (impliedNodeFormat === /*ts.ModuleKind.ESNext*/ 99) {
```
johnnyreilly, line 157: "I got 99 problems but EcmaScript modules ain't one"

src/instances.ts
```diff
@@ -459,7 +472,14 @@ export function getCustomTransformers(
     getCustomTransformers = customerTransformers;
   }
 
-  return getCustomTransformers(program, getProgram);
+  let transformers = getCustomTransformers?.(program, getProgram);
+  if (loaderOptions.transpileOnly) {
+    (transformers ??= {}).before = [
+      getSetImpliedNodeFormatTransformer(instance, loaderContext, getProgram),
+      ...(transformers?.before ?? []),
```
johnnyreilly, line 479: this reminds me, I wonder whether we should bump the `target` for the emitted JavaScript of ts-loader at some point.   I suspect it's quite an old version

src/servicesHost.ts
```diff
@@ -134,6 +130,42 @@ function makeResolversAndModuleResolutionHost(
...
+    loaderOptions: {
+      resolveModuleName: customResolveModuleName,
+      resolveTypeReferenceDirective: customResolveTypeReferenceDirective,
+    },
+  } = instance;
+
+  
+
+  // make a (sync) resolver that follows webpack's rules
+  const resolveSync = makeResolver(loader._compiler!.options);
```
alexander-akait, line 153: Do you really need this? Based on logic it is just an error - https://github.com/TypeStrong/ts-loader/blob/main/src/resolver.ts#L11

Webpack has great API for resolving, you can use (just an example of code, we should have two resolvers for cjs and esm in the real life):
```
const resolve = loaderContext.getResolve({
    dependencyType: "typescript",
    conditionNames: ["types", "..."],
    mainFiles: ["index", "..."],
    extensions: [".ts", ".cts", "..."],
});
```

And hide a lot of options for basic `resolve` under the hood - less options better DX, but yeah, I think it should be done in another PR, here is just about ESM

alexander-akait (commented):
Just some notes about code

johnnyreilly:
This looks generally good - and it looks like @alexander-akait is providing some excellent feedback.  I realise I should sort out upgrading the ts-loader test pack to 5.1 as well; will sort that in the next week sometime.

src/servicesHost.ts
andrewbranch: This is just moved code, not code I wrote

alexander-akait:
just idea - maybe when developer has `javascript/dynamic` (i.e. define `type: "javascript/dynamic"`) and you want to say `javascript/esm` we should print a warking, for example like "Typescript emits ES modules, but you handle it as regular Javascript, please remove `type: "javascript/dynamic"` or set it to `javascript/esm`" (we have `loaderContext.emitWarning(new Error(message))` API for this).

alexander-akait:
@andrewbranch Your value, you override it, that is why I think we need to emit a warning to catch some weird configurations/options

johnnyreilly:
Test pack migrated to 5.1 in https://github.com/TypeStrong/ts-loader/pull/1616

andrewbranch:
Just wanted to drop in and say this is still on my radar; just figuring out how to expose the related config in TypeScript itself is proving difficult.

johnnyreilly:
Merry Christmas @andrewbranch 🎄⛄!

PR #1626, opened 2023-08-28
map input source map

### LINKED ISSUE #1457: Questions about sourcemap
Why is the upstream sourcemap not processed in the loader?
 > https://github.com/TypeStrong/ts-loader/blob/main/src/index.ts#L34

```js
function loader(this: webpack.LoaderContext<LoaderOptions>, contents: string, map?: Sourcemap) {
    if (map) {
       ...
    }
}
```

Ka0o0:
Reopening of https://github.com/TypeStrong/ts-loader/pull/1367 
Resolves #1457

Since there was a question about the performance in the original PR: this won't cause any performance regressions for the cases where ts-loader is the first loader, so basically for any plain `.ts` file. For the other cases I would say that functionality goes over performance, since it's currently a serious problem to not have input source maps handled.


johnnyreilly:
> this won't cause any performance regressions for the cases where ts-loader is the first loader,

Can you explain why this is the case please?

Ka0o0:
> > this won't cause any performance regressions for the cases where ts-loader is the first loader,
> 
> Can you explain why this is the case please?

We only need to call the source-map lib if we have an input source map, otherwise we can do an early return, see this here:  https://github.com/TypeStrong/ts-loader/pull/1626/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80R685
I've added a comment to highlight this.


> 
> Also this PR doesn't do `async` the "webpack way": https://webpack.js.org/api/loaders/#asynchronous-loaders
> 
> > A single result can be returned in [sync mode](https://webpack.js.org/api/loaders/#synchronous-loaders). For multiple results the this.callback() must be called. **In [async mode](https://webpack.js.org/api/loaders/#asynchronous-loaders) this.async() must be called to indicate that the [loader runner](https://github.com/webpack/loader-runner) should wait for an asynchronous result. It returns this.callback(). Then the loader must return undefined and call that callback.**
> 
> Significantly, use of `async` / `await` in webpack loaders isn't supported to the best of my knowledge

Yes you are right, the webpack loader interface doesn't support the await/async syntax. In this PR the only method that got changed to an async from a sync method is the `makeSourceMapAndFinish` method. Now, it will immediately return and evaluate in the "background" which is fine since it anyway already had a reference to the `callback` method.
The async/await syntax in this case is only syntactic sugar. If you want, I can also remove that.

johnnyreilly:
If you could convert the code to make it do async in the webpack style I'll take a look. I'm on the fence with this feature tbh but I will consider it. It involves taking another dependency which I'm generally hesitant about

johnnyreilly:
Also don't you need to .destroy as well?

https://www.npmjs.com/package/source-map#new-sourcemapconsumerrawsourcemap

Ka0o0:
Great thanks! I've now removed the usage of async/await.

Just one note about the dependency. `source-map` is developed by Mozilla and is even referenced by the `@types/webpack` npm package, since they use the same source map data structures. In general incredibly many projects rely on this so it's very likely that a user has this dependency somewhere in his project anyway. Also, it has no further dependencies.

johnnyreilly:
Also don't you need to .destroy as well?

https://www.npmjs.com/package/source-map#new-sourcemapconsumerrawsourcemap

Ka0o0:
Good point, I've added that.

src/index.ts
```diff
@@ -661,6 +690,48 @@ function makeSourceMap(
...
+ * This is required when ts-loader is not the first loader in the Webpack loader chain.
+ * If the input source map is not set, then the newly generated source map is returned.
+ */
+function mapToInputSourceMap(
+  sourceMap: RawSourceMap,
+  loaderContext: webpack.LoaderContext<LoaderOptions>,
+  inputSourceMap: Record<string, any>
+): Promise<RawSourceMap> {
+  return new Promise<RawSourceMap>((resolve, reject) => {
+    const inMap = inputSourceMap as RawSourceMap;
```
johnnyreilly, line 704: Rather than casting the incoming source map as a `RawSourceMap` and mutating it, could we create a `RawSourceMap` based upon the `inputSourceMap`?

src/index.ts
```diff
@@ -661,6 +690,48 @@ function makeSourceMap(
   };
 }
 
+/**
+ * This method maps the newly generated @param{sourceMap} to the input source map.
+ * This is required when ts-loader is not the first loader in the Webpack loader chain.
+ * If the input source map is not set, then the newly generated source map is returned.
```
johnnyreilly, line 696: Is this line still true? It looks like that's handled elsewhere now?

src/index.ts
```diff
@@ -130,7 +144,22 @@ function makeSourceMapAndFinish(
...
+  // if we don't have an input source map, we only need to return the newly generated source map
+  // this avoids that we have to make a possibly expensive call to the source-map lib
+  if (inputSourceMap === undefined) {
+    callback(null, output, sourceMap);
+    return;
+  }
+
+  // otherwise we have to make a mapping to the input source map which is asynchronous
+  mapToInputSourceMap(sourceMap, loaderContext, inputSourceMap)
+    .then((mappedSourceMap: RawSourceMap) => {
```
johnnyreilly, line 157: Do we need the type annotation on the method? Doesn't TypeScript infer it?

src/index.ts
johnnyreilly: Looking at the type definition, these seem like quite different types?

https://github.com/mozilla/source-map/blob/60adcb064bf033702d954d6d3f9bc3635dcb744b/source-map.d.ts#L15-L21

src/index.ts
Ka0o0: Thanks, that was an old comment.

src/index.ts
Ka0o0: yes

src/index.ts
Ka0o0: Webpack declares this type, which is the same up to some optionals: https://github.com/webpack/webpack/blob/1f13ff9fe587e094df59d660b4611b1bd19aed4c/types.d.ts#L11978C2-L11986C2  

Unfortunately, they don't export it, which is why I was using the `Record<string, any>` type. I could potentially re-declare the interface inside ts-loader but I think simply casting the type should be sufficient in this case.

How would you do this?

johnnyreilly: Let's ignore the casting issue for now - how would you feel about doing this?

> Rather than casting the incoming source map as a `RawSourceMap` and mutating it, could we create a `RawSourceMap` based upon the `inputSourceMap`?

Ka0o0: I've changed that part so that I don't modify the original object.

johnnyreilly: Thanks - let's see if the existing tests pass

johnnyreilly:
One day we'll get to have the tests fully run

johnnyreilly:
Okay, tests pass. I need to update the test pack for TypeScript 5.2. I'm trying to decide if we should add a comparison test for this also

Ka0o0:
> Okay, tests pass. [I need to update the test pack for TypeScript 5.2](https://github.com/TypeStrong/ts-loader/pull/1633). I'm trying to decide if we should add a comparison test for this also. The `yarn.lock` doesn't seem to have been updated BTW

okay, so I guess you want to merge the other PR first?

Sorry the lock file slipped through then, I'll update it.

Regarding the test, I would love to add a test but I think for this I would need to have more guidance from you. As I wrote in the original issue, my problem with the current ts-loader is that I'm not able to debug Vue projects since the input source map generated by the vue-loader was discarded. For the test, is it okay to add the vue-loader as another dev dependency?

src/index.ts
```diff
@@ -130,7 +144,22 @@ function makeSourceMapAndFinish(
   );
 
   setModuleMeta(loaderContext, instance, fileVersion);
-  callback(null, output, sourceMap);
+
+  // if we don't have an input source map, we only need to return the newly generated source map
+  // this avoids that we have to make a possibly expensive call to the source-map lib
+  if (inputSourceMap === undefined) {
```
Ka0o0, line 150: Also `sourceMap` could potentially be undefined if the ts compiler is not configured to generate source maps. 

This is a bit confusing because the developer has to configure source maps in two places: The webpack config and also the tsconfig. I'm wondering if the property `webpack.LoaderContext.sourceMap` indicates whether webpack would like to have source maps generated or not. If yes, then I guess it makes sense to issue a warning to the user when the ts config source map option and the webpack source map config differ. @johnnyreilly  What do you think?

johnnyreilly: Potentially yes, let's come back to this later though

Ka0o0: Okay, I've fixed the bug where `sourceMap` being undefined would crash ts-loader and then we can later concentrate on the warning.

johnnyreilly:
Yeah I want to get the test pack PR merged first.  Getting meaningless CI failures so I'm using this as an excuse to thin the test matrix a bit. It's way too big.  Could you update the yarn.lock file please?

Ka0o0:
I've rebased my changes onto main and updated the `yarn.lock` file. However, if you look at the file you see that yarn  touched the whole file and removed the double quotes. Not sure why, I made sure I've had the current devcontainer running and simply ran `yarn add source-map`. Any idea what I did wrong or is it fine like this?

johnnyreilly:
I wouldn't worry - probably just different yarn versions

johnnyreilly:
Could you have a go at writing a new comparison test please?  You can learn about comparison tests here: https://github.com/TypeStrong/ts-loader/blob/main/test/comparison-tests/README.md

I would think you can copy / paste the existing [sourceMaps](https://github.com/TypeStrong/ts-loader/tree/main/test/comparison-tests/sourceMaps) test as a starting point.  Don't worry if it turns out to be problematic.  I think I put this test in the "nice to have" camp rather than the "we need this" camp.

Ka0o0:
I've added a comparison test which requires the `vue-loader` as described in my previous comment. I'm a bit unsure if it's okay to add another dependency, although it's just for the tests in this case...

johnnyreilly:
if it's a `devDependency` then it's not a problem - it's runtime dependencies that we care about.

What you've done looks like a good start - but I wonder if we could do the symLink that currently sits in the `package.json` in the same place that the others are done?

https://github.com/TypeStrong/ts-loader/blob/02c20692dc79b216578f60730602212dd3f12081/test/comparison-tests/create-and-execute-test.js#L106

Ka0o0:
Okay I think I actually don't have to link anything really - the thing with `npm link` was only that it was installing the dependencies of the test case. I've added the installation into the `create-and-execute-test.js` script (https://github.com/TypeStrong/ts-loader/pull/1626/commits/3b6223e8946846c5d19286074a96a96a43716b75). I've also had to update the package and yarn lock files and the expected output.

johnnyreilly:
Sorry this keeps slipping my mind.

Do you want to update the version number in the package.json and add an entry to the CHANGELOG.md please? https://github.com/TypeStrong/ts-loader/blob/main/CHANGELOG.md

v9.5.0 seems appropriate

Ka0o0:
No worries, done. Hope the description is okay like this.

johnnyreilly:
Released! Thanks for your work! https://github.com/TypeStrong/ts-loader/releases/tag/v9.5.0

PR #1665, opened 2025-01-10
fix: add more detailed error messages

hai-x:
Small change and make error messages more detailed.

image

### DIFF
--- a/src/servicesHost.ts
+++ b/src/servicesHost.ts
@@ -382,7 +382,6 @@ function createWatchFactory(
     }
     return result;
   }
-  ``;
 
   function invokeRecursiveDirectoryWatcher(
     directory: string,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 9.5.2
+* [fix: add more detailed error messages](https://github.com/TypeStrong/ts-loader/pull/1665) - thanks @hai-x
+
 ## 9.5.1
 * [fix: inputSourceMap can be null](https://github.com/TypeStrong/ts-loader/pull/1639) [#1638] - thanks @johnnyreilly and @michaeltford
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ts-loader",
-  "version": "9.5.1",
+  "version": "9.5.2",
   "description": "TypeScript loader for webpack",
   "main": "index.js",
   "types": "dist",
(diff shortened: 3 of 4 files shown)


johnnyreilly:
Thanks - I think if I understand it correctly, you want to alter the error message? Could you adjust the PR so it only does that please?

src/servicesHost.ts
```diff
@@ -382,7 +382,6 @@ function createWatchFactory(
     }
     return result;
   }
-  ``;
```
hai-x, line 385: Another small fix.

hai-x:
> Thanks - I think if I understand it correctly, you want to alter the error message? Could you adjust the PR so it only does that please?

Ajusted. Could you please review it again?

johnnyreilly:
Let's get the GitHub actions running....

src/servicesHost.ts
johnnyreilly: Good spot!

johnnyreilly:
Please could you update the version in the package.json (patch increment) and put an entry in the CHANGE LOG.md too?

johnnyreilly:
You are using thread-loader I'm guessing? What's the motivation?

hai-x:
> Please could you update the version in the package.json (patch increment) and put an entry in the CHANGE LOG.md too?

Sure.

hai-x:
> Please could you update the version in the package.json (patch increment) and put an entry in the CHANGE LOG.md too?

Done.

johnnyreilly:
Thanks! Published with https://github.com/TypeStrong/ts-loader/releases/tag/v9.5.2

johnnyreilly:
Looks like there's a small issue with the release pipeline. Will have to fix that before I can ship the change

hai-x:
> Looks like there's a small issue with the release pipeline. Will have to fix that before I can ship the change

Yes. It seems to be caused by conflict between pnpm version and node version.

hai-x:
> > Looks like there's a small issue with the release pipeline. Will have to fix that before I can ship the change
> 
> Yes. It seems to be caused by conflict between pnpm version and node version.

I have seen that you have open a related PR. https://github.com/TypeStrong/ts-loader/pull/1666/files

johnnyreilly:
Hopefully this will do it: https://github.com/TypeStrong/ts-loader/actions/runs/12726168149/job/35474238704

Will fix up version numbers later

Scutua (approved):

PR #1676, opened 2025-08-23
chore: TypeScript 5.9 upgrade

johnnyreilly:
First code change in ts-loader for a long time, this was caused due to compilation issues with TS 5.9 and the version of webpack being used in the ts-loader repo. 

This is probably the underlying cause in TS 5.9: https://devblogs.microsoft.com/typescript/announcing-typescript-5-9/#notable-behavioral-changes

I *think* that this lead to changes in the webpack types (certainly the existing types did not compile with TS 5.9 anymore).  webpack 5 seems to have been upgraded, and  I suspect it's @alexander-akait's pull request: https://github.com/webpack/webpack/pull/19777 that lead to the change which allowed ts-loader to fix the 5.9 related issues.  

Upgrading the `devDependency` to the latest webpack@5 in ts-loader and recompiling surfaced different issues in the ts-loader codebase, likely linked to types getting tighter in webpack.  The changes have been catered for in ts-loader, mostly this takes the form of more exhaustive null checks / type narrowing shenanigans.

The comparison testpack had to be regenerated entirely to align with the newer version of webpack 5.

## Other changes

This PR also drops testing against 4.x versions of TypeScript.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 9.5.3
+* [chore: typescript 5.9 upgrade](https://github.com/TypeStrong/ts-loader/pull/1676) - thanks @johnnyreilly
+
 ## 9.5.2
 * [fix: add more detailed error messages](https://github.com/TypeStrong/ts-loader/pull/1665) - thanks @hai-x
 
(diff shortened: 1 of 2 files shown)


alexander-akait:
@johnnyreilly Yeah, webpack is now using `strict` mode for types, so some types need to be checked on `null`/`undefined` and in some places narrowing is required

PR #1695, opened 2026-05-21
fix: TS5011 errors with TypeScript 6.0: transpileModule called with rootDir: undefined

tweet:
backport change to webpack 4 version

#1678

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 25
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v6
 
       - name: install
         run: yarn install
@@ -24,7 +24,7 @@ jobs:
     runs-on: windows-latest
     timeout-minutes: 25
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v6
 
       - name: copy files
         shell: pwsh
@@ -52,10 +52,10 @@ jobs:
         ts: [3.8.3, 3.9.3, 4.0.3, 4.1.5, 4.2.4, 4.3.2, 4.4.2, 4.5.2, next]
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v6
 
       - name: install node
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v6
         with:
           node-version: ${{ matrix.node }}
 
@@ -79,10 +79,10 @@ jobs:
         ts: [3.8.3, 3.9.3, 4.0.3, 4.1.5, 4.2.4, 4.3.2, 4.4.2, 4.5.2, next]
     runs-on: windows-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v6
 
       - name: install node
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v6
         with:
           node-version: ${{ matrix.node }}
 
@@ -112,7 +112,7 @@ jobs:
     name: Lint
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v6
 
       - name: install
         run: yarn install
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,14 +4,19 @@ on:
 
 name: release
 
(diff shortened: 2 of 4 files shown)


johnnyreilly:
If we want to merge and release this we'll also need to migrate this branch to use trusted publishing 

- https://github.com/TypeStrong/ts-loader/pull/1680
- https://github.com/TypeStrong/ts-loader/commit/2421dcfe865e4a9e5f9be452f2d7845ba724185e
- https://github.com/TypeStrong/ts-loader/commit/4a60de489ebf9117b331b323412a12416c5e1ef8

johnnyreilly:
I'm considering adding webpack 4 support back directly and not just on 8.x. Could you confirm if the branch in this pull request works for you? https://github.com/TypeStrong/ts-loader/pull/1697

If you `yarn build` you should be able to plug it into your project and test easily

tweet:
I've test this branch's build, but build with failure because of different `source-map` dependency version used (It's `0.6.1` in my project, but ts-loader needs `0.7.x`).
There are some errors as follows.
```js
error  in ./src/xxx/xxx.vue?vue&type=script&lang=ts
Syntax Error: TypeError: sourceMapConsumer.destroy is not a function
    at Array.forEach (<anonymous>)
```

```js
Syntax Error: TypeError: module.addError is not a function
    at Array.forEach (<anonymous>)
```

johnnyreilly:
Ah thanks - let me investigate!

johnnyreilly:
Can you confirm which specific version of webpack 4 you are using please? Also, where is your version of source-map coming from? 0.6.1 I mean.

johnnyreilly:
I've updated the branch version to guard against unusable versions of `source-map` - this should resolve your situation I think.  I'm still curious as to what brought source-map@0.6.1 into your project?

tweet:
Some of dependency paths:
```
webpack(4.47.0)#terser-webpack-plugin(^1.4.3 -> 1.4.6)#source-map(^0.6.1)
vue(2.7.16)#@vue#compiler-sfc(2.7.16)#source-map(^0.6.1)
webpack#terser-webpack-plugin#terser#source-map-support#source-map
html-webpack-plugin#html-minifier#clean-css#source-map
webpack#webpack-sources(^1.4.3)#source-map
ts-jest(^29.4.6)#handlebars(4.7.9)#source-map
```

johnnyreilly:
okay great - so it's a transitive dependency of webpack itself.  Good to know.

Does the updated branch work for you?  It's now aligned with the webpack-4 branch of ts-loader which didn't actually use source-map at all.  With my changes it only uses source-map if the expected features are present. 0.6.1 does not have those.

tweet:
This error still reported.
```js
Syntax Error: TypeError: module.addError is not a function
    at Array.forEach (<anonymous>)
```

johnnyreilly:
Can you provide more information please? Is there a stack trace?

tweet:
Oh, the upper errors info were processed by friendly-errors-plugin, original is this
```plain
ModuleBuildError: Module build failed (from xxx/node_modules/ts-loader/index.js):
  TypeError: module.addError is not a function
      at xxxx\node_modules\ts-loader\dist\instances.js:365:64
      at Array.forEach (<anonymous>)
      at reportTranspileErrors (xxx\node_modules\ts-loader\dist\instances.js:365:40)
      at successLoader (xxx\node_modules\ts-loader\dist\index.js:69:43)
      at Object.loader (xxx\node_modules\ts-loader\dist\index.js:65:5)
      at xxx\node_modules\webpack\lib\NormalModule.js:316:20
      at xxx\node_modules\loader-runner\lib\LoaderRunner.js:367:11
      at xxx\node_modules\loader-runner\lib\LoaderRunner.js:233:18
      at runSyncOrAsync (xxx\node_modules\loader-runner\lib\LoaderRunner.js:143:3)
      at iterateNormalLoaders (xxx\node_modules\loader-runner\lib\LoaderRunner.js:232:2)
      at xxx\node_modules\loader-runner\lib\LoaderRunner.js:205:4
      at xxx\node_modules\enhanced-resolve\lib\CachedInputFileSystem.js:85:15
      at process.processTicksAndRejections (node:internal/process/task_queues:85:11)
```

johnnyreilly:
I've just made one more change which I hope should handle that.  Could you give it a test when you get a moment please? I'm hoping that should be it

tweet:
My project build is successful now.

johnnyreilly:
Excellent - will probably be cutting a release with @acutmore later this week

johnnyreilly:
We've just released 9.6.0 - hopefully this works for your use case

PR #1696, opened 2026-05-21
Draft of Rspack compatiblity

### LINKED ISSUE #1694: rspack compatibility
Hello,

I'm using this loader with both Webpack and Rspack, and it is working nicely.

However, when using it with Rspack, it requires installing Webpack, which I would like to avoid.

Would you be open to a PR that allows the use of Rspack or Webpack independently?

My idea is that when using a hook, when we get an instance of `compiler`, there is a `compiler.webpack` property (also in Rspack) with a full webpack export inside, which can avoid importing the webpack package itself.

onigoetz:
Fixes #1694

This is a first draft of how rspack compatibility would look like.

I did not add any test yet, but it's a draft to have early feedback.

If this approach looks good to you I'll add the relevant tests and finalize the parts that are not fully fleshed out yet

### DIFF
--- a/src/after-compile.ts
+++ b/src/after-compile.ts
@@ -1,6 +1,6 @@
 import * as path from 'path';
 import type * as ts from 'typescript';
-import * as webpack from 'webpack';
+import type * as webpack from 'webpack';
 
 import * as constants from './constants';
 import { getEmitFromWatchHost, getEmitOutput } from './instances';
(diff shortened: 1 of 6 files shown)


onigoetz:
This might actually be more difficult than expected; as some APIs around `Module` are not currently supported by Rspack:  https://github.com/web-infra-dev/rspack/issues/14119

johnnyreilly:
Oh right - let's see what happens there

PR #1700, opened 2026-06-19
chore: remove semver from dependencies field

hyperz111:
We only use `.gte` method from `semver` package on end-user code. Which can be written like this.

```text
score = current - expected
if (score >= 0):
  return true;
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,6 @@
     "chalk": "^4.1.0",
     "enhanced-resolve": "^5.0.0",
     "micromatch": "^4.0.0",
-    "semver": "^7.3.4",
     "source-map": "^0.7.4"
   },
   "devDependencies": {
@@ -92,6 +91,7 @@
     "mocha": "^6.0.0",
     "prettier": "^2.0.5",
     "rimraf": "^2.6.2",
+    "semver": "^7.3.4",
     "typescript": "^6.0.2",
     "typescript-eslint": "^8.59.4",
     "webpack": "^5.74.0",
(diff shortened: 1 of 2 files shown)


johnnyreilly:
Thanks. This may be a possible optimisation. We've used semver forever, I'm not sure how much I want to remove it. Maybe though

johnnyreilly:
@copilot please review this PR. Examine any possible issues that there might be

hyperz111:
> Thanks. This may be a possible optimisation. We've used semver forever, I'm not sure how much I want to remove it. Maybe though 

I only move semver to devDependencies. Using `semver.gte` method means we have to do a full version parse and it's a bit slower.

We can use [semiver](https://npmx.dev/semiver) if we need to compare pre-release version, because my code isn't doing that.

But I think we can do this.

johnnyreilly:
We definitely can do this, at least for now though I think I'm happy to leave things as they are. Thanks though.

johnnyreilly:
It is interesting to think about this though https://github.com/TypeStrong/ts-loader/tasks/45425eb4-463b-4a8e-90e3-3784eb11735b?author=johnnyreilly&q=is%3Aopen

johnnyreilly:
hmmm now I'm wondering more - I'm getting copilot to experiment: https://github.com/TypeStrong/ts-loader/pull/1701

johnnyreilly:
Interestingly, ts-loader supports TypeScript 3.6.3+.  The semver is used to test whether the compiler version is 3.6.3+.  Given that's the minimum supported version, it's arguable whether the test is even necessary.  I'm not sure it is

johnnyreilly:
I'm still not sure if I'll make changes to ts-loader or not - but I am enjoying pondering it!

johnnyreilly:
Closing in favour of #1701 - thanks for gettting me thinking though!

PR #1706, opened 2026-08-31
feat: add benchmark test harness for comparing ts-loader build performance

johnnyreilly:
Generates a synthetic project and times cold builds and incremental rebuilds (low- vs. high-fan-out file touches, under both transpileOnly modes) to compare one ts-loader checkout's speed against another. Wired into CI as a report-only PR check (Ubuntu + Windows) so perf-sensitive branches get an ongoing, automatic comparison against main rather than relying on ad hoc local measurement.


acutmore (approved):
Great idea!

johnnyreilly:
Thanks @acutmore! I want to get a sense of the difference in perf that 7.1 will make, but this is just generally useful I think

johnnyreilly:
@copilot here are some potential problems/fixes to implement:

Root causes and fixes
The systematic bias (all 6 rows positive, P ≈ 1.6% if random)
Root cause 1 — Shared heap (primary driver)
Both sides loaded their own typescript package from different node_modules paths (rootA/node_modules/typescript and rootB/node_modules/typescript). Because they are different filesystem paths, Node's require() cache treats them as separate modules and loads both into the same process heap simultaneously. With 300 fixture files in full type-check mode, each TypeScript language service instance is substantial. Having both in one process roughly doubled the heap and GC pressure. With only 6 measured samples, even one or two GC pauses landing during side A's runs were enough to make all 6 deltas positive.
This also explains why typeCheck deltas (+9.5 %, +5.5 %, +2.6 %) were consistently larger than transpileOnly ones (+3.0 %, +0.4 %, +1.9 %): full type checking allocates far more objects per file than transpile-only, making it proportionally more sensitive to GC pressure.
Root cause 2 — aggregateTimeout: 100 ms
Every incremental rebuild measurement was inflated by 100 ms of artificial debounce wait (the webpack watcher's settle delay). A rebuild that took ~60 ms was reported as ~160 ms — a 160 % overhead that reduces sensitivity to real differences.
Root cause 3 — Too few samples (6)
6 samples give very little protection against CI host jitter. A single outlier shifts the median significantly.
Fixes applied
File
Change
test/benchmark-tests/run-side.mts
New subprocess entry-point. Runs all (warmup + measured) iterations for one side and emits { durations: number[] } JSON to stdout. Each side now gets a fresh OS process with exactly one TypeScript + webpack instance on its heap.
test/benchmark-tests/run-benchmark.mts
Replaced the in-process interleaved loop with runSideProcess() / runScenario() which spawn run-side.mts via spawnSync. scenarioIndex alternates which side runs first across scenarios to balance time-varying host effects. MEASURED_ITERATIONS bumped 6 → 10.
test/benchmark-tests/scenarios.mts
aggregateTimeout: 100 → aggregateTimeout: 0 so incremental timings measure actual build cost, not debounce latency.

johnnyreilly:
@copilot cold build looks good - the variance is small.

Incremental rebuild, both types, are quite different - please investigate and implement fix such that variance for each type of check are low. There are no changes on src/*.TS right now and so variance should be minimal. See recent results below 

| Scenario | transpileOnly | PR branch median (ms) | base branch median (ms) | Δ vs base branch |
| --- | --- | --- | --- | --- |
| Cold build | false | 836.0 | 838.4 | -0.3% |
| Incremental rebuild (leaf touch) | false | 74.3 | 68.5 | +8.5% |
| Incremental rebuild (hub touch) | false | 386.3 | 375.9 | +2.8% |
| Cold build | true | 470.8 | 466.9 | +0.8% |
| Incremental rebuild (leaf touch) | true | 31.2 | 33.5 | -6.9% |
| Incremental rebuild (hub touch) | true | 32.5 | 31.7 | +2.4% |

johnnyreilly:
@copilot the variance seems as wide as ever. Also the cold build variance which was small before the last commit is now large. See details below. Please revert changes to cold build and work on the incremental rebuild scenario. If it makes sense, separate cold build and incremental rebuild - they may be fundamentally different?


### Benchmark (Windows)

| Scenario | transpileOnly | PR branch median (ms) | base branch median (ms) | Δ vs base branch |
| --- | --- | --- | --- | --- |
| Cold build | false | 993.2 | 770.0 | +29.0% |
| Incremental rebuild (leaf touch) | false | 72.9 | 66.9 | +8.9% |
| Incremental rebuild (hub touch) | false | 365.7 | 375.8 | -2.7% |
| Cold build | true | 429.5 | 413.4 | +3.9% |
| Incremental rebuild (leaf touch) | true | 29.9 | 33.7 | -11.1% |
| Incremental rebuild (hub touch) | true | 30.6 | 34.1 | -10.2% |

_PR branch = `C:\source\ts-loader`, base branch = `C:\source\ts-loader-main`. 2 warmup + 10 measured iterations per scenario, median reported. Report-only - no threshold fails this check._

johnnyreilly:
@copilot please revert the code back to https://github.com/TypeStrong/ts-loader/pull/1706/commits/8fbb6004937c338b841cb99dbed5c02f42635dff - the changes since have made things worse not better

PR #1708, opened 2026-09-06
Migrate chalk from 4.1.2 to 5.0.0

enismustafaj:
Automated migration for `chalk` 4.1.2 -> 5.0.0, generated and tested by apiweiser-cli's coding agent.

### Breaking

- This package is now pure ESM. Please read this before upgrading.
- If you use TypeScript, you need TypeScript 4.7 or later.
- Require Node.js 12.20
- Move some properties off the default export to individual named exports:
  - chalk.Instance -> Chalk
  - chalk.supportsColor -> supportsColor
  - chalk.stderr -> chalkStderr
  - chalk.stderr.supportsColor -> supportsColorStderr
- Remove .keyword(), .hsl(), .hsv(), .hwb(), and .ansi() coloring methods
- The tagged template literal support moved into a separate package: chalk-template

### Improvements

- Bundle dependencies - Chalk no longer has any dependencies
- Install size is less than half of v4
- Add overline style
- Preserve function prototype methods


johnnyreilly:
Thanks but this repository does not accept unknown automated PRs. We're also planning to migrate away from Chalk entirely

